PR #1, opened 2026-05-23
Add poster-aware collection templates and async sync queueing

Quick104:
## Summary
- Ported built-in collection templates to include TMDB discover and franchise coverage, plus poster assets and template metadata.
- Propagated template defaults into created collections, including sort order, poster URL, and collection management behavior.
- Added poster-aware apply/import flows so existing collections can be reused, posters are preserved or applied deterministically, and large template bundle syncs can be queued asynchronously.
- Tightened collection sync behavior for MDBList, TMDB, and Trakt sources by using shared limit helpers and more accurate sync accounting.
- Documented the collection template poster generation workflow in the admin wiki.

## Testing
- Not run (not requested).
- Added/updated unit coverage for collection template definitions and poster asset expectations in `internal/collections/templates/templates_test.go`.
- No frontend or integration verification run.

## Summary by CodeRabbit

* **New Features**
  * Choose collection poster (server default or custom URL) when applying templates or importing; apply flow can queue template-bundle apply jobs and shows "sync queued" info.

* **Improvements**
  * Bundled template posters can be persisted to public asset storage; admin-uploaded posters are preserved while template-origin posters are tracked and refreshable. Increased item limits (UI and backend) and clearer sync/timeout reporting; live job status banner for template applies.

* **Tests**
  * Added template asset and poster-upload unit tests.

* **Documentation**
  * Added poster generation workflow and expanded multi-repo guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Silo-Server/silo-server/pull/1?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #17, opened 2026-05-27
feat(audiobooks): Audiobookshelf-compat + sort/filter overhaul + accumulated work

RXWatcher:
> [!WARNING]
> **This PR is enormous (214 commits, 469 files, +89,647 / -2,134 lines, 28
> new migrations) and accumulates roughly three weeks of work across many
> sessions. It needs a VERY thorough review before merging.**
>
> A line-by-line read of the full diff is impractical — review it in slices:
>
> 1. **`internal/audiobooks/`** — the ABS-compat package (largest blast
>    radius; touches login, sessions, libraries, items, RSS, collections,
>    playlists, smart collections, bookmarks). Also the standalone
>    `:13378` listener wiring in `cmd/silo/main.go`.
> 2. **`internal/catalog/`** — sort/filter scope changes, new
>    Author/Narrator/Series facet + sort fields, typeahead endpoint,
>    1000-row facet cap.
> 3. **`migrations/132..160`** — 28 new migrations. Three are renumber
>    chores from main merges. **156 is destructive** (drops the five
>    `abs_*` tables); **158 truncates `audiobook_series`**; both already
>    applied on silo-prod.
> 4. **`web/src/**`** — audiobook library UX, filter sections, typeahead
>    `FacetSearchSelect`, plus inherited main changes (PageBack,
>    metadata curation permission UI).
> 5. **`internal/scanner/`** — audiobook parser fix (`album` no longer
>    falls back to Series — fixed 87% of `audiobook_series` table noise);
>    perf work (parallel enricher, deferred cover extraction, unchanged-
>    folder skip path).
>
> Migration 156 cannot be undone on a live DB without data loss. Migration
> 158 is intentionally destructive (clears the table for repopulation by
> the parser-fixed scanner). Migration 159/160 are renumber chores; the
> running silo already has them applied. Migration 161+ does not exist.

## Summary

Long-lived integration branch consolidating ~3 weeks of work, headlined by
the audiobookshelf-compatible API and the audiobook-library UX overhaul.

### Highlights

**ABS-compat surface (silo speaks audiobookshelf)**
- Dedicated `:13378` listener serving the full ABS protocol — login,
  libraries, items, sessions, RSS, collections, playlists, smart
  collections — for the official Audiobookshelf mobile + desktop clients.
- Login supports both `alice` and `alice#kids` username forms (the latter
  selects a specific household profile; password stays the user's).
- WebSocket socket.io shim, `/authorize` token rotation for ABS v2.26+,
  `/me`, `/libraries`, `/authors`, `/series`, `/playlists`, `/collections`,
  `/bookmarks`, `/me/listening-sessions`, RSS feed publish.
- Token storage isolated from silo's primary session store; logout-
  without-active-session works.

**Audiobook library UX**
- Per-library sort relevance scope: audiobook libraries no longer show
  video-only sorts (Resolution, IMDb/RT ratings, Content Rating).
- New audiobook-native sort options: **Author / Narrator / Series** (the
  last sorts by `series_name` → `series_index` for narrative order).
- Per-library filter section gating: Director/Writer/Producer,
  Studio/Network, Video Quality, etc. hidden on audiobook libraries.
- New filter sections: **Author / Narrator / Series** backed by
  `item_people` (kind=7/8) and `audiobook_series`.
- Filter response capped at 1000 entries/facet (was 11.7 MB on a 220k
  library); high-cardinality facets switch to server-side typeahead via
  `/api/v1/catalog/filters/search` + new `FacetSearchSelect` component.

**Scanner / enricher perf**
- Parallel enricher sweep with configurable worker count.
- Inline cover extraction lifted out of the scan path; deferred to the
  enricher with a local-file fallback.
- Folder change-detection skip path so unchanged folders cost nothing.
- `audiobook_series` upsert + `ReplacePeople` skip-when-unchanged.
- `internal/scanner/audiobook.go` parser no longer falls back to the
  `album` tag for Series — that tag holds the book title in audiobook
  conventions, and the fallback was producing one fake singleton
  "series" per book (141k / 87% noise on this server's library).
  Migration 158 truncates the table so the next scan rebuilds it from
  real `series` / `mvnm` tag data only.

**Collections unification (migration 156)**
- Drops the five `abs_*` collection tables (`abs_user_collections`,
  `abs_collection_items`, `abs_playlists`, `abs_playlist_items`,
  `abs_smart_collections`); ABS store adapters now query the canonical
  `user_personal_collections` tables.

**Other shipped in this branch**
- Phase-1 ABS close-out: author/series detail endpoints, listening
  stats, continue-listening toggles, public RSS XML routes.
- Media requests + Radarr/Sonarr fulfillment plumbing (pre-existed in
  the branch from earlier work, brought along).
- Skeleton merge of recent main: jellycompat signed image tags,
  PageBack component, metadata curation permission, episode catalog
  performance fast path, relative-date filters.

### Schema impact

28 new migrations (133–160; three are renumber chores from the two
main merges). Notable:

- `156_unify_user_collections` drops 5 `abs_*` tables (irreversible in
  practice — down migration restores empty tables but data is lost).
- `158_audiobook_series_truncate` clears the table to be repopulated
  by the next scan after the parser fix.
- `145_audiobook_series` creates the `audiobook_series` table.

The startup migration runner applies these automatically on deploy.

### Risks / call-outs

- **Massive diff** — see the warning at the top.
- Migration 156 is destructive (drops `abs_*` tables). Already applied
  on silo-prod during the deploy of `55442e6`.
- Migration 158 also already applied; `audiobook_series` rebuilt to
  ~43k rows from the parser-fixed scan before the scan was interrupted
  by a later deploy. A fresh full scan will fully repopulate.
- ABS-compat clients connect to `:13378`, separate listener from the
  main silo port. Docker compose + Dockerfile expose the new port.
- The 220k-book library's catalog `/filters` response weighs ~146 KB
  with the 1000-row cap. The typeahead endpoint handles the rest.

### Follow-ups (deferred, plans on disk)

- `docs/superpowers/plans/2026-05-27-audiobook-catalog-filter-fields.md`
  (this PR implements it)
- `docs/superpowers/plans/2026-05-27-audiobook-series-data-cleanup.md`
  (options B / D landed via migration 158 + parser fix)
- `docs/superpowers/plans/2026-05-27-catalog-facet-typeahead.md` (this
  PR implements it)
- The four `2026-05-27-audiobook-{cover,enricher,scan-*}` plan files
  in the working tree are notes from the perf work; intentionally
  uncommitted per project preference.

## Test plan

- [ ] `make lint` (`golangci-lint` + frontend eslint) clean
- [ ] `cd web && pnpm run format:check` clean for files this branch
      touches (pre-existing baseline failures on `main` remain)
- [ ] `cd web && pnpm run build` clean
- [ ] `go test ./...` clean
- [ ] `cd web && pnpm vitest run` clean (33 pre-existing failures on
      `main` remain)
- [ ] Live deploy verified: `:8080/api/v1/health` → ok,
      `:8096` Jellyfin compat ok, `:13378/ping` ABS ok
- [ ] ABS mobile login with `user` and `user#profile` forms
- [ ] Audiobook library page: sort dropdown shows audiobook-relevant
      options only; filter sheet hides video-only sections; Author/
      Narrator/Series typeahead returns matches
- [ ] Migration 156 + 158 + 159 + 160 applied cleanly on silo-prod

## AI-use disclosure

Authored end-to-end with Claude Code (Opus 4.7, 1M context). All commits
are co-authored to `Claude Opus 4.7 (1M context) <noreply@anthropic.com>`.
The branch is a multi-session accumulation; the design decisions and
trade-offs were made interactively with the user across roughly a
dozen sessions logged in `docs/superpowers/plans/`.


Quick104:
Validated the reported code-review findings. Disposition: all eight are valid. Focused verification also passes locally: `go test ./internal/audiobooks/abs ./internal/audiobooks`.

### Findings

- **[P0] ABS login trusts spoofable proxy headers on the public listener.**  
  `internal/audiobooks/abs/login.go:31` trusts `X-Silo-User-Id` and mints tokens without credentials. `/login` is mounted unauthenticated at `internal/audiobooks/abs/handler.go:305`, and the dedicated ABS listener mounts the handler directly at `cmd/silo/main.go:1751`. A client reaching `:13378` can spoof the header and bypass passwords.

- **[P1] ABS ignores Silo profile/library access controls.**  
  `internal/audiobooks/abs/libraries_handler.go:23` lists all enabled audiobook libraries, and `internal/audiobooks/media_store.go:58` lists by audiobook/library only. The normal `AccessFilter` scope is not applied, so restricted profiles can browse/play audiobook content outside their policy.

- **[P1] Same-account profile isolation is broken for ABS collections/playlists/smart collections.**  
  Direct get/update/delete checks compare only `UserID`, not `ProfileID`: `internal/audiobooks/abs/collections_handler.go:225`, `internal/audiobooks/abs/playlists_handler.go:247`, `internal/audiobooks/abs/smart_collections_handler.go:131`. Profiles on the same account can access and mutate each other’s private objects.

- **[P1] Logout does not invalidate refresh tokens.**  
  `internal/audiobooks/abs/login.go:515` revokes only the supplied bearer/access JTI. `internal/audiobooks/abs/login.go:397` accepts refresh tokens directly and checks only the refresh JTI, so a logged-out client or stolen refresh token can mint fresh tokens.

- **[P1] Migration 156 drops existing ABS collection/smart-collection data.**  
  `migrations/156_unify_user_collections.up.sql:22` migrates only `abs_playlists`, then drops `abs_collection_items`, `abs_user_collections`, and `abs_smart_collections` at `migrations/156_unify_user_collections.up.sql:54`. The down migration recreates those tables empty, making this unrecoverable user data loss.

- **[P1] The audiobooks feature flag does not gate the ABS listener.**  
  `migrations/160_audiobooks_feature_flag.up.sql:5` defaults `audiobooks.enabled=false`, but `cmd/silo/main.go:1345` builds the service/handler unconditionally and `cmd/silo/main.go:1751` starts the listener when `audiobookshelf_compat.listen` is set. The default listen address is `:13378` at `internal/config/db_loader.go:360`, and Docker exposes it at `docker-compose.yml:46`.

- **[P2] ABS library filtering is incorrect for large libraries.**  
  `internal/audiobooks/abs/libraries_handler.go:155` caps filtered fetches at 5000 and then filters in memory at `internal/audiobooks/abs/libraries_handler.go:178`. Matches outside the first 5000 sorted books are silently missed.

- **[P2] Login rate limiting can be bypassed by spoofing `X-Forwarded-For`.**  
  `internal/audiobooks/abs/login_ratelimit.go:101` uses client-supplied `X-Forwarded-For` before `RemoteAddr`, and `internal/audiobooks/abs/login.go:66` uses that value as the rate-limit key on the standalone login path. Attackers can rotate XFF values to get fresh buckets.

Quick104:
## Code review — max effort

5 finder angles × ≤8 candidates, 7 parallel verifiers, plus a sweep pass over uncovered modules. Final list of 15 ranked most-severe first.

Phase tally: 33 CONFIRMED, 3 PLAUSIBLE, 0 REFUTED, +8 sweep candidates.

---

### 1. `migrations/156_unify_user_collections.up.sql:55` — drops user-created collections & smart collections on apply

DROPs `abs_user_collections`, `abs_collection_items`, and `abs_smart_collections` without first copying rows into `user_personal_collections`. Only `abs_playlists` data is migrated. Any deployment that ran 149/150/153 and accumulated rows loses them. The matching `down.sql` recreates the tables empty, so rollback can't recover the data either.

### 2. `migrations/146_audiobook_title_cleanup.down.sql:6` — clobbers original_title for every audiobook on rollback

```sql
UPDATE media_items mi
SET title = mi.original_title, sort_title = LOWER(mi.original_title),
    original_title = '', updated_at = NOW()
WHERE mi.type = 'audiobook'
  AND COALESCE(mi.original_title, '') <> ''
  AND mi.original_title <> mi.title;
```

Touches every audiobook with `original_title` set — including rows populated by metadata plugins, manual edits, and `item_repo` upserts. The migration's own comment concedes "we can't perfectly identify which rows were cleaned." Rolling 146 back permanently destroys original_title for unrelated rows.

### 3. `internal/audiobooks/abs/file_handler.go:154` — public-track session ID is a permanent bearer capability

`handlePublicTrack` streams audio if `PlaybackSessionStore.GetPlaybackSession(sid)` returns non-closed. Migration 143 has no `expires_at`, no TTL check, no rate limit, no user verification. A session ID that leaks via proxy access logs, URL screenshots, browser history, or intercepted requests grants unlimited streaming forever — even after the user logs out, has access revoked, or the book is removed from the library.

### 4. `internal/audiobooks/abs/play_response.go:205` — bearer JWT embedded in track URLs as `?token=`

```go
contentURL := baseURL + "/abs/api/items/" + contentID + "/file/" + ino
if accessToken != "" { contentURL += "?token=" + accessToken }
```

Every play-start response writes the raw ABS access JWT into the track URL query string. JWTs end up in reverse-proxy access logs, browser history, S3/Referer headers, and any structured log of the audio request stream. Anyone with log access lifts live access tokens and impersonates users until expiry (~1h).

### 5. `internal/audiobooks/abs/login.go:437` — refresh-token TOCTOU mints two valid auth chains

```go
row, err := h.deps.TokenStore.GetTokenByJTI(r.Context(), claims.JTI)
if row.RevokedAt != nil { ... }   // ← check
// ... mint new (access, refresh) pair, persist ...
h.deps.TokenStore.RevokeTokenByJTI(r.Context(), claims.JTI)  // ← revoke (later)
```

No transaction, no row lock, no atomic UPDATE … WHERE revoked_at IS NULL RETURNING. Two concurrent `/auth/refresh` calls with the same refresh JWT both pass the check, both mint, and both revoke the same old JTI — leaving two parallel valid refresh chains alive. A stolen refresh token can be raced against the legitimate client to fork an undetectable session.

### 6. `internal/audiobooks/abs/login_ratelimit.go:102` — X-Forwarded-For trusted unconditionally; rate limit trivially bypassed

`clientIP()` honors the leftmost `X-Forwarded-For` header with no trusted-proxy allowlist. Attacker rotates XFF on every `/login` attempt; each spoofed IP gets its own `loginLimitBurst=10` token bucket, defeating per-IP credential-stuffing protection on any deployment that doesn't strip XFF at the edge (including direct-TCP probing of the `:13378` listener).

### 7. `internal/imagecache/imagecache.go:294` — SSRF via plugin-supplied image URLs

`downloadImage` calls `http.DefaultClient.Do` with no scheme/host allowlist; the default client follows 3xx redirects. A malicious plugin returns `SourceURL='http://attacker/redir'` which 302s to `http://169.254.169.254/latest/meta-data/iam/security-credentials/` or `http://127.0.0.1:9200/_cluster/state`. The bytes get thumbhashed and uploaded to the cache backend — if that prefix is publicly readable, internal data is exfiltrated; otherwise it's still an arbitrary internal-port probe.

### 8. `internal/auth/session.go:119` — NULL ip_address can't be scanned; auth session listing 500s for ABS users

INSERT was fixed to allow NULL when `session.IPAddress == ""` (ABS cred validator passes `ip=''`), but `scanSession` still scans `host(ip_address)` into a plain `string`, not `*string`/`pgtype.Text`. The user's next `GET /auth/sessions` (or any RevokeSession / GetByID admin path) errors `cannot scan NULL into *string` → 500. Session-listing UI is broken for every account that ever logged in through ABS-compat.

### 9. `internal/audiobooks/service.go:184` — `audiobooks.enabled` kill-switch is never checked

Migration 160 seeds `audiobooks.enabled='false'` as a documented "master kill-switch." `Service.Enabled()` is implemented at this line but no caller invokes it. `cmd/silo/main.go` builds and mounts the ABS handler unconditionally, then `_ = audiobooksService`. Routes, sidebar nav, scanner branches all run regardless of the flag. There is no working way to disable the feature.

### 10. `internal/audiobooks/abs/smart_collections_handler.go:384` — candidates miss Authors/Narrators/Series/Publisher/Duration

```go
it := smartcoll.Item{
    ID: mi.ContentID, Title: mi.Title, Genres: mi.Genres,
    Year: mi.Year, Language: mi.OriginalLanguage,
}
// Authors, Narrators, Series, Publisher, DurationSeconds never populated
```

The evaluator (`smartcoll/evaluator.go`) matches rules on those exact fields. A user creating "Books by Stephen King" gets zero matches because `Authors` is always empty. Every author/narrator/series/publisher/duration smart-collection rule silently lists zero items — feature appears completely broken on first use. Code comment at :378 admits "best-effort (zero-valued in v1)."

### 11. `internal/audiobooks/abs/handler.go:298` — `/ping`, `/init`, `/healthcheck` registered twice; ABS handshake responses silently overwritten

```go
for _, prefix := range []string{"", "/abs/api"} {
    r.Get(prefix+"/ping", h.handleABSPing)        // ABS-shaped JSON
    r.Get(prefix+"/init", h.handleABSInit)
    r.Get(prefix+"/healthcheck", h.handleABSPing)
}
// ...later...
for _, prefix := range []string{"", "/abs", "/api", "/abs/api"} {
    r.Get(prefix+"/ping", h.handlePing)            // generic {success:true}
    r.Get(prefix+"/init", h.handleInit)
    r.Get(prefix+"/healthcheck", h.handleHealthcheck)
}
```

chi silently overwrites the same method+path. The second loop wins for `""` and `"/abs/api"`. ABS mobile clients expecting `{pong:true, server:"audiobookshelf", version:…}` (the "must match continuum-plugin exactly" shape) receive `{success:true}` — connect-server flow rejects the server. Same break for `/init` and `/healthcheck`.

### 12. `internal/audiobooks/abs_playlist_store.go:221` — playlist ON CONFLICT key omits sub_item_id; duplicate podcast episodes dropped

```sql
ON CONFLICT (user_id, collection_id, media_item_id) DO NOTHING
```

For podcasts, `media_item_id` is the show and `sub_item_id` is the episode. Adding two episodes of the same podcast to one playlist silently DO-NOTHINGs the second insert. User sees only episode 1; no error, no UI feedback. The file's own header comment documents this caveat.

### 13. `internal/audiobooks/abs/libraries_handler.go:666` — Series field built from Studios (publisher); books grouped by publisher

```go
series := make([]SeriesObj, 0, len(item.Studios))
for _, s := range item.Studios {
    series = append(series, SeriesObj{ID: slugify(s), Name: s})
}
```

`audiobook_scan.go:485` writes `book.Publisher` into `item.Studios`. The actual series name lives in a separate `audiobook_series` table. ABS responses surface publisher names as series; `CollapseBySeries` then folds every Penguin Random House book into a single tile.

### 14. `internal/audiobooks/abs/progress.go:208` — handleSetItemProgress read-modify-write race; multi-device updates silently overwrite each other

```go
var cur ProgressRow
if existing, err := h.deps.ProgressStore.GetProgress(...); err == nil && existing != nil {
    cur = *existing
}
// ... mutate fields ...
h.deps.ProgressStore.UpsertProgress(r.Context(), next)
```

No version/serial/timestamp guard. Device A and Device B PATCH within the same heartbeat window, both read baseline currentTime=100, A wants 110+isFinished=true, B wants 105. Whichever upsert lands second wins — `isFinished=true` is silently lost, or position rewinds from 110 → 105.

### 15. `web/src/pages/audiobooks/player/useAudiobookPlayback.ts:245` — end-of-chapter sleep timer is dead code

```ts
const currentChapter = chapters.find(
  (c) => currentTime >= c.start_seconds && currentTime < c.end_seconds
) ?? null;
// ...
if (sleepSetting.kind !== "end-of-chapter" || !currentChapter) return;
if (currentTime < currentChapter.end_seconds) return;  // ← always true by construction
const audio = audioRef.current;
if (audio && !audio.paused) audio.pause();
```

`currentChapter` is non-null only when `currentTime < currentChapter.end_seconds` (by the find predicate). The early return at line 245 fires every render; the pause path is unreachable. The marquee sleep mode for this PR never pauses playback.

---

**Top 3 to fix before merge:** **#1** (data destruction on migrate apply), **#3** (permanent unauthenticated stream capability), **#4** (JWT leak through URL query).

RXWatcher:
Verification update after merging latest `main` into `feat/audiobooks` and adding the ABS listener feature-flag fix (`b9a9f4b`).

What passed:
- `go test ./...`
- `cd web && pnpm run build`
- Restored pre-audiobook dump `/root/silo-db-backups/silo-before-audiobooks-pr17-20260606T123833Z.dump` into temp DB `silo_audiobooks_rehearsal_20260606` on `silo-new`.
- Ran the Silo migrator against that temp DB: migrations applied through version 180, including audiobook migration versions 143-161.
- Temp DB confirms `server_settings.audiobooks.enabled=false` after migration.
- Targeted listener gate regression passes: `go test ./internal/config -run TestLoadFromDBAudiobooksFeatureFlagGatesCompatListener -v`.

GitHub now reports the PR as mergeable. Runtime boot smoke against the restored temp DB was blocked by an existing invalid stored plugin archive in that database before HTTP serving completed, so I did not count that as listener-bind proof; the config regression covers the specific disabled-listener behavior changed here.

RXWatcher:
Closing this oversized PR in favor of a stacked series of smaller PRs. The existing `feat/audiobooks` branch is preserved as the source-of-truth while the stack is rebuilt into logical review units. Split plan: `docs/superpowers/plans/2026-06-06-audiobooks-stacked-pr-split.md`.

PR #29, opened 2026-05-31
fix(auth): avoid unnecessary session invalidation on updates that do not affect security-relevant fields

zZebrahz:
## Summary

Fixes admin user updates so submitting access-related fields with their current values does not unnecessarily invalidate active sessions or profile tokens.

- Compare submitted access-policy fields against the current user before revoking sessions.
- Return `404` for missing users on admin update paths instead of treating all pre-update fetch failures as not found.
- Keep `access_policy_revision` stable when access-policy update fields are present but unchanged.
- Extend the focused admin handler test coverage for changed vs unchanged revocation inputs.

## Root Cause

The admin update path treated any submitted access-sensitive field as a reason to revoke sessions, even when the effective value did not change. Tightening that decision exposed a related mismatch in the repository: `UserRepository.Update` still bumped `access_policy_revision` whenever an access-policy field was non-nil, which could invalidate profile verification tokens without a matching session revocation.

## Validation

- `go test -count=1 ./internal/auth ./internal/access`
- `git diff --check`

Could not run `go test -count=1 ./internal/api/handlers` in the local container because `github.com/h2non/bimg` requires `libvips`/`vips.pc`, which is not installed in this environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin updates no longer revoke sessions unless security-relevant fields (password, role, permissions, enabled, max playback quality) actually change; changing library access alone no longer triggers logouts.
  * Access-policy revision now advances only when access-control values truly differ, preventing spurious bumps.

* **Tests**
  * Tests updated to validate revocation logic and library-access behavior, including nil vs empty library cases.

### DIFF
--- a/internal/api/handlers/admin.go
+++ b/internal/api/handlers/admin.go
@@ -774,7 +774,6 @@ func updateMayRequireSessionRevocation(input models.UpdateUserInput) bool {
 	return input.Password != nil ||
 		input.Role != nil ||
 		input.Enabled != nil ||
-		input.LibraryIDs != nil ||
 		input.Permissions != nil ||
 		input.MaxPlaybackQuality != nil
 }
@@ -791,14 +790,6 @@ func updateRequiresSessionRevocation(current *models.User, input models.UpdateUs
 	}
 	if input.Enabled != nil && *input.Enabled != current.Enabled {
 		return true
-	}
-	if input.LibraryIDs != nil {
-		if (*input.LibraryIDs == nil) != (current.LibraryIDs == nil) {
-			return true
-		}
-		if *input.LibraryIDs != nil && !slices.Equal(*input.LibraryIDs, current.LibraryIDs) {
-			return true
-		}
 	}
 	if input.Permissions != nil && !slices.Equal(*input.Permissions, current.Permissions) {
 		return true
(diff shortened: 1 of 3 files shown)


zZebrahz:
Going to quickly revise so that library access changes do not invalidate sessions

zZebrahz:
Updated the branch to make library access changes refresh policy without hard-revoking auth sessions.

What changed:
- `LibraryIDs` updates are no longer considered a session revocation trigger in the admin handler.
- `access_policy_revision` still advances through the repository when the library scope actually changes, so profile/PIN policy tokens and request-time access scope still refresh.
- Existing TV/client sessions should remain valid instead of forcing re-authentication after routine library access edits.

Validation:
- `go test -count=1 ./internal/api/handlers`

zZebrahz:
Follow-up fix pushed for the selected-libraries -> all-libraries case.

Confirmation:
- The admin handler was no longer hard-revoking auth sessions for `LibraryIDs` changes.
- The remaining invalidation came from `UserRepository.Update`: changing `users.library_ids` still bumped `access_policy_revision`, which invalidated durable profile/session-adjacent tokens.

Fix:
- `library_ids` updates still persist normally.
- `library_ids` no longer participates in the `access_policy_revision` bump.
- Request-time access enforcement still reads `users.library_ids` from the DB, so library access changes apply without forcing profile/session re-authentication.

Validation:
- `go test -count=1 ./internal/auth ./internal/access ./internal/api/handlers`

zZebrahz:
@Quick104 this is ready for your review

PR #32, opened 2026-05-31
[codex] show plugin OAuth providers on login

### LINKED ISSUE #30: OAuth providers are filtered out in useAuth before Login.tsx renders, so plugin OAuth sign-in options never appear
## Summary
When plugin OAuth providers are enabled, `/api/v1/auth/providers` correctly returns them, but the login page still does not show OAuth sign-in options (e.g. Discord). The frontend drops OAuth providers during auth state initialization.

## Steps to reproduce
1. Install and enable a plugin auth provider with OAuth mode (for example Discord).
2. Enable its auth binding.
3. Call `/api/v1/auth/providers` and confirm an OAuth provider is present, e.g.:
   - `id: "plugin:3:discord"`
   - `mode: "oauth"`
4. Open the login page.

## Expected behavior
The login page should show OAuth sign-in options such as “Sign in with Discord”.

## Actual behavior
Only credential-based login options are shown, even though OAuth providers are returned by the API.

## Root cause
In `web/src/hooks/useAuth.tsx`, providers are filtered before being stored:

`setProviders((availableProviders ?? []).filter((entry) => entry.mode === "credentials"));`

This removes OAuth providers before `Login.tsx` can render them.

## Suggested fix
Keep both credentials and OAuth providers in auth state (remove this filter or make it mode-aware in UI rendering), and let `Login.tsx` render appropriate actions per provider mode.

## Impact
- OAuth plugin auth appears broken in the UI despite correct backend configuration.
- Admins can see provider data in API responses, but users cannot initiate OAuth login from the login page.

Quick104:
## Summary

Fixes #30 by preserving usable providers returned from `/api/v1/auth/providers` in the web auth context. `Login.tsx` already separates credential and OAuth providers, but the auth hook was filtering the list down to credential providers before the login page could render OAuth sign-in buttons.

This also keeps the backend provider list aligned with route availability: OAuth providers are only advertised when the router actually mounted the `/api/v1/auth/oauth/{installation_id}/init` flow. Servers without `SILO_PUBLIC_URL` configured will not show unusable OAuth buttons.

## Impact

Custom auth plugins that declare OAuth-only login now appear on the login screen when OAuth is configured and submit through `/api/v1/auth/oauth/{installation_id}/init` as intended. If OAuth routes are unavailable, those providers stay hidden instead of leading users to a 404.

@thezak48 could you test this PR with your custom OAuth plugin and confirm the sign-in option now appears and starts the OAuth flow correctly on a server with `SILO_PUBLIC_URL` configured?

## Validation

- `go test ./internal/api/handlers`
- `go test ./internal/api`
- `pnpm --dir web exec vitest run src/hooks/AuthProvider.test.tsx src/hooks/useAuth.test.ts`
- `pnpm --dir web exec prettier --check src/hooks/AuthProvider.test.tsx src/hooks/useAuth.tsx`
- `pnpm --dir web exec tsc -b --pretty false`
- Previous changed-file ESLint check: `pnpm --dir web exec eslint src/hooks/useAuth.tsx src/hooks/AuthProvider.test.tsx`

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication providers are now correctly filtered based on OAuth route availability. OAuth providers are hidden when routes are unavailable and displayed when available.
  * Provider list during setup now includes all available authentication methods.

* **Tests**
  * Added test coverage for authentication provider filtering behavior.

### DIFF
--- a/internal/api/handlers/auth.go
+++ b/internal/api/handlers/auth.go
@@ -17,9 +17,10 @@ import (
 
 // AuthHandler handles authentication-related HTTP endpoints.
 type AuthHandler struct {
-	service *auth.Service
-	jwt     *auth.JWTService
-	device  *auth.DeviceLoginService
+	service              *auth.Service
+	jwt                  *auth.JWTService
+	device               *auth.DeviceLoginService
+	oauthRoutesAvailable bool
 }
 
 // NewAuthHandler creates a new AuthHandler backed by the given auth, JWT,
@@ -30,6 +31,13 @@ func NewAuthHandler(service *auth.Service, jwt *auth.JWTService, device *auth.De
 		jwt:     jwt,
 		device:  device,
 	}
+}
+
+// SetOAuthRoutesAvailable controls whether OAuth login providers are
+// advertised by /auth/providers. The router only mounts OAuth routes when the
+// server has enough configuration to complete the flow.
+func (h *AuthHandler) SetOAuthRoutesAvailable(available bool) {
+	h.oauthRoutesAvailable = available
 }
 
 // --- Request/Response types ---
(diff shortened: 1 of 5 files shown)


thezak48:
Fixes the issue for #30

PR #33, opened 2026-05-31
[codex] keep plugin auto-provision passwords bcrypt-safe

### LINKED ISSUE #31: OAuth auto-provision fails for plugin auth providers: generated password exceeds bcrypt 72-byte limit
## Summary
When `auto_provision` is enabled for a plugin OAuth auth provider (e.g. Discord), first-time OAuth sign-in fails because Silo generates a plugin-only password longer than bcrypt allows.

## Steps to reproduce
1. Install and enable an OAuth auth plugin (Discord works).
3. Enable auth binding with `auto_provision: true`. (not visible in the UI yet so a PUT request to the API handles that)
4. Sign in via OAuth with a user that does not yet exist in Silo.

## Expected behavior
Silo auto-creates the user account and completes login.

## Actual behavior
Login fails during auto-provisioning. Server log shows:

`silo ... level=WARN msg="oauth login completion failed" installation_id=3 error="auto-provision plugin user: hashing password: bcrypt: password length exceeds 72 bytes"`

## Root cause
In `internal/auth/plugin_provider.go`, `randomPluginOnlyPassword()` currently generates:

- `"plugin-only-" + hex.EncodeToString(32 random bytes)`

That yields a 76-character password (`12 + 64`), which exceeds bcrypt’s 72-byte max input length.

## Suggested fix
Ensure generated plugin-only passwords are <=72 bytes before hashing, for example by reducing random bytes from 32 to 24 (12 + 48 = 60 chars), or by enforcing a hard cap.

## Impact
This blocks onboarding for new OAuth users whenever plugin auth auto-provision is enabled.

Quick104:
## Summary

Fixes #31 by reducing the random byte payload used by `randomPluginOnlyPassword()` from 32 bytes to 24 bytes before hex encoding.

The generated plugin-only password changes from `plugin-only-` + 64 hex chars, 76 bytes total, to `plugin-only-` + 48 hex chars, 60 bytes total. That keeps the value below bcrypt's 72-byte input limit while preserving enough entropy for an internal password that is never meant for local login.

## Relationship to #30

This is related to the same plugin OAuth sign-in surface, but it is a separate backend auto-provisioning failure. #30 handled whether OAuth providers appear on the login page; this fixes first-time OAuth account creation once the OAuth flow completes.

## Validation

- `go test ./internal/auth`
- `go test ./internal/api/handlers`

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted plugin password generation to ensure compatibility with bcrypt's character limit, preventing potential hashing failures.

* **Tests**
  * Added validation test to confirm generated plugin passwords fit within bcrypt's 72-character input limit and hash successfully.

### DIFF
--- a/internal/auth/plugin_provider.go
+++ b/internal/auth/plugin_provider.go
@@ -279,7 +279,7 @@ func (p *PluginProvider) autoProvisionUser(
 }
 
 func randomPluginOnlyPassword() (string, error) {
-	buf := make([]byte, 32)
+	buf := make([]byte, 24)
 	if _, err := rand.Read(buf); err != nil {
 		return "", err
 	}
--- a/internal/auth/plugin_provider_test.go
+++ b/internal/auth/plugin_provider_test.go
@@ -1,0 +1,21 @@
+package auth
+
+import (
+	"testing"
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+func TestRandomPluginOnlyPasswordFitsBcryptLimit(t *testing.T) {
+	password, err := randomPluginOnlyPassword()
+	if err != nil {
+		t.Fatalf("randomPluginOnlyPassword() error = %v", err)
+	}
+
+	if len(password) > 72 {
+		t.Fatalf("password length = %d, want <= 72", len(password))
+	}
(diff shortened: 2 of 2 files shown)


thezak48:
I have tested and this fixes #31

PR #36, opened 2026-06-01
fix(metadata): size merge-step args to the placeholders each statement binds

fluxis:
## Problem
`canonicalizeMediaItemReferencesTx` ran every merge step with both `$1` (source) and `$2` (canonical), regardless of how many placeholders the step's SQL actually bound. The second step, `drop source provider ids` (`DELETE ... WHERE content_id = $1`), binds only `$1`, so passing two args made pgx reject the `Exec` under the default `QueryExecModeCacheStatement` with `mismatched param and argument count` (`extended_query_builder.go:35` enforces `len(ParamOIDs) == len(args)`). That aborted the whole merge transaction, so any item that got past the "refusing to merge" guard never de-duplicated.

Observed in production logs as repeated `provider-id drift row repair failed ... drop source provider ids: mismatched param and argument count`.

## Approach
- Extract the merge-step list to a package var.
- Derive each step's args from the highest positional placeholder its SQL binds (`maxPlaceholder`) rather than a brittle `strings.Contains(sql, "$2")` substring check that would also match `$20`/`$25` and silently mis-size a future step.
- `mergeStepArgs` panics on an unsupported placeholder count, caught by `TestMergeStepPlaceholdersAreBounded` (asserts every step binds `$1` and at most `$2`), so a malformed step fails loud in tests rather than at runtime.

## Risks / follow-up
Low. Pure refactor of arg-passing; all 61 step SQL bodies are unchanged. This branch enables the merge/canonicalization path to commit for the first time in production — a separate PR (record-stale-IDs-against-canonical-item) hardens the stale-ID follow-up that interacts with it.

## Verification
`go test ./internal/metadata/` and `go vet ./internal/metadata/` pass. (Repo has no DB integration harness; the pgx arity behavior is verified against `pgx@v5.9.2` source.)

## AI-use disclosure
Implemented with Claude Code (Opus 4.8), TDD, after a multi-agent review of the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed database operation failures that could occur during media item data synchronization and merge operations.

* **Tests**
  * Added comprehensive test coverage for database integrity validation and data consistency checks.

### DIFF
--- a/internal/metadata/provider_id_integrity.go
+++ b/internal/metadata/provider_id_integrity.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


fluxis:
@coderabbitai review

PR #43, opened 2026-06-02
feat(autoscan): periodic Radarr/Sonarr import + rename rescans

fluxis:
## Summary

Adds **Autoscan**: a periodic poller that watches each configured Radarr/Sonarr instance's history for changes and triggers targeted Silo library rescans, so newly-imported or renamed files appear without waiting for a full scan.

The feature reuses the existing `request_integrations` rows as its arr sources (no second place to configure server URLs/keys) and is **off by default** — operators opt in per instance from the Requests admin surface.

Three layers, built bottom-up:

1. **Polling core** — every N minutes, for each enabled source, read `/api/v3/history/since`, map arr-side paths to Silo paths via per-instance path rewrites, dedup to parent folders, and enqueue scans through the existing `scantrigger` → `scanqueue` pipeline. Redis-backed debounce suppresses repeat scans of the same `(folder, path)` within a TTL.
2. **Rewrite-sync** — a convenience that queries an instance's arr root folders and suffix-matches each to a Silo media folder to propose path rewrites. Additive and manual-preserving: synced rows never overwrite operator-entered rewrites; the operator stays the source of truth.
3. **Rename-handling** — history polling also surfaces `episodeFileRenamed` / `movieFileRenamed` events (both the new path and the old `sourcePath`), since a rename moves a file with no import event and can cross folders.

## Why this approach

- **Reuses request integrations as sources.** Arr base URL + Fernet-encrypted API key already live in `request_integrations`; Autoscan references them by FK (`ON DELETE CASCADE`) instead of duplicating credential storage.
- **Targeted scans, not full rescans.** Resolving each changed path to its owning `MediaFolder` and enqueueing a subtree/file scan keeps the feature cheap under load — consistent with the repo's performance-first priority.
- **Generic, no deployment constants.** Path rewrites and suffix-matching are driven entirely by what each arr instance and the Silo catalog report at runtime; nothing about any specific host topology is baked in.

## Schema

`migrations/171_autoscan.{up,down}.sql` (no collision — `main` tops out at 170):

- `autoscan_settings` — singleton row (`id boolean PRIMARY KEY CHECK (id)`): `enabled`, `poll_interval_minutes` (CHECK > 0), `debounce_seconds` (CHECK >= 0). Seeded disabled.
- `autoscan_sources` — `integration_id` PK/FK → `request_integrations(id)` `ON DELETE CASCADE`; `enabled`, `path_rewrites jsonb`, `last_poll_at`.

## API (admin-gated, under `/admin`)

| Method | Path | Purpose |
|--------|------|---------|
| GET    | `/admin/autoscan/settings` | read global settings |
| PUT    | `/admin/autoscan/settings` | update settings |
| GET    | `/admin/autoscan/sources` | list sources (joined to integrations) |
| PUT    | `/admin/autoscan/sources/{id}` | upsert per-source enable + rewrites |
| POST   | `/admin/autoscan/trigger` | run one poll cycle now (async, 202) |
| GET    | `/admin/autoscan/status` | poll/interval status |
| GET    | `/admin/autoscan/sources/{id}/rewrite-suggestions` | suggest rewrites from arr root folders |

## Frontend

New **Autoscan** tab in the Requests admin page (`AdminRequests.tsx`): global enable + interval/debounce, per-source enable, and a collapsible path-rewrite editor with a "Sync rewrites" action that previews proposed / unmatched / ambiguous / already-covered matches before applying. Types + React Query hooks in `web/src/api/types.ts` and `web/src/hooks/queries/useAutoscan.ts`.

## Testing

- Go unit tests across `internal/autoscan`: poll cycle (dedup, distinct-paths-same-folder, disabled no-op, source-failure-doesn't-advance, release-claim-on-enqueue-failure), history parsing (imports + renames, ignores unrelated events), suffix-match suggester (unique/ambiguous/covered/no-op/normalization), rewrite + dedupe helpers, and handler tests (`FK → 404`, trigger dispatch).
- **Full Go suite green — 59 packages, 0 failures, including `-race`.** Note: the autoscan handler tests live in `internal/api/handlers`, which links libvips via CGO; run tests in an environment with `libvips-dev` (CI / the Docker `golang:1.26` build stage). A bare host without it reports `[build failed]` for those packages and skips their tests.
- `go vet` + `gofmt` clean.

## Risks / follow-ups

- Poll window is bounded (overlap buffer + 24h max-lookback floor) so a long outage can't produce an oversized history response; worst case is a missed pre-window change, recoverable by a manual scan.
- Client repos (`silo-android`, `silo-apple`) need no changes — admin-only server feature, no client-visible API surface.

## Note on the diff

This branch was cut before the vitest-4 hotfix (#42) merged, so the diff includes one unrelated file, `web/src/lib/recipes.test.ts`. That is the exact change in #42; once #42 merges and this branch is rebased onto `main`, git drops it as a duplicate. It does not double-apply.

## Release Notes

* **New Features**
  * **Autoscan**: Automatically triggers library scans when new media is detected from enabled integrations
  * Added Autoscan admin tab with global settings (enable/disable, poll interval, debounce), per-source configuration, manual trigger, and status display
  * **Path rewrite suggestions** to help map import paths to media folders

### DIFF
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/Silo-Server/silo-server/internal/api"
 	"github.com/Silo-Server/silo-server/internal/api/handlers"
 	"github.com/Silo-Server/silo-server/internal/auth"
+	"github.com/Silo-Server/silo-server/internal/autoscan"
 	"github.com/Silo-Server/silo-server/internal/cache"
 	"github.com/Silo-Server/silo-server/internal/catalog"
 	"github.com/Silo-Server/silo-server/internal/catalogseed"
@@ -73,6 +74,7 @@ import (
 	"github.com/Silo-Server/silo-server/internal/s3client"
 	"github.com/Silo-Server/silo-server/internal/scanner"
 	"github.com/Silo-Server/silo-server/internal/scanqueue"
+	"github.com/Silo-Server/silo-server/internal/scantrigger"
 	"github.com/Silo-Server/silo-server/internal/sections"
 	"github.com/Silo-Server/silo-server/internal/server"
 	"github.com/Silo-Server/silo-server/internal/subtitles"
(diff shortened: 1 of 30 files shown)


fluxis:
Pulling this down to adjust course on the approach — will reopen or supersede once revised. Branch and commits preserved.

PR #49, opened 2026-06-05
fix(jellycompat): fill large browse pages

Quick104:
## Summary

Fixes UHF-style Jellyfin compatibility `/Items` browse requests that ask for large pages and deep `StartIndex` offsets, such as `Limit=1000&StartIndex=1925000&SortBy=PremiereDate`.

## Root Cause

The Jellyfin compatibility layer passed the client limit into catalog browse, but catalog silently capped pages at 100. That under-filled large Jellyfin pages and forced clients to do much more pagination work. The movie `PremiereDate` sort also used a text `COALESCE(...)` expression, which prevented the large-offset movie browse path from using a release-date btree index effectively.

## Changes

- Propagate `EnableTotalRecordCount` into the direct Jellyfin browse adapter.
- Let Jellycompat explicitly request up to 1000 catalog rows for compatibility browse pages while preserving the default 100-row cap for native catalog callers.
- Keep exact totals out of the sorted page query; exact counts use the separate count query path or are inferred when safe.
- Make movie-only `PremiereDate` browse order by `mi.release_date` directly.
- Add `idx_media_items_movie_release_date_desc_content` for movie release-date deep paging.
- Add regression coverage for large compatibility pages, `include_total=false`, caller-specific catalog caps, and the movie-only release-date SQL shape.

## Validation

- `go test ./internal/jellycompat ./internal/catalog`

## AI Use

Implemented with AI assistance.

## Release Notes

* **Performance Improvements**
  * Added response compression for API payloads
  * Optimized movie browse queries with efficient release date sorting
  * Implemented batch image resolution for faster content loading

* **API Enhancements**
  * Browse pagination now supports configurable maximum page limits
  * Improved total record count reporting with optional calculation modes

### DIFF
--- a/internal/catalog/browse.go
+++ b/internal/catalog/browse.go
@@ -33,6 +33,7 @@ type BrowseFilters struct {
 	Sort               string   // created_at, release_date, rating_imdb, rating_tmdb, year, sort_title, recently_added
 	Order              string   // asc, desc
 	Limit              int
+	MaxLimit           int // optional caller-specific cap; zero keeps the default browse cap
 	Offset             int
 	SnapshotAt         *time.Time // pagination fence: exclude items created after this timestamp
 }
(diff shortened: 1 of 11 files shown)


Quick104:
@coderabbit

PR #54, opened 2026-06-06
feat(audiobooks): native API and player MVP

RXWatcher:
Second PR in the replacement audiobook stack.

Base: `stack/audiobooks-foundation`

Scope:
- Native `/api/v1/audiobooks` list/detail/progress endpoints
- Frontend audiobook query hooks and types
- `/audiobooks` route, sidebar entry, detail page, chapter list, and basic HTML5 audio player

Intentionally out of scope:
- Audiobookshelf compatibility listener/API
- ABS collections/playlists/RSS/smart collections
- Later player redesign/polish

Verification performed on stack tip before opening:
- `go test ./...`
- `cd web && pnpm run build`

### DIFF
--- a/docs/superpowers/plans/2026-05-24-audiobooks-absorption-3-api-and-frontend.md
+++ b/docs/superpowers/plans/2026-05-24-audiobooks-absorption-3-api-and-frontend.md
@@ -1,0 +1,326 @@
+# Audiobooks Absorption — Sub-plan 3: Silo-native API + Frontend (MVP)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development to implement task-by-task. Steps use checkbox (`- [ ]`) syntax.
+
+**Goal:** Land the minimum viable silo-native audiobooks UI: backend REST endpoints + React pages so a user can browse audiobook libraries scanned by sub-plan 2, see book details with chapters, play with progress saved. Author/series index pages, smart collections, share links, and other features from the spec's "nice-to-haves" list are deferred to a future sub-plan if the demand surfaces.
+
+**Architecture:** Three new Go handler files under `internal/api/handlers/audiobooks_*.go` for list/detail/progress. New `web/src/pages/audiobooks/` directory with three pages: Library, Detail, and an inline player. Hooks under `web/src/hooks/audiobooks/`. The streaming path reuses silo's existing `/api/v1/stream/{session_id}` endpoint — no new transcode code.
+
+**Tech stack:** Go + chi router + pgx for backend; React 19 + TanStack Query + radix-ui + tailwind for frontend.
+
+**Source spec:** `docs/superpowers/specs/2026-05-24-audiobooks-absorption-design.md`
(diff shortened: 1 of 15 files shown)


Quick104:
### Code review

Found 4 issues:

1. Player pauses and re-posts progress on every re-render, making playback unusable. `report` is memoized over the whole `reportProgress` mutation object, but `useReportAudiobookProgress()` (TanStack `useMutation`) returns a new object each render, so `report` is recreated every render. The "report and pause on unmount" effect (`}, [report]`) therefore runs its cleanup on every render, calling `audio.pause()` and posting progress on each `timeupdate` tick. Depend on `reportProgress.mutate` instead.

https://github.com/Silo-Server/silo-server/blob/fe9162070da49a922b198286fa4b5fed83654656/web/src/pages/audiobooks/AudiobookPlayer.tsx#L50-L62

2. All audiobook covers are broken: `poster_url` returns the raw storage key. The list and detail handlers assign `item.PosterPath` (an S3 object key) straight to `poster_url`, while every other handler runs the path through `presignURL(...)`. The frontend renders `` directly, so no image loads. Same pattern in `audiobooks_detail.go:60`.

https://github.com/Silo-Server/silo-server/blob/fe9162070da49a922b198286fa4b5fed83654656/internal/api/handlers/audiobooks_list.go#L69-L71

3. Detail endpoint returns audiobooks from libraries the caller cannot access. `HandleGetAudiobook` calls `GetByID` then only checks `item.Type`; it never calls `EnsureAccessible(ctx, contentID, requestAccessFilter(r))` the way sibling handlers (favorites, ratings, catalog detail) do. Any authenticated user can fetch metadata and their progress for any audiobook id, bypassing library/parental scope. The list endpoint applies the access filter; detail does not.

https://github.com/Silo-Server/silo-server/blob/fe9162070da49a922b198286fa4b5fed83654656/internal/api/handlers/audiobooks_detail.go#L25-L36

4. Audiobooks can never be marked completed. The progress handler passes `duration=0` to `SetProgress`, but both store implementations gate completion on `duration > 0`, so `completed` never becomes true. `AudiobookDetail` keys Resume vs Play on `progress.completed`, so a finished book always shows "Resume" and never clears from continue-listening. The comment claiming "position-only heuristics" does not match `SetProgress`'s behavior.

https://github.com/Silo-Server/silo-server/blob/fe9162070da49a922b198286fa4b5fed83654656/internal/api/handlers/audiobooks_progress.go#L58-L63

RXWatcher:
Verified the top-level review findings against the current branch tip `98d1b388`:

- Player progress callback now destructures `mutate` and depends on the stable mutation function.
- List/detail poster URLs now go through `presignAudiobookPoster(...)`.
- Detail access now uses `GetByIDsWithAccess(..., requestAccessFilter(r))`.
- Progress reporting now derives duration from the audiobook file(s) before calling `SetProgress`.

No unresolved review threads are present on this PR.

Quick104:
Superseded by #73, which replaces the native /audiobooks API/UI direction with first-class catalog/item/progress integration.

PR #65, opened 2026-06-07
fix(jellycompat): enforce allow_4k_transcode

### LINKED ISSUE #64: jellycompat ignores allow_4k_transcode
The `allow_4k_transcode` setting works on the native playback API, but the Jellyfin compat layer never reads it. Any Jellyfin client (Infuse, Findroid, Streamyfin, etc.) whose device profile can't direct play a 4K file will negotiate a transcode and jellycompat just starts ffmpeg. So the setting only covers half the playback surface.

Checked against `ad2183a` (current `ghcr.io/silo-server/silo-server:latest`).

## Repro

1. Leave `allow_4k_transcode` unset (or false).
2. Auth against the compat API and send `PlaybackInfo` for a 2160p item with a device profile that can't direct play it (h264-only is enough).
3. Fetch the returned `TranscodingUrl`.

I did this with curl against my own server on a movie that has both a 2160p DV and a 1080p version. PlaybackInfo happily offers the 4K source for transcoding:

```
Remux-1080p 8-bit AVC      | direct:false stream:false transcode:true
Remux-2160p HEVC DV HDR10  | direct:false stream:false transcode:true   <- expected this to be filtered
```

Requesting master.m3u8 for the 2160p MediaSourceId returns 200 and this starts:

```
ffmpeg ... -i /mnt/.../Furiosa A Mad Max Saga (2024) [Remux-2160p HEVC DV HDR10 10-bit TrueHD Atmos 7.1].mkv
  -c:v h264_qsv -preset veryfast -global_quality 23
  -vf scale_vaapi=format=nv12,hwmap=derive_device=qsv,format=qsv
  -c:a aac -b:a 192k -ac 2 -f hls ...
```

For comparison, the native player on the same title does the right thing: force a video transcode and the session quietly switches to the 1080p file.

This also reproduces with a stock client, no crafted profile needed: jellyfin-web in a mobile browser that doesn't advertise HEVC kicked off an identical full-resolution `h264_qsv` encode of a 2160p HEVC HDR episode that only exists in 4K (so no alternate version to fall back to — it just encoded it). Desktop Chrome and Firefox mask the bug nowadays since they decode HEVC natively, so those stay on video copy with only the audio converted.

Two more things I noticed along the way, possibly worth their own issue:

- the scale filter only sets the pixel format, no w/h, so it's encoding at full 2160p
- there's no `-b:v` even though the device profile sent `MaxStreamingBitrate: 20000000`. Visible from a real client too: moving the quality slider in jellyfin-web re-runs PlaybackInfo but the stream just keeps going at full rate, so the quality selector effectively does nothing against the compat API

## Where it diverges

Native enforcement is in two places:

- `internal/api/handlers/playback.go:624` reads the setting into `playback.AdminSettings`, and `internal/playback/resolver.go:179` skips 2160p files during version selection when it's off
- `internal/api/handlers/playback.go:1989` guards the transcode start: 4K source gets swapped to an alternate non-4K version (`findAlternateFile`) or the request fails with 422. Video copy is exempt, which makes sense since nothing gets encoded.

The compat layer has neither — `grep -r allow_4k internal/jellycompat/` comes back empty. `buildPlaybackSource` (`handlers_playback.go:533`) computes `supportsTranscoding` purely from the client profile and request flags, and `ensureTranscodeSession` (`streams.go:809`) calls `playback.StartTranscode` directly with `TargetCodecVideo: "h264"` hardcoded, so the native handler's guard is never in the path.

## Possible fix

`jellycompat.PlaybackHandler` already has a `SettingsRepo` (wired in `router.go:96`), so no new plumbing needed:

1. In `buildPlaybackSource`, force `SupportsTranscoding = false` when the version is 2160p and the setting is off. PlaybackInfo already returns all file versions, so clients fall back to a non-4K source or direct play, same as the native resolver.
2. Guard the session start in `ensureTranscodeSession` too, same as native (swap to alternate version, else fail), so a hand-built master.m3u8 request can't get around the negotiation.
3. Keep the video-copy exemption — remuxing 4K with audio transcode is cheap and the native guard allows it.

Can put together a PR if this looks right.

## Environment

- `ghcr.io/silo-server/silo-server:latest`, build `ad2183a`
- Docker compose, single node, QSV

d3v1l1989:
Fixes #64.

The jellycompat layer never consulted `allow_4k_transcode`, so any Jellyfin client whose device profile couldn't direct play a 4K file got a full video encode of the 4K source, regardless of the setting (full background and repro in #64).

## What this does

Negotiation (`buildPlaybackSource`): when the setting is off and the version is 4K, `SupportsTranscoding` is forced to false so PlaybackInfo doesn't offer a `TranscodingUrl` for that source. Clients then fall back to a lower-resolution version or direct play on their own, which matches what the native resolver does by skipping 4K files during version selection.

Enforcement (`ensureTranscodeSession`, `startRemoteTranscode`): same check before starting ffmpeg locally or on a transcode node, returning an error so a hand-built `master.m3u8` request can't get around the negotiation. The manifest handlers map it to a 403 instead of a generic 500.

Deliberately exempt, matching the native guard's video-copy exemption:

- `TranscodeAudio` sessions (video copy + audio to AAC) stay allowed for 4K. This is the common browser case (HEVC decodes natively, EAC3/TrueHD doesn't) and blocking it would make 4K with surround audio unplayable in browsers.
- Seek restarts need no extra guard: `TranscodeSession.Restart` reuses the original opts, so a copy session can't escalate to an encode.

Default matches native: nil/unset settings repo means 4K video transcodes are denied.

One deliberate difference: the resolution check follows the native resolver's `is4K` (`>= 2160p`, so 4320p is covered too) rather than the native start guard's exact `"2160p"` comparison. Blocking 4K but allowing 8K didn't seem intended.

## Testing

- Unit tests for the negotiation gate (blocked / allowed / audio-copy exempt / non-4K untouched), the settings read, and both start guards. `go test` passes for the jellycompat and playback packages, golangci-lint clean on the touched package.
- Tested live on my server (the #64 setup): with the patch, PlaybackInfo returns the 2160p source with `SupportsTranscoding: false` for an h264-only profile, requesting its master.m3u8 returns 403, real clients fall back to the 1080p version on their own, and 4K video-copy remux sessions (Chrome/Firefox with surround audio) keep working.

Per the contributing guidelines: implemented with AI assistance (same as the investigation in #64). I've read the diff and verified the behavior on a live deployment myself.

## Summary by CodeRabbit

* **New Features**
  * Added server-side control to restrict 4K video transcoding when the `allow_4k_transcode` setting is disabled.
  * System now returns a 403 Forbidden response when 4K video transcoding is attempted but disallowed by policy.
  * Audio-only transcoding remains available regardless of 4K video transcoding settings.

### DIFF
--- a/internal/jellycompat/audio_selection_test.go
+++ b/internal/jellycompat/audio_selection_test.go
@@ -142,6 +142,7 @@ func TestBuildPlaybackSource_SeedsRequestedAudioStreamIndex(t *testing.T) {
 		version,
 		DeviceProfile{},
 		playbackInfoRequest{AudioStreamIndex: compatIntValuePtr(requestedAudioStreamIndex)},
+		true,
 	)
 
 	if source.SelectedAudioStreamIndex == nil {
--- a/internal/jellycompat/handlers_playback.go
+++ b/internal/jellycompat/handlers_playback.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 
+	"github.com/Silo-Server/silo-server/internal/access"
 	"github.com/Silo-Server/silo-server/internal/catalog"
 	"github.com/Silo-Server/silo-server/internal/config"
 	"github.com/Silo-Server/silo-server/internal/models"
@@ -152,6 +153,22 @@ func (h *PlaybackHandler) playbackThresholds(ctx context.Context) userstore.Prog
 		}
 	}
 	return t
+}
+
+var errTranscode4KDisallowed = errors.New("4k video transcode disallowed by server settings")
+
+// allow4KVideoTranscode reads the allow_4k_transcode server setting,
+// defaulting to deny like the native playback handler.
+func (h *PlaybackHandler) allow4KVideoTranscode(ctx context.Context) bool {
+	if h.SettingsRepo == nil {
+		return false
+	}
+	v, _ := h.SettingsRepo.Get(ctx, "allow_4k_transcode")
+	return v == "true"
+}
+
+func is4KResolution(res string) bool {
+	return access.CompareQuality(res, "2160p") >= 0
 }
 
 // NewPlaybackHandler creates a playback handler.
@@ -300,6 +317,9 @@ func (h *PlaybackHandler) startRemoteTranscode(
 	initialSeekSeconds float64,
 	transcodeNodeURL string,
 ) error {
+	if !source.TranscodeAudio && is4KResolution(source.Version.Resolution) && !h.allow4KVideoTranscode(ctx) {
+		return errTranscode4KDisallowed
+	}
 	if d := float64(source.Version.Duration); d > 0 && initialSeekSeconds > d {
 		initialSeekSeconds = d
 	}
@@ -416,8 +436,9 @@ func (h *PlaybackHandler) HandlePlaybackInfo(w http.ResponseWriter, r *http.Requ
 	sources := make([]PlaybackMediaSource, 0, len(detail.Versions))
 	sourceDTOs := make([]mediaSourceDTO, 0, len(detail.Versions))
 
+	allow4KTranscode := h.allow4KVideoTranscode(r.Context())
 	for _, version := range detail.Versions {
-		source := h.buildPlaybackSource(routeItemID, playSessionID, version, profile, req)
+		source := h.buildPlaybackSource(routeItemID, playSessionID, version, profile, req, allow4KTranscode)
 		if req.MediaSourceID != "" && source.ID != req.MediaSourceID {
 			continue
 		}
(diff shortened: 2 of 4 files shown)


d3v1l1989:
Took the test suggestion — the stub can inject a read error now and there's a case pinning down deny-on-error (255473e).

Leaving the `== "true"` comparison as is, intentionally: the native handler reads the same key with exact equality (`internal/api/handlers/playback.go:624`), and the whole point of this PR is keeping the two layers in agreement. Lenient parsing on one side would let a value like "TRUE" enable 4K transcodes in compat while native still denies — the same class of inconsistency this fixes. Happy to move both to ParseBool in a follow-up if preferred.

PR #80, opened 2026-06-08
fix(catalog): qualify content_id in discovery queries under library join

d3v1l1989:
## Problem

Two catalog issues were failing in nearby paths:

1. Home/library discovery sections backed by `DiscoveryRepository` fail at runtime for library-scoped viewers with:

```text
ERROR: column reference "content_id" is ambiguous (SQLSTATE 42702)
```

This affects the `hidden_gems`, `critically_acclaimed`, and `forgotten_favorites` section recipes when access joins `media_item_libraries`.

2. The late orphan cleanup queries used after folder membership deletion did not reuse the full provisional media-item predicate, so the package-level regression test failed and the delete path could omit the status guard.

## Root cause

`ListByRatingThreshold`, `ListUnplayedHighRated`, and `ListForgottenFavorites` conditionally join `media_item_libraries mil` to enforce library access. `media_item_libraries` also has a `content_id` column, but the queries selected from the unqualified `itemColumns` constant.

For late orphan cleanup, the collect/delete SQL fragments used only `orphanedMediaItemSafetyConditions` instead of `orphanedProvisionalMediaItemConditions`, dropping the `mi.status IN (pending, unmatched, ambiguous)` guard.

## Fix

- Select `qualifiedItemColumns("mi")` in the discovery query builders.
- Reuse `orphanedProvisionalMediaItemConditions` in both late orphan collect/delete queries.

## Testing

- [x] go test ./internal/catalog -run 'Discovery|RatingThreshold|Unplayed|ForgottenFavorites|TestLateOrphanCleanupUsesProvisionalSafetyPredicate' -count=1
- [x] go test ./internal/catalog -count=1
- [x] git diff --check

### DIFF
--- a/internal/catalog/discovery_repo.go
+++ b/internal/catalog/discovery_repo.go
@@ -36,10 +36,10 @@ type RatingFilter struct {
 	Filter AccessFilter
 }
 
-// ListByRatingThreshold returns media items whose rating_imdb is >= f.Min,
-// ordered by rating_imdb DESC NULLS LAST.  Items without an IMDb rating are
-// always excluded.
-func (r *DiscoveryRepository) ListByRatingThreshold(ctx context.Context, f RatingFilter) ([]*models.MediaItem, error) {
+// buildRatingThresholdQuery builds the SQL statement and bind args for ListByRatingThreshold.
+// It returns an empty query string when access rules exclude every
+// library, signalling the caller to skip the query and return no rows.
+func buildRatingThresholdQuery(f RatingFilter) (string, []any) {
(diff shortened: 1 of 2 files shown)


Quick104:
Closing this in favor of #88.

#88 keeps the useful discovery query-builder extraction and column qualification from this PR, then extends the fix to use the existing `buildLibraryScopeJoin` semi-join shape so library-scoped discovery queries avoid both `content_id` ambiguity and multi-library row fanout. It also includes the missing late-orphan predicate fix described here.

PR #87, opened 2026-06-08
Fix late orphan provisional cleanup guard

RXWatcher:
## Summary
- Reuse the full provisional media-item predicate for late orphan collect/delete queries after folder membership deletion.
- Keeps matched media items from being eligible for the provisional-only late orphan cleanup path.

## Test Plan
- [x] go test ./internal/catalog -run TestLateOrphanCleanupUsesProvisionalSafetyPredicate -count=1 -v
- [x] go test ./internal/catalog -count=1
- [x] git diff --check


RXWatcher:
Rolled into #80 so the catalog fixes are reviewed together.

PR #90, opened 2026-06-08
fix(scanner): skip misplaced TV episodes in movie libraries

d3v1l1989:
## Problem

In a **movie-type library**, a TV show laid out as `Show/Season NN/SxxExx.mkv` (for example a fan "supercuts" pack dropped into an anime-movies folder) is turned into **one bogus movie item per episode**, every one titled after the season folder — e.g. 34 separate items all named **"Season 01"**. They pile up on *Recently Added*, never match a movie provider, and clicking one opens a "movie" that is really an episode.

Root cause: the movie skeleton path treats every file as a movie. A `Season NN` folder even parses a bogus provider id (the season number becomes `tmdb="01"`), so the items aren't even caught as ID-less.

## Fix

- **`naming.IsMisplacedSeriesFile`** — detects a file that has an `SxxExx` name **and** lives inside an explicit `Season NN`/`Specials` (or `Extras`) directory. Deliberately strict: it requires *both* signals so a movie whose release name merely contains an `SxxExx` substring (e.g. `...S01E43...` inside a `Title (Year)/` folder) is **not** affected.
- **`createOrFindSkeleton`** — in a strict movie/`movies` library, such a file is recorded as a skipped root (`reason=series_in_movie_library`) and **no item is created**. The check runs on the structural signal alone, *not* gated on missing provider ids (a `Season NN` folder otherwise yields the bogus `tmdb="01"`). `mixed` libraries are left untouched since they legitimately host series.
- **Movie match-queue worker** — dequeues files the skeleton step deliberately skips (`ItemStatus=="skipped"`) instead of erroring with *"movie queue claimed without a content id"* and retrying forever.

## Verification

- New unit test `internal/naming/misplaced_series_test.go` covers the misplaced-TV case and the legit-`SxxExx`-movie case; `go test ./internal/naming/ ./internal/metadata/` passes.
- Deployed to a live instance: the 34 junk "Season 01" items disappeared and **do not return on rescan**; a `skipped_media_roots` row is recorded; the queue drains cleanly with no retry/error spam; and the legitimate movies in the same library that merely have `SxxExx` in their filename (inside `Title (Year)/` folders) are untouched and still matched.

---
*Built with AI assistance (per CONTRIBUTING) — investigated against the live `naming`/scanner code, verified with unit tests and on a production deploy before submitting.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Added detection for TV series files incorrectly placed in movie library locations. When detected, these files are now automatically skipped from processing instead of causing errors, and their locations are recorded for user reference.

* **Tests**
  * Added test coverage for TV episode file detection including various naming patterns and directory structures.

### DIFF
--- a/internal/metadata/movie_match_queue_repo.go
+++ b/internal/metadata/movie_match_queue_repo.go
@@ -43,6 +43,31 @@ func requirePositiveMovieQueueID(label string, value int) error {
 	return nil
 }
 
+// movieQueueFileEligibleCond is the predicate deciding whether a media file
+// belongs in the movie match queue. Queries embedding it must alias
+// media_files as mf, media_folders as folders, and media_items as mi.
+//
+// Files beneath a root skipped as misplaced series are excluded durably:
+// their content_id is never set, so without the exclusion every library sync
+// would re-enqueue them only for the worker to skip them again.
+const movieQueueFileEligibleCond = `folders.enabled = true
+	  AND (
+		lower(trim(folders.type)) IN ('movie', 'movies') OR
+		(lower(trim(folders.type)) = 'mixed' AND lower(trim(mf.base_type)) = 'movie')
+	  )
+	  AND mf.missing_since IS NULL
+	  AND (
+		mf.content_id IS NULL OR mf.content_id = '' OR
+		lower(trim(COALESCE(mi.status, ''))) IN ('pending', 'unmatched', 'ambiguous')
+	  )
+	  AND NOT EXISTS (
+		SELECT 1
+		FROM skipped_media_roots sr
+		WHERE sr.media_folder_id = mf.media_folder_id
+		  AND sr.reason = '` + skippedReasonSeriesInMovieLibrary + `'
+		  AND strpos(mf.file_path, sr.root_path || '/') = 1
+	  )`
+
 func (r *MovieMatchQueueRepository) EnqueueMovieFile(ctx context.Context, fileID int) error {
 	if err := r.requireConfigured(); err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


d3v1l1989:
Thanks — addressed the nitpick: added an `Extras` positive-case row to `TestIsMisplacedSeriesFile` so the `extras?` branch of `inferSpecialsDirRe` is locked in. `go test ./internal/naming/ ./internal/metadata/` passes.

Quick104:
Reviewed and pushed 19083b25 addressing the findings:

**Endless re-enqueue (main fix).** A skipped file's `content_id` is never set, and the movie queue eligibility predicate (`mf.content_id IS NULL OR ...`) knew nothing about skipped roots — so every library sync re-enqueued the file just for the worker to skip, upsert, and dequeue it again, forever. The queue predicates now exclude files beneath a `series_in_movie_library` skipped root, making the skip durable: the `skipped_media_roots` row is the single source of truth, and deleting it (e.g. from admin) makes the files eligible for matching again. Your worker-side dequeue stays — it drains rows claimed before the root was recorded.

While touching this, the eligibility predicate (previously duplicated verbatim in eight queries) is extracted into one shared `movieQueueFileEligibleCond` fragment.

**file_count accuracy.** Each per-file `Upsert(FileCount: 1)` overwrote the row (`file_count = EXCLUDED.file_count`), so a 34-episode pack reported `file_count=1`. A new `UpsertObservedFile` derives the count from `media_files` under the root, idempotently — this also fixes the same undercount in the pre-existing `missing_folder_ids` path.

**Cleanups.** The two near-identical Upsert+log blocks in `createOrFindSkeleton` are folded into a `recordSkippedRoot` helper, `libraryType` is normalized once and reused, the reason strings are named constants, and a no-op `filepath.Base` on already-split segments in `IsMisplacedSeriesFile` is removed.

Verified: `go build ./...` and `go test ./internal/naming/ ./internal/metadata/ ./internal/libraryingest/ ./internal/adminjob/` pass.

Issue #45, opened 2026-06-03
Encrypt arr integration credentials at rest (Requests + Autoscan)


fluxis:
### Summary
`silo-server` stores Sonarr/Radarr API keys **plaintext** in the database:
- `request_integrations.api_key_ref` — the Requests subsystem writes the raw key directly (`internal/requests/repository.go`).
- `autoscan_connections.api_key_ref` — the Autoscan subsystem does the same (mirrors Requests).

The `SecretResolver` wired into both (`catalog.ServerSettingsRepo`) is a plaintext `server_settings` key/value store — there is no Fernet/AES encryption layer for credentials. (The only crypto in the repo is a purpose-built AES-GCM for OAuth completion tokens in `auth/oauth_store.go`, not a reusable credential store.)

There is also a **ref-vs-literal ambiguity**: `resolveAPIKey`/`ConnectionResolver.Resolve` treat `api_key_ref` as a secrets-store key, but fall through to using it **verbatim** when `secrets.Get` returns empty — so whether a stored value is a reference or a literal key depends on accidental `server_settings` key collisions.

### Why this is a standalone item
This is **repo-wide**, not specific to one feature — surfaced during the Autoscan review but present in Requests today. Fixing it properly means:
1. Introduce a real secret store in `silo-server` (Fernet/AES, `SECRET_KEY`-derived).
2. Migrate **both** `request_integrations` and `autoscan_connections` credentials onto it (encrypt-on-write, decrypt-on-resolve).
3. Remove the verbatim-fallback ambiguity.

It was deliberately scoped **out** of the Autoscan PR (#44) so that feature stays consistent with the existing pattern rather than introducing a one-off encryption scheme.

### Acceptance
- No plaintext arr API keys at rest (Requests + Autoscan).
- `api_key_ref` is unambiguously a reference (no literal fallback).
- Migration encrypts existing rows.

brettpetch:
May be worth moving S3 keys as well?

PR #95, opened 2026-06-08
feat(security): encrypt server-owned credentials at rest (#45)

Quick104:
## Problem

Third-party integration credentials and other server-owned secrets are stored in PostgreSQL as plaintext — Sonarr/Radarr API keys, S3 keys, watch-sync OAuth tokens, history-import tokens, the ABS signing key, and a dozen sensitive `server_settings`. A DB dump or compromise exposes all of them. The two arr `api_key_ref` columns were additionally resolved through a plaintext `server_settings` key/value store with an ambiguous ref-vs-literal fallback.

Implements the plan for [issue #45](https://github.com/Silo-Server/silo-server/issues/45).

## Approach

A new `internal/secret` package owns one primitive: **AES-256-GCM** with a random 12-byte nonce, a versioned `enc:v1:` envelope, and a data key **HKDF-SHA256-derived from a required `SECRET_KEY`** env var (so the key lives outside Postgres and secrets survive a full DB dump). Every ciphertext is GCM-bound to its row via **AAD** (`table:column:<pk>`, or `server_settings:<key>`), so a DB-write attacker can't transplant a credential blob between rows.

- **`SECRET_KEY` required at bootstrap** (fatals if absent/<32 chars); the cipher is constructed once and threaded as an explicit dependency — never a global.
- **`server_settings`:** an `EncryptedSettingsRepo` decorator transparently encrypts the audited `SensitiveSettingKeys` (now the single source of truth for both encryption *and* admin redaction). The config hot-reload watcher and watch-sync's raw settings reads decrypt too.
- **Arr keys** inline-encrypted; the ambiguous `SecretResolver` indirection deleted from `requests`/`autoscan`.
- **Per-table columns** encrypted with encrypt-on-write / decrypt-on-read and a row-bound AAD: subtitles (`api_key`/`password`), watch-sync (`access_token`/`refresh_token`), webhook-sync (`access_token`), history-import (admin + session tokens), and the jellycompat session's bridged Silo access/refresh tokens.
- **Idempotent startup backfill** (best-effort, primary-node gated) sweeps existing plaintext to ciphertext, with a resolve-then-encrypt path for the legacy arr references.

**No schema migration** — every target column is already `text` and now holds ciphertext; the backfill is pure Go. **No client/frontend changes** — API responses already mask these to `has_*` booleans.

## Read-path contract

Empty → empty; legacy plaintext passes through unchanged (backfill window); an `enc:v1:` value is decrypted and any failure (wrong key/tamper/truncation) is propagated, never swallowed or used as ciphertext.

## Risks / operations

- **`SECRET_KEY` must be set and backed up separately from DB dumps** (treat like a CA private key). Loss makes encrypted secrets unrecoverable. **Downgrade hazard:** an older binary would read `enc:v1:auth.jwt_secret` as a literal JWT secret. Full operator runbook in `docs/architecture/secret-encryption.md`.

## Out of scope (follow-ups)

- **Equality-looked-up secrets** (`api_keys.api_key`, `webhook_sync_connections.webhook_secret`, `jellycompat_sessions.token`, `watch_together_rooms.join_token`) — matched by exact value, so they need a deterministic **blind-index hash**, not randomized encryption.
- **`plugin_runtime_configs.config_value`** — cross-repo (`Silo`) + manifest-driven; needs a coordinated design.

## Testing

`go build ./...`, `go vet ./...`, gofmt clean, `make verify-local-paths`, and unit tests across `secret` (cipher round-trip/tamper/AAD/version/idempotency + backfill resolve-then-encrypt), the settings decorator, and all touched repos. No frontend changes. New code is lint-clean (pre-existing repo lint debt untouched).

## AI-use disclosure

Implemented by Claude Code (Opus 4.8) from the attached plan.

## Summary by CodeRabbit

* **New Features**
  * At-rest encryption for sensitive server credentials in the database with automatic, idempotent backfill on startup (best-effort, primary node only).

* **Documentation**
  * Added a detailed architecture guide covering the encryption design, envelope format, and operational guidance.

* **Chores**
  * A SECRET_KEY environment variable is now required at startup; see docs for generation, minimum length, and backup guidance.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -49,6 +49,16 @@ POSTGRES_SHM_SIZE=8gb
 # Optional NVIDIA override controls (used by docker-compose.nvidia.yml).
 # NVIDIA_GPU_COUNT=1
 
+# REQUIRED: master key for at-rest credential encryption. Silo refuses to start
+# without it. Integration API keys, S3 keys, and other server-owned secrets are
+# encrypted under a key derived from this value, so it must be at least 32
+# characters and kept secret. Generate one with:
+#   openssl rand -base64 48
+# Back it up SEPARATELY from your database dumps (treat it like a CA private
+# key): losing it makes every encrypted secret unrecoverable. See
+# docs/architecture/secret-encryption.md.
+# SECRET_KEY=replace-with-output-of-openssl-rand-base64-48
+
 # Run from source / advanced overrides
 # Only DATABASE_URL is required when running Silo outside the default docker compose stack.
 # DATABASE_URL=postgres://silo:password@localhost:5432/silo
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -77,6 +77,7 @@ import (
 	"github.com/Silo-Server/silo-server/internal/s3client"
 	"github.com/Silo-Server/silo-server/internal/scanner"
 	"github.com/Silo-Server/silo-server/internal/scanqueue"
+	"github.com/Silo-Server/silo-server/internal/secret"
 	"github.com/Silo-Server/silo-server/internal/sections"
 	"github.com/Silo-Server/silo-server/internal/server"
 	"github.com/Silo-Server/silo-server/internal/subtitles"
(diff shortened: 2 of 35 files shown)


Quick104:
@coderabbit

PR #98, opened 2026-06-08
feat(ebooks): extract local covers during scan

RXWatcher:
## Summary
- Extract EPUB manifest covers into ebook posters during scan.
- Use the first readable CBZ page as a poster.
- Pick up same-directory ebook sidecar covers like cover.jpg, folder.png, front.webp, poster.jpg, and thumbnail.png.

## Test Plan
- [x] go test ./internal/scanner -count=1

### DIFF
--- a/internal/catalog/item_repo.go
+++ b/internal/catalog/item_repo.go
@@ -70,12 +70,21 @@ const itemColumns = `content_id, type, title, sort_title, default_metadata_langu
 	content_rating, runtime, overview, tagline,
 	rating_imdb, rating_tmdb, rating_rt_critic, rating_rt_audience,
 	imdb_id, tmdb_id, tvdb_id,
-	poster_path, poster_thumbhash, backdrop_path, backdrop_thumbhash, logo_path,
-	metadata_s3_path, metadata_etag, season_count,
+	COALESCE(poster_path, ''), COALESCE(poster_thumbhash, ''), COALESCE(backdrop_path, ''), COALESCE(backdrop_thumbhash, ''), COALESCE(logo_path, ''),
+	COALESCE(metadata_s3_path, ''), COALESCE(metadata_etag, ''), season_count,
 	studios, networks, countries, keywords, original_language, release_date::text, first_air_date, last_air_date, air_time, air_timezone,
 	show_status,
 	matched_at, last_refreshed, refresh_failures,
 	episode_metadata_incomplete, episode_metadata_last_checked_at, locked_fields, status, created_at, updated_at`
+
+func qualifiedNullableStringColumn(alias, col string) string {
+	switch col {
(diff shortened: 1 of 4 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai review

Quick104:
Pushed 0ed64f49 addressing the code-review findings:

| # | Finding | Fix |
|---|---------|-----|
| 1 | One `cover.jpg` in a flat multi-book directory became every book's poster and blocked embedded covers | Generic sidecar names are only trusted when the directory contains a single ebook; an image named after the book file (`Book.jpg`) always belongs to it. One cover is applied per reconcile (sidecar wins over embedded). |
| 2 | Sibling files clobbered provider-matched items' curated title/year via the group-key merge branch | Matched items are joined without applying the sibling's embedded metadata. |
| 3 | Distinct books with colliding sparse metadata silently merged into one item | Group lookup now excludes items that already own a file of the same format — grouping merges *different formats* only. |
| 4 | New PDF string decoder mojibaked UTF-8 metadata (`Café` → `CafÃ©`) | UTF-8 (incl. BOM) is detected and preserved; Windows-1252 is now only the fallback for non-UTF-8 bytes. |
| 5 | `meta name="cover"` pointing at an XHTML cover page selected the XHTML doc and the book lost its cover | Manifest selection requires an image media-type and `properties="cover-image"` outranks the EPUB2 meta id. |
| 6 | Group-key scheme changed without a version bump, so pre-existing libraries never got sibling-format grouping | `ebookGroupKeyVersion = 2`; rows written under older versions skip the unchanged-file shortcut once and get re-keyed. |
| 7 | Hand-rolled 6-label charset table dropped metadata for windows-1251/koi8-r/etc.; FB2 had no charset reader at all | Both now decode through `x/net/html/charset.NewReaderLabel` (full IANA registry). |
| 8 | A replaced `cover.jpg`/fixed embedded cover could never refresh a stale poster | Posters under `local/ebooks/` are owned by the scan pipeline and refresh when the extracted bytes change (thumbhash compare); provider artwork still always wins. |
| 9 | Check-then-set poster write raced the enrichment sweep | Replaced with an atomic conditional `UPDATE ... WHERE poster IS NULL/empty/local-prefixed` (`ItemRepository.SetLocalPoster`). |
| 10 | CBZ covers picked by byte-order sort (`ch10/` before `ch2/`) | Natural ordering via an O(n) min-scan (also drops the full-page sort). |

Also folded in: `.fb2.zip` double extension now fully stripped from filename-derived titles/group keys, the group key is computed once per reconcile instead of three times, and the double `GetPosterPath` query per file is gone.

`go test ./internal/scanner ./internal/catalog ./internal/imagecache ./internal/ebooks ./internal/metadata -count=1` passes; `go vet` and `gofmt` clean on touched files.

PR #104, opened 2026-06-09
feat(requests): pluginize request fulfillment behind request_router.v1

fluxis:
## Summary

Pluginizes request fulfillment so the host no longer talks to Sonarr/Radarr (or any backend) directly. The host now **computes governance** (which qualities a requester is allowed) and **delegates whole-request fulfillment** to a `request_router.v1` plugin over gRPC, persisting the targets the plugin reports back.

- **Two-tier connection model** — `request_integrations` is generalized into generic host columns (`base_url`, `api_key_ref`, `enabled`, `capability_id`, `installation_id`, `supported_media_types`) plus a `plugin_config` JSON blob. Legacy per-backend columns dropped; `plugin_config` is the sole source of truth.
- **Host seam** — new `RequestRouterProvider` + typed `pluginhost` `RequestRouter` capability client/resolver. In-host Sonarr/Radarr fulfillment code removed. Fulfillment is pure-plugin (no in-host fallback; no plugin ⇒ approve fails "no fulfillment backend configured").
- **Schema-driven admin form** — one generic `SchemaForm` renderer drives any plugin's `admin_form` descriptor (sections, `show_when`, dynamic options, validation, multi-select, exclusive groups). The bespoke arr form is gone; `PluginConfigForm` was refactored onto `SchemaForm`.
- **Host owns the 4K-tier fact** — `FulfillRequest.qualities` carries typed `RequestedQuality{id, is4k}`; the host stamps `is4k`, plugins shed all tier-string literals.
- **At-rest credential cipher (#95 adoption)** — plugin API keys are encrypted via the new `secret.Cipher` (`enc:v1:` at rest in `request_integrations.api_key_ref`, decrypted on the read path); the old `SecretResolver` indirection is removed. Requires `SECRET_KEY` at boot.
- **Autoscan compatibility preserved** — autoscan still reuses the same `request_integrations` rows via `request_integration_id` (locked by a regression test + compile guard in `internal/api/autoscan_wiring_test.go`).

Pairs with the published `silo-plugin-sdk` v0.6.0 (`request_router` capability) and the first-party `silo-plugins-requests-arr` / `-seerr` plugin repos.

Spec: `docs/superpowers/specs/2026-06-07-requests-pluginization-design.md`
Plan: `docs/superpowers/plans/2026-06-07-requests-pluginization.md`
(plus follow-on specs/plans for the schema-driven form, typed 4K signal, SDK helpers, single-default enforcement, and the host-rebase + #95 credential adoption)

## Test Plan

- [x] `go test ./internal/requests/...` passes (in `golang:1.26 + libvips-dev`, built against published SDK v0.6.0)
- [x] Frontend `pnpm lint` — 0 errors; `prettier --check` clean on all branch-touched files
- [x] `make verify-local-paths` passes; no machine-local `replace` in `go.mod` (SDK pinned at v0.6.0)
- [x] Live e2e on a deployed instance: install arr + seerr plugins → add connections via the schema-driven form → options load from Radarr/Sonarr → save → approve request → targets queued → reconcile advances → "no backend" block when uninstalled → autoscan reuse still polls
- [ ] Reviewer: confirm migrations `20260607131954`, `20260607211831`, `20260608131649` apply cleanly on a fresh DB and against a populated pre-pluginization DB

## Summary by CodeRabbit

* **New Features**
  * Plugin-driven request routing system replaces built-in fulfillment adapters.
  * Enhanced admin forms support sectioned layouts, conditional field visibility, and dynamic option loading.
  * Request integrations now support plugin-based configuration and validation.

* **Improvements**
  * Structured validation errors for request integrations with per-field error reporting.
  * Dynamic quality and configuration options for request fulfillment.


fluxis:
## Review feedback addressed

Thanks @coderabbitai. Summary of dispositions (pushed in `8cd6aab`, `5dcb9a4`, `1d68ed1`):

**Fixed**
- **`plugins/service.go` — singleflight ctx coupling:** the deduped `ensureClient` launch now runs under `context.WithoutCancel(ctx)`, so a leader caller's cancellation/timeout no longer tears down the shared launch that other waiters depend on. Added a cancellation-isolation test.
- **`api/requests_wiring.go` — nil deps:** `RequestRouterClient` returns a controlled error and `AttachRequestRouter` no-ops when the plugin service is absent (degrades to the existing "no backend configured" failure instead of panicking). Added tests.
- **`schemaForm.ts` — boolean coercion:** `Boolean("false")` was `true`; string booleans are now parsed explicitly.
- **`schemaForm.ts` — `array:num`:** now coerces decimals (`"1.5"`); `array:int` stays integer-only. 
- **`AdminRequests.tsx` — capability sub-id:** the editor now tracks `capability_id` alongside `installation_id` (composite `<Select>` value) so a multi-capability installation resolves the exact backend.
- **`AdminRequests.tsx` — config reset:** switching the selected plugin now clears `pluginConfig` (and stale save errors) so the previous plugin's keys never reach the new plugin's options probe/save.

**Declined**
- **`user_identity_lookup.go` — nil guard:** intentionally not adding the suggested "return empty identity on nil lookup". The adapter is only constructed with a non-nil lookup at wiring time, and silently returning an empty identity would mask a wiring misconfiguration — a silent failure we'd rather surface than swallow.

**Obsolete**
- The markdownlint comments on `docs/superpowers/plans/*` are moot — those files were removed from this PR after the review ran (they're internal design docs, kept out of the upstream diff).

Quick104:
@codex

PR #109, opened 2026-06-09
fix(progress): re-enter Continue Watching when re-watching a finished item

d3v1l1989:
## Problem

Re-watching an already-finished item never re-enters the **Continue Watching** row until the user manually marks it *unwatched*.

The row is built by `ListProgress(profileID, "in_progress")`, whose SQL filters `user_watch_progress` to `WHERE completed = FALSE`. But the playback heartbeat path (`UpdateProgress`, used by both the native player and the jellycompat stream handler) latched `completed` one-way:

```sql
completed = CASE WHEN excluded.completed THEN TRUE ELSE user_watch_progress.completed END,
position_seconds = GREATEST(excluded.position_seconds, user_watch_progress.position_seconds)
```

Once an item is finished (`completed = TRUE`, position reset to duration), starting it again reports a fresh low position — but the `CASE` keeps `completed = TRUE` and `GREATEST` keeps the old end-of-file position, so the re-watch is invisible to the in-progress query forever. Only `ClearProgress`/`ClearProgressBatch` (mark-unwatched) releases the latch.

## Fix

In `UpdateProgress`, release the latch and adopt the fresh resume point when an early-position heartbeat lands on an already-completed row — gated by two guards:

- **Fraction guard** (`RestartDetectFraction = 0.5`): the incoming position must be below 50% of the max known runtime, distinguishing a genuine rewatch-from-the-start from a brief end-of-file seek. `GREATEST/MAX(excluded, stored)` duration handles multi-version/edition rewatches where the runtime differs.
- **Time-gap guard** (`RestartMinGapSeconds = 60`): at least 60s must have elapsed since the row was last updated, rejecting a stale heartbeat from a concurrent/offline session arriving just after completion.

Both guards are AND-ed, so any failure (including a NULL `julianday` in the SQLite path) defaults to keeping the item completed — the conservative choice.

Scoped to `UpdateProgress` only. In-progress rows (`completed = FALSE`) are provably unchanged (both new branches require `completed = TRUE`), and existing finished rows are untouched until actually re-watched.

The SQLite `userdb` store gets the same logic, plus the min-resume early-return guard it was previously missing (silent divergence from the Postgres store).

## Testing

- New table-driven tests (`internal/userdb`): in-progress unchanged, rewatch releases latch + adopts position, late heartbeat (>50%) does not release, recent completion (<60s) blocked by the time guard, completion still latches, multi-version duration via `MAX`, zero stored duration fallback, below-min-resume discard. `go test ./internal/userdb/ ./internal/userstore/...` passes; `go build` clean; `gofmt` clean.
- Postgres CASE logic validated against a live populated catalog; deployed and confirmed fixing the reported behavior.

---
AI-assisted: implementation and tests were written with Claude and reviewed line-by-line by me.

### DIFF
--- a/internal/userdb/helpers.go
+++ b/internal/userdb/helpers.go
@@ -11,7 +11,9 @@ func generateUUID() string {
 	return uuid.New().String()
 }
 
-// nowUTC returns the current time in ISO 8601 format (UTC).
-func nowUTC() string {
+// nowUTC returns the current time in ISO 8601 format (UTC). It is a package
+// var so tests can pin the clock (e.g. to exercise the rewatch time-gap guard
+// deterministically).
+var nowUTC = func() string {
 	return time.Now().UTC().Format(time.RFC3339)
 }
--- a/internal/userdb/progress.go
+++ b/internal/userdb/progress.go
@@ -20,15 +20,39 @@ type WatchHistoryEntry = userstore.WatchHistoryEntry
 // The position is only updated if the new value is greater than the existing one.
 // The completed flag is set to true when position/duration exceeds the watched threshold.
 func UpdateProgress(db *sql.DB, profileID, mediaItemID string, position, duration float64, thresholds userstore.ProgressThresholds) error {
+	if duration > 0 && position > 0 && position/duration < userstore.MinResumeFraction(thresholds.MinResumePct) {
+		return nil
+	}
 	now := nowUTC()
+	// Mirrors the Postgres pgstore UpdateProgress: position moves forward via
+	// MAX, completed latches one-way TRUE — EXCEPT a genuine rewatch of a
+	// finished item (an early-position heartbeat, below RestartDetectFraction
+	// of the runtime, landing on an already-completed row at least
+	// RestartMinGapSeconds later) releases the latch and adopts the fresh
+	// resume point so the item re-enters "Continue Watching". The time gap
+	// rejects stale heartbeats from a concurrent/offline session that arrive
+	// just after completion. julianday() diffs are in days → *86400 = seconds.
 	query := `
 		INSERT INTO watch_progress (profile_id, media_item_id, position_seconds, duration_seconds, completed, updated_at)
 		VALUES (?, ?, ?, ?, ?, ?)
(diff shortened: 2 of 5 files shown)


Quick104:
Thanks for this — the diagnosis is exactly right (the one-way `completed` latch + `GREATEST` position pinning makes rewatches permanently invisible to Continue Watching), and the table-driven tests are solid. We're going to close this in favor of a model-level fix, for two reasons we confirmed while reviewing the guard mechanism:

1. **The 60s time-gap guard is self-defeating for immediate rewatches.** Blocked heartbeats still execute the `DO UPDATE`, so `updated_at = excluded.updated_at` refreshes on every rejected report. The gap is then measured against the *last heartbeat* (~10s cadence), not against completion — once one qualifying heartbeat lands inside the 60s window, the latch can never release for the rest of the session, and the rewatch's progress is silently discarded. Reproduced with a clock-pinned test: a 10-minute episode restarted 30s after finishing stayed `completed=true` through 4+ minutes of heartbeats. Any item shorter than ~20 min restarted promptly hits this, as does any seek past 5% within a minute of finishing.

2. **A position-0 heartbeat un-watches a finished item.** The min-resume early return requires `position > 0`, so an exactly-zero report falls through to the upsert, where `0 < MAX(duration)*0.5` and the time guard (item finished long ago) both pass → latch released at position 0. The native player's `persistProgress` has no `position > 0` guard (the jellycompat path does), so opening a watched item and backing out before playing clears its watched state.

Rather than adding more guards, we're adopting the Jellyfin invariant that removes the need for restart detection entirely: **completion resets `position_seconds` to 0** (instead of pinning it to duration), `completed` stays a pure one-way watched latch, and the resume query keys on `position_seconds > 0` instead of `completed = FALSE`. A rewatch heartbeat then hits `GREATEST(new, 0)` and re-enters Continue Watching naturally — no thresholds, no clock math, no race windows — and the watched badge survives the rewatch (matching Plex and current Jellyfin master, see jellyfin/jellyfin#15762).

A replacement PR is in progress and will reference this one. Thanks again for the thorough writeup — it's what made the underlying model problem obvious.

PR #111, opened 2026-06-09
fix(jellycompat): report fileless episodes as LocationType=Virtual

d3v1l1989:
## Problem

Provider-metadata-only episodes — unaired/missing entries pulled from TVDB/TMDB that have no underlying media file — were mapped with `LocationType=FileSystem` and an empty `MediaSources` list. Jellyfin's contract is that such items report `LocationType=Virtual`.

Because they weren't marked `Virtual`, clients that build playback queues from the episode list (Wholphin, Infuse, Findroid, …) treat them as playable, queue them, and fail on advance with "no media sources". On a large catalog these are common — they're interspersed inside real seasons (e.g. a 12-episode season with 4 fileless entries).

Wholphin specifically filters `LocationType=Virtual` out of its auto-advance playlist, so marking these `Virtual` lets next-episode / skip-outro jump cleanly to the next real episode, and the episode list greys them out as expected.

## Fix

The `LocationType` decision is centralized in a new `applyPlayableLocation` helper (which also clears `VideoType` on virtual items, matching Jellyfin) and applied on **all three** mapper paths:

1. **Detail path** (`itemFromDetailWithFields`): version data is authoritative — zero file versions → `Virtual`, no `MediaSources` (original commit).
2. **List path** (`itemFromList`): a new tri-state `HasMediaFiles` signal on `upstreamListItem`, populated by the episode-targets batch query via an `EXISTS` check against `media_files` (same pattern NextUp already uses). `nil` (producer didn't check — movies, series lists) preserves the historical `FileSystem` default.
3. **Episodes fallback path** (`episodeFromUpstream`, used by `/Shows/{id}/Episodes` when no detail Fields are requested — the path Infuse/Findroid-style clients hit): reuses the already-fetched episode targets, so no extra query.

The pool-less fallback loader uses a new `EpisodeRepository.HasFilesByIDs` batch method.

## Known follow-up

Items whose file went missing (`missing_since` set) are reported as `Virtual` rather than Jellyfin's `Offline`. Both are unplayable so client behavior is correct, but a future change could distinguish them.

## Testing

- `TestVirtualEpisodeLocationType`: detail path — fileless episode → `Virtual` with no `MediaSources`/`VideoType`; real episode → `FileSystem` with `MediaSources`.
- `TestVirtualEpisodeListPaths`: `itemFromList` and `episodeFromUpstream` matrix — `HasMediaFiles` false → `Virtual`, true/nil → `FileSystem`.
- Full `internal/jellycompat` + `internal/catalog` suites pass; `go vet` clean.

> Best reviewed alongside #110 — the repeated-`Fields`-param fix is the primary cause of the Wholphin "no media sources" report; this change is complementary hardening for the genuinely-fileless episodes.
>
> Authored with AI assistance; reviewed and verified by maintainer + author.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of episodes without associated media files.
  * Episodes are now properly identified and distinguished between virtual content and file-backed episodes based on actual media availability.

* **Tests**
  * Added test coverage for episode location type determination logic.

### DIFF
--- a/internal/catalog/episode_repo.go
+++ b/internal/catalog/episode_repo.go
@@ -638,6 +638,32 @@ func (r *EpisodeRepository) GetByIDs(ctx context.Context, contentIDs []string) (
 	return scanEpisodes(rows)
 }
 
+// HasFilesByIDs reports which of the given episodes are backed by at least one
+// live (non-missing) media file. Episodes absent from the result map have no
+// file — e.g. provider-metadata-only entries for unaired episodes.
+func (r *EpisodeRepository) HasFilesByIDs(ctx context.Context, contentIDs []string) (map[string]bool, error) {
+	result := make(map[string]bool, len(contentIDs))
+	if len(contentIDs) == 0 {
+		return result, nil
+	}
+	query := `SELECT episode_id FROM media_files
+		WHERE episode_id = ANY($1) AND missing_since IS NULL
+		GROUP BY episode_id`
+	rows, err := r.pool.Query(ctx, query, contentIDs)
+	if err != nil {
+		return nil, fmt.Errorf("checking episode file presence: %w", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var episodeID string
+		if err := rows.Scan(&episodeID); err != nil {
+			return nil, fmt.Errorf("scanning episode file presence: %w", err)
+		}
+		result[episodeID] = true
+	}
+	return result, rows.Err()
+}
+
 // ListBySeries returns all episodes for a given series, ordered by season and
 // episode number.
 func (r *EpisodeRepository) ListBySeries(ctx context.Context, seriesID string) ([]*models.Episode, error) {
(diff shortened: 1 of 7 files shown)


Quick104:
Thanks for the fix — the `Virtual` contract is exactly right. While reviewing I noticed the same fileless episode still reported `FileSystem` on the list-side mappers (`itemFromList` and the `/Shows/{id}/Episodes` non-detail fallback via `episodeFromUpstream`), so clients that don't request `Fields=MediaSources` wouldn't see the fix.

I pushed e4d69c4c on top of your commit (via maintainer edit) extending it:

- centralized the decision in `applyPlayableLocation` (also clears `VideoType` on virtual items, matching Jellyfin)
- plumbed a tri-state `HasMediaFiles` signal into the episode-targets batch query (`EXISTS` against `media_files`, same pattern NextUp uses) and a new `EpisodeRepository.HasFilesByIDs` for the pool-less fallback
- the episodes handler reuses the already-fetched targets, so no extra query on the hot path
- added `TestVirtualEpisodeListPaths` covering the false/true/nil matrix on both list mappers

PR body updated accordingly. One known follow-up noted there: missing-file items are conflated with never-aired ones (`Virtual` vs Jellyfin's `Offline`).

PR #126, opened 2026-06-10
feat(nodepool): node groups, per-node caps, and local transcode fallback control

### LINKED ISSUE #93: feat(distributed): Enable egress from transcode workers
**Issue:** User would like to egress directly from transcode workers to proxy hosted on same host OR egress directly from the transcode worker.

**Expected behaviour:** Transcode nodes being able to egress direct to clients.
**Experienced behaviour:** Transocde is being bounced within internal network, effectively doubling egress.

Allow egress from transcode workers (assume they have access to storage directly, egress to internet from transcode-1.mydomain.com, or allow pinning of transcode egresses to specific proxies IE transcode-1.mydomain.com can only egress through proxy-1.mydomain.com, etc.)

Maybe something like:
transcode-url:
proxy-url:

Ideally, any machine able to transcode I'd also like to proxy from.

Additional Feature, unrelated:
Set max transcodes per node. Set max total streams per node, loadbalance from there.

Quick104:
## Problem

Closes #93. In distributed deployments, the proxy serving a transcoded stream is picked round-robin with no relationship to the transcode node producing it. Every HLS segment then travels transcode node → proxy → client, and when those nodes sit on different hosts/LAN segments, internal egress is effectively doubled.

## What changed

**Node grouping.** Stream nodes get an optional free-form `group` label (admin UI + API). Nodes sharing a group are treated as co-located: when a transcode node from group G is selected, the stream is always served by a proxy from G (round-robin when G has several), so transcoded bytes stay on that LAN. A group is only eligible while **all** of its enabled members are healthy — a sick proxy sidelines the group's transcode nodes rather than letting pinned streams dangle. Ungrouped nodes keep the historical behavior exactly (least-connections transcode pick, global round-robin proxy, soft session affinity).

**Unified selection.** A new `nodepool.Planner` (`PlanSession`) is now the single place transcode/proxy pairs are chosen. It replaces the independent `ProxyPool.Pick()` / `TranscodePool.Acquire()` calls at five call sites (native start, audio change, transcode start, jellycompat stream + manifest) and deletes the duplicated soft-affinity `pickTranscodeNode` copy in jellycompat.

**Per-node capacity caps** (also from #93). New nullable `max_jobs` per node — max concurrent transcodes for transcode nodes, max concurrent streams for proxies; empty = unlimited. Admission control combines health-reported job counts with short-lived per-session reservations that stop counting as soon as a fresher health report arrives (or after 90s), bridging the 30s health-check window without any session-teardown bookkeeping. Proxy nodes previously reported no load at all; their health endpoint now returns real stream counts, with HLS sessions tracked via a new idle-expiry `Touch` mechanism in the session tracker.

**Proxy bandwidth caps.** Proxy nodes now measure their stream egress — a rolling 60-second average over everything under `/stream`, counted by a metering middleware — and report it through the health endpoint (visible live in the admin UI). A new nullable `max_bandwidth_kbps` per proxy ("Max Egress Bandwidth (Mbps)" in the UI) gates admission: a new stream is only routed to a proxy where its expected bitrate (transcode target bitrate, or source bitrate for direct play/remux) fits within the cap. Recently admitted streams are bridged as weighted bandwidth reservations for the meter window, since the rolling average only converges on a new stream's rate gradually. A group whose proxies lack bandwidth headroom is treated as full, same as the job cap. Active streams are never interrupted — the cap only gates new admissions.

**Local fallback control.** New `playback.local_transcode_fallback` setting (Admin → Settings → Playback, default **on** = historical behavior). When off, transcode requests fail with 503 instead of falling back to ffmpeg on the API server when no eligible node exists — for admins who never want the API host transcoding. Enforced on both native and Jellyfin-compat paths.

**Data race fix.** The health checker used to mutate shared `*Node` structs in place while request paths read them. Health results are now published as immutable node copies swapped in under the pool lock (verified with `go test -race`).

## Notes for review

- Migration `20260610130619_node_groups_and_caps.sql` adds four nullable/defaulted columns; existing deployments are unaffected (NULL group = ungrouped, NULL caps = unlimited). Not yet applied to the shared dev DB.
- Clearing semantics: empty-string group and caps `<= 0` reset to NULL — avoids the "cap 0 = never selected" foot-gun and keeps the update API simple.
- Bandwidth admission is deliberately conservative during the 60s bridge window (measured egress + full reservation can briefly double-count a ramping stream); the safe direction for an admission control.
- The egress meter wraps `/stream` response writers, which bypasses the `sendfile` fast path for direct play through a proxy — a deliberate trade for accurate measurement.
- Stream URLs, token claims, and client-facing API shapes are unchanged — no Android/Apple client follow-up needed; only the admin web UI changed.
- A group whose proxies are all at cap (jobs or bandwidth) is treated as full: its transcode nodes are skipped (a group's capacity is bounded by its proxies).
- Known pre-existing issue left out of scope: remux sessions start on `/stream/transcode/` but switch to `/stream/remux/` after an audio change; tracked separately.
- Verified: `go build ./...`, full `go test ./...` (`-race` on changed packages; 19 planner tests + egress meter tests), golangci-lint clean for changed lines, `pnpm run build`, ESLint/Prettier, `make verify-local-paths`.

## AI-use disclosure

Implemented with Claude Code (architecture explored, designed, and reviewed via multi-agent passes).

## Summary by CodeRabbit

* **New Features**
  * Node grouping and per-node concurrency/bandwidth caps plus an intelligent session planner for proxy/transcode placement.

* **Improvements**
  * Admin UI: view/edit group, max streams and bandwidth; health now reports egress bandwidth; active stream tracking improved.
  * Player warms connections to stream origins.
  * New setting: Local Transcode Fallback toggle.

* **Behavior**
  * Streams may be routed via remote nodes; transcode-required requests can return 503 when remote transcode is disallowed.

### DIFF
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -670,6 +670,7 @@ func main() {
 
 		deps.ProxyPool = proxyPool
 		deps.TranscodePool = transcodePool
+		deps.NodePlanner = nodepool.NewPlanner(proxyPool, transcodePool)
 
 		healthChecker := nodepool.NewHealthChecker(proxyPool, transcodePool, nodeRepo)
 		healthChecker.Start(appCtx)
@@ -1897,8 +1898,7 @@ func main() {
 			DB:               deps.DB,
 			SecretCipher:     dataCipher,
 			ClientIPResolver: ipResolver,
-			ProxyPool:        deps.ProxyPool,
-			TranscodePool:    deps.TranscodePool,
+			NodePlanner:      deps.NodePlanner,
 			JWTSecret:        cfg.Auth.JWTSecret,
 			RecWorker:        recWorker,
 		}
--- a/internal/api/handlers/nodes.go
+++ b/internal/api/handlers/nodes.go
@@ -26,7 +26,7 @@ type NodeRepository interface {
 	Create(ctx context.Context, input nodepool.CreateNodeInput) (*nodepool.Node, error)
 	Update(ctx context.Context, id int, input nodepool.UpdateNodeInput) (*nodepool.Node, error)
 	Delete(ctx context.Context, id int) error
-	UpdateHealth(ctx context.Context, id int, healthy bool, activeJobs int) error
+	UpdateHealth(ctx context.Context, id int, healthy bool, activeJobs, egressKbps int) error
 }
 
 // NodeListEnabled queries enabled nodes by type for pool reload.
@@ -70,6 +70,7 @@ type ForceReloadResult struct {
 type checkNodeResult struct {
 	Healthy    bool `json:"healthy"`
 	ActiveJobs int  `json:"active_jobs"`
+	EgressKbps int  `json:"egress_kbps"`
 }
 
 // HandleListNodes handles GET /admin/nodes.
@@ -181,15 +182,16 @@ func (h *NodeHandler) HandleCheckNode(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	healthy, activeJobs := nodepool.CheckNode(r.Context(), node)
+	healthy, activeJobs, egressKbps := nodepool.CheckNode(r.Context(), node)
 
-	if err := h.repo.UpdateHealth(r.Context(), id, healthy, activeJobs); err != nil {
+	if err := h.repo.UpdateHealth(r.Context(), id, healthy, activeJobs, egressKbps); err != nil {
 		slog.Error("persisting health check result", "node_id", id, "error", err)
 	}
 
 	writeJSON(w, http.StatusOK, checkNodeResult{
 		Healthy:    healthy,
 		ActiveJobs: activeJobs,
+		EgressKbps: egressKbps,
 	})
 }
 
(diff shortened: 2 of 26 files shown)


Quick104:
## Added: node-mode stream-start latency improvements (`e7de2bb6`)

Traced the distributed startup waterfall (client → API → transcode node → proxy) and trimmed the safe overhead that integrated mode doesn't pay:

- **Player preconnect** — the web player now warms DNS/TCP/TLS to the proxy origin the moment `/playback/start` returns, overlapping the handshakes with the transcode dispatch instead of paying them at the first manifest fetch (easily 200–600ms on a WAN proxy).
- **Faster 202 from the transcode node** — the Redis session-track write moves off the request path, and a replaced session's segment directory is renamed aside and deleted in the background. The synchronous `os.RemoveAll` could add seconds to quality switches on slow disks while the client sat waiting.
- **Proxy→node transport tuning** — larger idle-connection pool (the Go default of 2 idle conns per host caused churn/re-handshakes with many viewers on one proxy→node pair) and a 60s response-header timeout, so a hung transcode node can no longer hang client requests indefinitely.
- **jellycompat dispatch timeout** — `startRemoteTranscode` now has the same 10s bound as the native path; an unreachable node previously hung the compat manifest request until the OS gave up.

Deliberately **not** done here: making the API→node `/transcode/start` dispatch asynchronous (it would remove a full round trip + ffmpeg fork from the client's critical path, but changes failure semantics — a down node would surface as manifest errors instead of a clean 502 at start). Worth its own design discussion.

PR #138, opened 2026-06-11
feat(manga): manga library type — series grouping, reading loop, AniList/MangaDex metadata + status badge

Quick104:
## ⚠️ Prerequisite

**Requires [Silo-Server/silo-plugin-sdk#4](https://github.com/Silo-Server/silo-plugin-sdk/pull/4) merged and tagged `v0.7.0` before this can build.**
The final commit pins `github.com/Silo-Server/silo-plugin-sdk v0.7.0`, which adds `MetadataItem.status`. `go.sum` already carries the canonical `v0.7.0` content hashes, so CI will resolve cleanly the moment the tag exists — no further `go.sum` edits needed.

## Problem

Silo had no first-class **manga** library type. Manga sit awkwardly between ebooks (per-file `.cbz`/`.epub`) and series (one enrichment target per title with many child items): scanning a manga folder as loose ebooks produced one card per volume with no series-level cover, synopsis, or read-through, and library-listing sections (recently-added/released/random) were polluted by individual chapters.

## Approach

A new `type='manga'` library forked from the ebook pipeline:

- **Scanner** groups `.cbz`/`.epub` under a series folder into one `type='manga'` series plus linked `type='ebook'` chapter items (`manga_chapters` join table). One card, one enrichment target.
- **Catalog/browse/detail** surface per-series volume + chapter counts, per-chapter read state, and a continue/start-reading loop; library-listing sections exclude manga chapters so only the series card shows.
- **Metadata** is provided by the new external `silo-plugin-manga-metadata` plugin (`metadata_provider.v1`, capability `manga-metadata`), enriching series with cover, synopsis, genres, author/artist, year, banner-as-backdrop, and **publication status** (Ongoing/Completed/Hiatus/Cancelled/Upcoming). Status flows through `MetadataItem.status` → `ShowStatus` → a color-coded chip on the detail page and browse cards.
- **Status badge** required a new SDK field (`MetadataItem.status`, field 31). The host consumes it at `internal/metadata/plugin_provider.go`; the final commit here de-vendors the temporary SDK copy and depends on the published `v0.7.0`.

## Deferred follow-ups

1. **Per-source N+1 in the ebook/audiobook enrichers** — manga enrichment batches provider-id lookups, but the sibling ebook/audiobook enrichers still issue one lookup per item. Generalizing the batched path is left as a separate change to keep this PR scoped.
[...]

PR #142, opened 2026-06-12
fix(manga): exclude chapters from the matcher's unmatched-item lister (stacks on #138)

fluxis:
Stacked on #138 (targets its branch `codex/rebase-pr-137`) — meant to be merged into that PR before it lands.

## Problem

Manga chapters are `type='ebook'` items that stay `status='pending'` by design: provider metadata lives on the `type='manga'` series item, and chapters are already excluded from the ebook enricher's claim query. But the metadata matcher's scan-final retry pass (`RetryUnmatchedItemsByFolderAndPathPrefix`) lists every `pending` item and synchronously runs a full provider search per item.

On a real library (31,564 chapters), each chapter resolved to the ebook-metadata plugin at its ~1 req/s rate limit — **8h46m of guaranteed no-matches appended to a 2-minute scan** (measured live: `scan_duration=2m3s, retry_duration=8h46m13s, retried_items=31564, still_unmatched=31564`). Earlier scans never survived to completion (killed by deploy restarts mid-march), so the library's `last_scanned_at` stayed NULL and every nightly scan restarted the full 9-hour march, hammering book-source APIs with chapter-filename searches.

## Fix

Add the same `manga_chapters NOT EXISTS` guard the ebook enricher already uses to `ItemRepository.ListUnmatchedByFolderAndPathPrefix`. No data migration needed — existing `pending` chapter rows simply stop being listed. The drainer/claim paths are already safe (file claimers skip files bound to items; verified `match_attempted_at` untouched across all 31,619 files).

## Verification

- TDD: SQL-shape test pinning the exclusion (fails without the clause), full `internal/catalog`, `internal/metadata`, `internal/scanner`, `internal/libraryingest` green; `go build` + `go vet` clean.
- Live re-scan of the same 31,564-chapter library after deploying this change: **27 seconds total**, `retry_duration=109ms, retried_items=0`, zero errors, `last_scanned_at` set.

AI-use disclosure: root-cause analysis (delve goroutine inspection on the live process), fix, and tests by Claude Code, reviewed and live-verified before submission.

### DIFF
--- a/internal/api/handlers/ebook_reader.go
+++ b/internal/api/handlers/ebook_reader.go
@@ -1046,15 +1046,25 @@ func (s *PGEbookReaderProgressStore) Upsert(ctx context.Context, progress EbookR
 	if s == nil || s.pool == nil {
 		return fmt.Errorf("ebook reader progress store is not configured")
 	}
-	if _, err := s.pool.Exec(ctx, `
+	// A routine autosave (e.g. reopening a finished book) must not silently drop
+	// a "finished" item below the threshold and un-mark it read; once finished,
+	// progress only moves on an explicit unread (which deletes the row). Below
+	// the threshold, progress tracks freely. Manga chapter ✓ marks ride on this
+	// same row, so the guard protects them too.
+	query := fmt.Sprintf(`
 		INSERT INTO ebook_reader_progress
 			(user_id, profile_id, content_id, file_id, location, progress, updated_at)
 		VALUES ($1, $2, $3, $4, $5, $6, $7)
 		ON CONFLICT (user_id, profile_id, content_id) DO UPDATE SET
 			file_id = EXCLUDED.file_id,
 			location = EXCLUDED.location,
-			progress = EXCLUDED.progress,
-			updated_at = EXCLUDED.updated_at`,
+			progress = CASE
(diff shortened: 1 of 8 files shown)


fluxis:
Appended the ebook + manga performance/correctness fixes from the same live-profiling pass (3 commits on top of the matcher-flood fix):

- `fix(scanner)`: ebook/comic files (epub/pdf/cbz/cbr, incl. manga chapters) no longer re-run `ffprobe` on every detail/watch load — they hit the probe-repair path that never converged for non-AV files. The ebook/manga **detail-page killer**.
- `perf+fix(ebooks)`: parallelize the ebook detail related-content queries; preserve finished read-state (a routine autosave could silently un-mark a finished book / clear a manga chapter ✓).
- `perf(manga)`: batch the per-chapter poster presign + add `rows.Err()`; add `idx_manga_chapters_series_volume` for the volume-count chip subquery; quiet the per-chapter scan log.

Note: the probe-repair fix is split with the audiobook PR (#169) — this PR carries the ebook early-return; #169 carries the audiobook video-gating. `probe_repair.go` will need a trivial 3-way merge when both land. CI is red until SDK v0.7.0 is tagged (same as the rest of this stack). AI-assisted.

PR #157, opened 2026-06-15
fix(jellycompat): match the Static stream query parameter case-insensitively

d3v1l1989:
## Problem

`GET /Videos/{id}/stream?static=true` (lowercase key) returns `404 "Playback session not found"` for clients that direct-play without first calling `PlaybackInfo`. `HandleVideoStream` reads the static-play guard case-**sensitively** (`r.URL.Query().Get("Static")`), so a lowercase `static` key never triggers the on-the-fly `createStaticPlaySession` fallback, and the earlier `resolvePlaybackRoute` error surfaces as a 404.

Observed against a live deployment: SenPlayer sends `static=true` (lowercase) and 404s ~83% of its direct-play attempts (≈246 fail / 50 ok in a ~20-minute capture). A valid `api_key` / `X-Emby-Token` / `PlaySessionId` is present, so this is not an auth failure. Any client that lowercases `static` is affected.

## Approach

Read the guard through the package's existing `newCaseInsensitiveQuery` helper (already used for `PlaySessionId` in `auth.go` and `HandleMasterManifest`) so `static` / `Static` / any casing matches. One line; the value comparison was already case-insensitive via `strings.EqualFold`. This is the same case-insensitive-query class as the prior Wholphin lowercase `playSessionId` fix.

## Testing

New `playback_static_directplay_test.go`:
- lowercase `static=true` now serves the file (was 404)
- uppercase `Static=true` still serves (regression guard for the original Infuse path)
- no `static` + no prior session still 404s (the fallback must not fire unconditionally)
- lowercase `static` + `mediaSourceId` (the exact SenPlayer call shape)

`go test ./internal/jellycompat/...` passes in a libvips-capable container. The test was mutation-verified (fails when the fix is reverted).

## Notes

Additive and behavior-preserving — it only broadens which query-key casings trigger the existing static-play path.

---
*AI-use disclosure: implemented and tested with AI assistance (Claude Code). The diff and tests were human-reviewed and run in a libvips Docker container.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed video streaming query parameters to be handled case-insensitively, improving compatibility with various client implementations.

### DIFF
--- a/internal/jellycompat/playback_static_directplay_test.go
+++ b/internal/jellycompat/playback_static_directplay_test.go
@@ -1,0 +1,141 @@
+package jellycompat
+
+import (
+	"context"
+	"net/http/httptest"
+	"net/url"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/go-chi/chi/v5"
+
+	"github.com/Silo-Server/silo-server/internal/catalog"
+	"github.com/Silo-Server/silo-server/internal/models"
+)
+
+// newStaticDirectPlayHandler builds a PlaybackHandler wired for an end-to-end
+// HandleVideoStream direct-play serve: a real temp file (so ServeDirectPlay
+// returns 200), an empty playback store (so resolvePlaybackRoute fails and the
+// Static fallback is exercised), and a stub session manager (HandleVideoStream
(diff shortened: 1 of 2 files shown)


Quick104:
### Code review

PR #158, opened 2026-06-15
fix(jellycompat): serve item and user images anonymously to match Jellyfin

d3v1l1989:
## Problem

Two image routes return `401` where genuine Jellyfin serves anonymously:

- `GET /Items/{id}/Images/{type}[/{index}]` returns `401` (not `404`) for absent/unsupported art when the request carries no session — e.g. chapter thumbnails (which Silo only ever serves as embedded `Chapters[].ImagePath`, never via this route) or any plain `` request. Jellyfin's `ImageController.GetItemImage` / `GetItemImageByIndex` carry no `[Authorize]` and document `200`/`404` only.
- `GET /Users/{id}/Images/Primary` is registered inside the session-auth group, so an avatar `` (which can't attach an auth header) `401`s. Jellyfin's `GetUserImage` / `GetUserImageLegacy` are anonymous as well.

Real clients (e.g. Fladder) hit both → missing chapter thumbnails and user avatars.

## Approach

- **Item images:** when there is no session after the existing token/tag fallbacks, return a clean `404 "Image not found"` instead of `401` — matching the contract and the handler's own default-case 404. The signed-tag rejection paths are untouched (a forged/stale tag still does not serve the image).
- **User avatar:** move `GET`/`HEAD /Users/{id}/Images/Primary` to a top-level anonymous route and seed the deterministic placeholder from the path pseudo-user id (no session needed). To keep the now-anonymous route from becoming a CPU amplifier, the placeholder is selected from a small precomputed palette by hashing the id, rather than generated per request.

## Testing

New tests: chapter image without session → `404` (not `401`); warm Primary cache redirects anonymously; user avatar serves a stable PNG without a session (byte-stable per id, varied across ids, plus a HEAD variant). Two existing tag-rejection tests updated `401`→`404` — the image is still never served, only the rejection status changed (now contract-correct). Green in a libvips container; tests mutation-verified.

## Notes / what to watch

- This changes a status code on existing compat routes (`401`→`404`). It is contract-correcting (Jellyfin documents `200`/`404`, never `401`, on these anonymous GETs); no client should depend on the `401`.
- Placeholder avatar bytes change for existing users (seed switched to the path id — cosmetic, no stored avatars exist) and there are now 16 color buckets rather than per-user-unique colors.
- These touch auth policy; happy to split further or gate behind discussion if maintainers prefer.

---
*AI-use disclosure: implemented and tested with AI assistance (Claude Code), including an adversarial security pass that caught and fixed a DoS-amplification risk on the now-anonymous avatar route. Diff and tests were human-reviewed and run in a libvips Docker container.*

### DIFF
--- a/internal/jellycompat/handlers_images.go
+++ b/internal/jellycompat/handlers_images.go
@@ -3,7 +3,6 @@ package jellycompat
 import (
 	"context"
 	"errors"
-	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
@@ -114,7 +113,10 @@ func (h *ImagesHandler) HandleItemImage(w http.ResponseWriter, r *http.Request) 
 		}
 	}
 	if session == nil {
-		writeError(w, http.StatusUnauthorized, "Unauthorized", "Missing authentication token")
+		// Jellyfin item/chapter image GETs are anonymous (200/404 only, never
+		// 401): media players can't attach auth headers to <img> requests, and
+		// absent or unsupported art (e.g. Chapter) must degrade to a clean 404.
+		writeError(w, http.StatusNotFound, "NotFound", "Image not found")
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


Quick104:
### Code review

PR #159, opened 2026-06-15
feat(jellycompat): support DELETE /Videos/ActiveEncodings transcode teardown

d3v1l1989:
## Problem

Jellyfin clients (e.g. JellyCon) call `DELETE /Videos/ActiveEncodings` when playback stops, to tear down the running HLS transcode. The compat layer didn't implement it, so a transcode kept running until the playback-session TTL (default 6 h) expired.

## Approach

Implement `DELETE /Videos/ActiveEncodings` in the session-auth group (Jellyfin marks the endpoint `[Authorize]`). It resolves the play session by `PlaySessionId` and tears it down, reusing a shared `teardownPlaySession` helper extracted from the existing Stopped-report stop path (removing duplicated teardown logic). Every step is idempotent, so an already-gone session is a clean `204`.

Ownership is enforced: only the session's own caller (matching `CompatToken`) may tear it down; an unknown or not-owned `PlaySessionId` is a uniform `204` no-op — no cross-session teardown and no ownership oracle — mirroring the ownership check the Stopped report already uses. `DeviceId` is intentionally ignored: Silo's store is keyed by `PlaySessionId`, clients always send it, and Jellyfin's own teardown matches by `playSessionId` when it is non-empty.

## Testing

New tests: happy path stops the upstream session and deletes the store entry (`204`); missing `PlaySessionId` → `204` no-op; unknown session → `204`; lowercase `playSessionId` key resolves; a foreign-token request does **not** tear down the session (ownership guard); and the real `DeviceId=...&PlaySessionId=...` client shape. Green in a libvips container; the ownership guard was mutation-verified (the foreign-session test fails when the check is removed).

## Notes / scope

This adds a new compat endpoint. If it falls under the v1 capability scope gate (`docs/architecture/v1-scope.md`), I'm happy to file a capability proposal first — flagging for triage. `DeviceId` is accepted but ignored by design (the store has no device index).

---
*AI-use disclosure: implemented and tested with AI assistance (Claude Code), including an adversarial review that caught and fixed a cross-session IDOR (a missing ownership check) before submission. Diff and tests were human-reviewed and run in a libvips Docker container.*

## Release Notes

* **New Features**
  * Added a `DELETE /Videos/ActiveEncodings` endpoint to stop active encoding sessions with session ownership validation.

* **Improvements**
  * Consolidated session teardown logic to ensure consistent behavior across playback stop operations and the new delete endpoint.

* **Tests**
  * Extended test coverage for the new endpoint, validating session ownership, idempotent behavior, and proper session cleanup.

### DIFF
--- a/internal/jellycompat/audio_selection_test.go
+++ b/internal/jellycompat/audio_selection_test.go
@@ -21,6 +21,7 @@ type testCompatSessionManager struct {
 	sessions        map[string]*playback.Session
 	audioTrackCalls []compatAudioTrackCall
 	progressCalls   int
+	stopCalls       []string
 }
 
 type compatAudioTrackCall struct {
@@ -67,7 +68,10 @@ func (m *testCompatSessionManager) UpdateAudioTrack(sessionID string, audioTrack
 	return nil
 }
 
-func (m *testCompatSessionManager) StopSession(string) error { return nil }
+func (m *testCompatSessionManager) StopSession(sessionID string) error {
+	m.stopCalls = append(m.stopCalls, sessionID)
+	return nil
+}
 
 func (m *testCompatSessionManager) GetSession(sessionID string) (*playback.Session, error) {
 	if session, ok := m.sessions[sessionID]; ok {
--- a/internal/jellycompat/router.go
+++ b/internal/jellycompat/router.go
@@ -202,6 +202,7 @@ func NewRouter(deps Dependencies) chi.Router {
 			r.Post("/Sessions/Playing", playbackHandler.HandleSessionPlaying)
 			r.Post("/Sessions/Playing/Progress", playbackHandler.HandleSessionPlayingProgress)
 			r.Post("/Sessions/Playing/Stopped", playbackHandler.HandleSessionPlayingStopped)
+			r.Delete("/Videos/ActiveEncodings", playbackHandler.HandleDeleteActiveEncodings)
 			r.Post("/Sessions/Logout", authHandler.HandleLogout)
 			r.Get("/socket", HandleSocket)
 		})
--- a/internal/jellycompat/streams.go
+++ b/internal/jellycompat/streams.go
@@ -672,6 +672,62 @@ func (h *PlaybackHandler) HandleSessionPlayingStopped(w http.ResponseWriter, r *
 	h.handlePlaybackReport(w, r, true)
 }
 
+// HandleDeleteActiveEncodings handles DELETE /Videos/ActiveEncodings.
+//
+// Jellyfin clients (e.g. JellyCon) call this endpoint when playback stops to
+// signal the server to tear down any running HLS transcode for the session.
+// Without it, the transcode process keeps running until the playback session
+// TTL expires (default 6 h). We honour the request by stopping the transcode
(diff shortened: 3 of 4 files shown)


Quick104:
### Code review

I validated the PR's claims and they largely hold: ownership enforcement, `DeviceId` handling, case-insensitive query parsing, and the shared `teardownPlaySession` extraction all check out, and the tests exercise them well. One issue worth addressing:

Found 1 issue:

1. `HandleDeleteActiveEncodings` drops the `UpstreamSessionID == ""` guard that the sibling `Stopped`-report path keeps, so the teardown is not actually "no-op-safe for an already-gone session" as `teardownPlaySession`'s doc comment claims.

`HandlePlaybackInfo` creates a `PlaybackSession` with an empty `UpstreamSessionID`; that field is only populated later by `ensureUpstreamPlayback` on the first manifest/stream request. If a client sends `DELETE /Videos/ActiveEncodings` with a matching token in that early window, the ownership guard passes and `teardownPlaySession` runs `h.playbackStore.Delete(playSession.ID)` — deleting a live session that was never torn down. A subsequent `HandleMasterManifest` then returns 404 ("Playback session not found"). `handlePlaybackReport` avoids exactly this by including `|| playSession.UpstreamSessionID == ""` in its guard.

https://github.com/Silo-Server/silo-server/blob/0b6c3e7b2615fc5958aefcd9f7c40698d74c7542/internal/jellycompat/streams.go#L702-L709

The sibling guard it diverges from:

https://github.com/Silo-Server/silo-server/blob/0b6c3e7b2615fc5958aefcd9f7c40698d74c7542/internal/jellycompat/streams.go#L748-L750

Adding `|| playSession.UpstreamSessionID == ""` to the guard (or short-circuiting inside `teardownPlaySession`) restores parity. Real-world frequency is low — clients normally call this only after a transcode is running — but it's a genuine divergence from the stated idempotency/no-op contract.

Note (process): per the root [CLAUDE.md](https://github.com/Silo-Server/silo-server/blob/0b6c3e7b2615fc5958aefcd9f7c40698d74c7542/AGENTS.md#L92-L105) "Silo v1 Process," a new user-facing capability needs a filed/triaged v1 capability proposal and a `Part of #NNN` link before a feature PR. The PR body already self-flags this for triage.

Quick104:
Pushed [`c08b51b3c0eb66c99b2211a85571830beb856c71`](https://github.com/Silo-Server/silo-server/commit/c08b51b3c0eb66c99b2211a85571830beb856c71) to address the review finding.

`HandleDeleteActiveEncodings` now mirrors the `Stopped` report path's guard — an unknown, not-owned, **or not-yet-started** (`UpstreamSessionID == ""`) `PlaySessionId` is a uniform idempotent 204 no-op. This stops a `DELETE /Videos/ActiveEncodings` that lands in the window between `PlaybackInfo` and the first manifest request from deleting a live session and 404-ing the pending manifest. Added `TestHandleDeleteActiveEncodings_NotYetStartedNotTornDown`, which fails if the guard is removed.

The scope-gate process note (v1 capability proposal + `Part of #NNN`) is unchanged and left for triage as you flagged in the PR description.

PR #160, opened 2026-06-15
fix(jellycompat): return seasons/episodes when browsing a series via /Items?ParentId=

d3v1l1989:
## Problem
Clients that browse a show through the generic `/Items?ParentId=` endpoint **without** `IncludeItemTypes` — e.g. the VoidTV Android-TV client's `getSeasons(seriesId)`/`getEpisodes(seasonId)`, which send `ParentId`+`SortBy=SortName` and no type filter — received the **top-level library views** (CollectionFolders) instead of the series' seasons / the season's episodes.

`HandleItems` only honored an item/season `ParentId` when `IncludeItemTypes=Season` was also present; otherwise `parentItemID` was dropped and the request fell through to the library-views fallback (or a global type browse). On a real Jellyfin server `/Items?ParentId={series}` returns seasons and `/Items?ParentId={season}` returns episodes.

Confirmed live against a production deployment: `/Items?ParentId={series}` and `{season}` both returned the library CollectionFolders, while the dedicated `/Shows/{id}/Seasons|Episodes` routes worked — which is why clients using those dedicated routes were unaffected and only `/Items?ParentId=` browsers (VoidTV) broke.

## Approach
- Decode a season `ParentId` into a dedicated `parentSeasonID` (the codec already kind-tags item vs season ids; decoding season first keeps them distinct).
- Route item/season parents in `HandleItems`:
  - season `ParentId` → that season's episodes
  - series `ParentId` → the series' seasons (default, and explicit `IncludeItemTypes=Season`), or the series' episodes for `IncludeItemTypes=Episode`; a type filter a series can't satisfy (e.g. `Movie`) returns an empty page rather than a wrong-typed listing.
- Extract `writeSeriesEpisodesResponse` as the shared core of `/Shows/{id}/Episodes` and the new browse paths (avoids duplicated logic). The generic `/Items` browse paths honor `StartIndex`/`Limit` (matching the `/Items` contract and the sibling seasons listing); `/Shows/{id}/Episodes` keeps its existing whole-season response.

## Testing
- New `parent_browse_test.go`: series→seasons, season→episodes, unsatisfiable-type-filter→empty, season-parent paging, season-parent-without-repo→empty-not-views, and the parser's season/item `ParentId` distinction. The series/season tests use a content double that panics in `ListUserLibraries`, so a views-fallthrough regression fails the test rather than silently passing.
- `go test ./internal/jellycompat/...`, `go vet`, and `gofmt` are clean, run in a libvips-capable container (the package needs libvips to build).

## Compat
Additive: `/Shows/{id}/Seasons|Episodes` behavior is unchanged; movie/episode `ParentId`s degrade to an empty page (never a 500).

## AI use
Implemented with AI assistance; the diff was reviewed in a separate pass and the tests were run in a Docker toolchain before submitting.

## Summary by CodeRabbit

* **New Features**
  * Added support for browsing episodes within seasons
  * Enhanced series navigation to display seasons by default, with option to view episodes directly
  * Improved pagination support for episode listings with accurate record counts and page navigation

### DIFF
--- a/internal/jellycompat/handlers_items.go
+++ b/internal/jellycompat/handlers_items.go
@@ -222,6 +222,13 @@ func (h *ItemsHandler) handleItemParentChildren(w http.ResponseWriter, r *http.R
 // episodes of that season. The season is resolved to its owning series so the
 // shared episode-listing path (also used by /Shows/{id}/Episodes) can be reused.
 func (h *ItemsHandler) handleSeasonEpisodeChildren(w http.ResponseWriter, r *http.Request, session *Session, query itemsQuery) {
+	// A season's only children are its episodes; a type filter that excludes
+	// Episode (e.g. IncludeItemTypes=Movie) yields nothing, mirroring the
+	// series-parent path's handling of unsatisfiable type filters.
+	if query.hasItemTypeFilter && !itemTypesContain(query.itemTypes, "episode") {
+		writeJSON(w, http.StatusOK, emptyQueryResult(query.startIndex))
+		return
+	}
 	if h.seasonRepo == nil {
 		writeJSON(w, http.StatusOK, emptyQueryResult(query.startIndex))
 		return
--- a/internal/jellycompat/handlers_items.go
+++ b/internal/jellycompat/handlers_items.go
@@ -38,6 +38,7 @@ type ItemsHandler struct {
 	detailSvc    *catalog.DetailService
 	itemRepo     itemRepoForBatchLoader
 	episodeRepo  episodeRepoForBatchLoader
+	seasonRepo   imageSeasonRepository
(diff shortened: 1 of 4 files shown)


d3v1l1989:
Addressed the review comment in 1f02307: `handleSeasonEpisodeChildren` now returns an empty page when `IncludeItemTypes` is present but excludes `Episode` (e.g. `IncludeItemTypes=Movie`), short-circuiting before the season/episode lookup — symmetric with the series-parent path's handling of unsatisfiable type filters. Added `TestHandleItems_SeasonParentNonEpisodeTypeIsEmpty` (asserts empty + that episode listing is not reached even though episodes exist).

PR #165, opened 2026-06-16
fix(jellycompat): reuse the play session for Static direct play instead of leaking one per request

d3v1l1989:
## Problem

Clients that direct-play via `GET /Videos/{id}/stream?…&Static=true` never call
`PlaybackInfo`, so they send their own client-generated `PlaySessionId` and repeat
it on every byte-range request. `resolvePlaybackRoute` looked that id up, missed
(the server never minted it), returned `ErrSessionNotFound`, and `HandleVideoStream`
fell to `createStaticPlaySession` → a **new upstream session for every request**.

Each new session counts toward the per-user `max_streams` cap and only ages out
after the 45s activity grace, so a single direct play's range requests pile up
orphaned sessions and quickly trip the cap → `429 TooManyStreams`, locking the
user out of their own playback. (Confirmed in real traffic: one playback's
`PlaySessionId` repeated 160+ times, each request minting a fresh session.)

## Approach

When the provided `PlaySessionId` is unknown (or owned by another caller), fall
through to the existing CompatToken-scoped `FindByRoute` reuse instead of
erroring, so all of a direct play's requests share one session. Reuse stays scoped
to the caller's own token, so a guessed/foreign id cannot bind another user's
session.

## Scope / what this does and doesn't do

- Fixes the **per-request leak** (N sessions per direct play → 1).
- Does **not** change the cap itself: 6+ *genuinely distinct* titles played within
  the 45s grace still legitimately hit `max_streams` (real concurrency, not a
  leak; `activeCountLocked` has no per-item dedup).
- The fall-through applies to any `/Videos/{id}/stream` (and the subtitle path)
  with an unknown `PlaySessionId`, not strictly `Static=true`. This is safe: the
  play method is recomputed per-request from the resolved source's
  direct-play/stream flags, `ensureUpstreamPlayback` restarts the upstream session
  on a method mismatch, and reuse is CompatToken-scoped. Transcode/HLS manifest +
  segment handlers resolve sessions via `playbackStore.Get` directly and are
  unaffected.

## Testing

- New `TestHandleVideoStream_StaticDirectPlayReusesSessionAcrossRequests` asserts
  `StartSession` runs exactly once across 3 repeated Static requests carrying the
  same client `PlaySessionId`. Mutation-verified (revert → 3 calls → RED).
- Existing Static-serve and no-Static-404 tests still pass.
- `gofmt`/`go vet` clean; full `internal/jellycompat` suite green in the
  libvips-capable container.

---
*AI-assisted: root-caused from production jellycompat request captures and reviewed
across regression / security / effectiveness passes; every line reviewed, and the
tests run and mutation-verified, by me.*

### DIFF
--- a/internal/jellycompat/audio_selection_test.go
+++ b/internal/jellycompat/audio_selection_test.go
@@ -22,6 +22,7 @@ type testCompatSessionManager struct {
 	audioTrackCalls []compatAudioTrackCall
 	progressCalls   int
 	stopCalls       []string
+	startCalls      int
 }
 
 type compatAudioTrackCall struct {
@@ -31,6 +32,7 @@ type compatAudioTrackCall struct {
 }
 
 func (m *testCompatSessionManager) StartSession(userID int, profileID string, fileID int, method playback.PlayMethod, transcodeAudio bool) (*playback.Session, error) {
+	m.startCalls++
 	session := &playback.Session{
 		ID:             "upstream-started",
 		UserID:         userID,
--- a/internal/jellycompat/playback_static_directplay_test.go
+++ b/internal/jellycompat/playback_static_directplay_test.go
@@ -121,6 +121,33 @@ func TestHandleVideoStream_NoStaticNoSessionReturns404(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


Quick104:
### Code review

PR #166, opened 2026-06-16
feat(recommendations): per-profile "Not Interested" to hide items from recommendations

d3v1l1989:
## What & why

Adds a per-profile **"Not Interested"** action so a viewer can hide a movie/show from recommendations — addressing the "you get stuck on the same recommendations unless you mark things watched" problem. Like YouTube's "Not interested" / Spotify's "don't recommend".

Hidden items are excluded from recommendation & discovery surfaces but **remain fully visible in library browsing and search**, and can be managed/un-hidden from a new **Not Interested** page.

## Approach

Mirrors the existing favorites/watchlist wiring rather than introducing new patterns.

**Backend**
- New `user_hidden_recommendations` table (Goose timestamped migration + matching SQLite `userdb` schema).
- Optional `HiddenRecommendationStore` interface (same opt-in pattern as `DeviceRegistry`), implemented on both Postgres and SQLite stores — avoids touching the large `UserStore` interface / all its mocks.
- Exclusion merged into **every** recommendation/discovery path that already excludes watched items: `SignalReader.WatchedItemIDSet` (For You / engine rows + worker cache), `repo.GetWatchedItemIDSet`/`…FromStore` (Because You Watched, Popular, Recently Added, Discover), `SimilarItems` ("More Like This", incl. jellycompat + audiobooks ABS), and a `NOT EXISTS` clause in the three discovery recipe queries (hidden_gems, critically_acclaimed, forgotten_favorites).
- `is_hidden` added to per-item `user_state`; profile-scoped `GET/PUT/DELETE /api/v1/hidden` (+ list); a `user_state.changed` realtime event for cross-device sync. **Additive-only** within `/api/v1`.

**Frontend**
- `useToggleHidden` / `useHiddenList` hooks, a "Not Interested" / "Show in Recommendations" entry in `MediaItemMenu`, a Hidden management page + sidebar entry, and `is_hidden` threaded through the realtime provider.

## Scope note

Library browse + search are intentionally **not** filtered (`query_builder.go` untouched) — hidden items stay searchable. Worker-cached "For You" rows refresh on the next worker pass, so there's a brief staleness window by design.

I'm aware `docs/architecture/v1-scope.md` is **not locked** and this is a new user-facing capability — happy to convert this into a v1 capability proposal / pause for triage if you'd prefer that route. Opening the PR so the implementation is visible alongside the proposal.

## Testing

- Go: `go build ./...`, `go vet`, `go test ./...` in a libvips-capable container — green (one pre-existing flaky `TempDir` cleanup test in jellyfin-compat, unrelated). New table-driven tests for the discovery exclusion.
- `golangci-lint run --new-from-rev=main` — 0 issues.
- Frontend: `tsc`, `eslint` (0 errors), `prettier --check`, `vitest` (17/17).
- `make verify-local-paths` — clean.

## AI-use disclosure

Implemented with AI assistance (Claude). Every line was reviewed and verified by me; the change went through several self-review passes before submission.

## Release Notes

* **New Features**
  * Added "Not Interested" functionality—mark items you don't want to see in recommendations
  * New "Not Interested" page to view and manage all hidden items
  * Added "Not Interested" option to media item action menus
  * Added sidebar navigation link for quick access to hidden items

* **Improvements**
  * Hidden items are now automatically excluded from recommendations and similar-item suggestions


d3v1l1989:
Thanks for the review — addressed in `ac064ec`:

- **Fail-open exclusion (Major):** `SimilarItems`, the audiobooks ABS recommender, and `GetWatchedItemIDSetFromStore` now propagate the watched/hidden lookup error (fail closed) instead of returning items the profile hid/watched. This also makes them consistent with the For-You/`GetWatchedItemIDSet` paths, which already failed closed.
- **Profile-switch cache bleed (Major):** added `["hidden"]` to the query-cache clear on profile switch in `App.tsx`.
- **SQLite index (nitpick):** added `idx_hidden_recommendations_profile_hidden_at (profile_id, hidden_at DESC)` — parity with the Postgres migration.
- **Tests (nitpick):** added the `is_hidden: true` → "Show in Recommendations" label case, and strengthened the discovery test to assert `uhr.user_id`/`uhr.profile_id` scoping predicates.
- **Docstrings:** added doc comments to the new exported store functions.

Re-verified: `go build ./...`, `go test` (libvips container), `golangci-lint --new-from-rev=main` (0 issues), `tsc`, `eslint`, `prettier --check`, `vitest`.

Quick104:
Thanks for putting this together and for exploring the feature end to end. We still think a per-profile **Not interested** signal is worth pursuing, but after review we do not want to continue with this implementation. The branch now conflicts with `main`, and the storage/runtime boundary plus the cross-client product contract need to be redesigned before implementation.

I’ve captured the intended behavior, open design questions, and acceptance criteria in #690 so the idea and the useful work here are not lost. Closing this PR in favor of that issue.

PR #200, opened 2026-06-25
fix(jellycompat): authenticate Android TV token-less direct play

Pukabyte:
## Problem

Stock **Jellyfin Android TV** fails direct play against the jellycompat layer.

Observed sequence (from production logs, one episode, one TV):
1. `POST /Items/{id}/PlaybackInfo` → **200** (user authenticated; a `PlaybackSession` is registered holding the `CompatToken`, `RouteItemID`, and media sources).
2. The app then requests direct play: `GET /Videos/{id}/stream?static=true&container=mkv&tag=…&mediaSourceId=…` — but the request is made by the player's HTTP stack (`okhttp/4.12.0`) with **no auth header, no `api_key`/`ApiKey` query param, and no `PlaySessionId`** (`auth_kind=none`).
3. `PlaybackSessionAuth` has only two ways in — a token, or a `PlaySessionId` lookup — so it returns **401**. The client retries, falls back to a transcode that stalls after a couple of segments, and surfaces *"player error occurred, will try again"*.

The Jellyfin phone app works because it sends `X-Emby-Token`. We already return a `DirectStreamUrl` containing `api_key` from `PlaybackInfo`, but this client ignores it and builds its own token-less URL — slipping past even the existing lowercase-`playSessionId` accommodation (added for Wholphin).

## Fix

Add a third fallback in `PlaybackSessionAuth`, **scoped strictly to the direct-play video stream routes** (`/Videos/{id}/stream` and `/Videos/{id}/stream.{container}`) via the chi route pattern so `/Items/{id}/Download` stays protected. It anchors auth on the `PlaybackSession` negotiated during the preceding (already-authenticated) `PlaybackInfo`, looked up by `mediaSourceId` when present (else the route item id), and resolves its `CompatToken` exactly like the existing `PlaySessionId` branch.

## Security note

This serves a token-less direct-play stream when an unexpired `PlaybackSession` exists for the item. That mirrors the existing token-less HLS exposure (which is `PlaySessionId`-gated); item ids are more guessable, so the lookup prefers `mediaSourceId` to narrow it, and it is gated to the stream routes only. A natural follow-up is to restrict the fallback to sessions negotiated by Android TV (flag set from the `PlaybackInfo` User-Agent) — happy to add that if preferred.

## Tests

`internal/jellycompat/auth_directplay_test.go`:
- token-less direct play **with** a matching session → served (200)
- no matching session → 401
- `/Items/{id}/Download` token-less → still 401 (fallback does not apply)

Verified `go vet` + full `internal/jellycompat` test suite pass in a libvips-capable container; no regressions.

## AI-use disclosure

Investigated and authored with assistance from Claude Code (Claude Opus 4.8).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback authentication for direct-play streaming routes, allowing requests to continue when a matching playback session is already available.
  * Added a fallback path for stream URLs with or without a container suffix, reducing unauthorized failures for valid playback requests.
  * Kept download requests unchanged, so authentication requirements still apply where expected.

### DIFF
--- a/internal/jellycompat/auth.go
+++ b/internal/jellycompat/auth.go
@@ -257,8 +257,10 @@ func PlaybackSessionAuth(sessions *SessionStore, playbackStore *PlaybackSessionS
 			// the PlaybackSession negotiated for this item: a successful PlaybackInfo
 			// already authenticated the user and registered a session holding the
 			// CompatToken. Scope this strictly to the direct-play video stream routes
-			// (NOT /Items/{id}/Download) via the chi route pattern, and prefer
-			// matching on mediaSourceId when present to narrow which item is served.
+			// (NOT /Items/{id}/Download) via the chi route pattern, prefer matching
+			// on mediaSourceId when present, and require the matched session's
+			// RouteItemID to equal the requested item so a source id can't
+			// authorize a stream for a different item.
 			if playbackStore != nil {
 				switch chi.RouteContext(r.Context()).RoutePattern() {
 				case "/Videos/{id}/stream", "/Videos/{id}/stream.{container}":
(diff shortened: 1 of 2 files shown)


Pukabyte:
@coderabbitai review

PR #220, opened 2026-06-25
Add catalog search provider support and admin controls

Quick104:
## Summary
- Add a pluggable catalog search provider with Meilisearch support and PostgreSQL fallback
- Wire search index synchronization into item, metadata, scanner, and cleanup paths
- Add admin settings UI, connection checks, and search status reporting for catalog search
- Include a Meilisearch service in `docker-compose.yml` for local search testing

## Testing
- Not run (not requested)

## Summary by CodeRabbit

* **New Features**
  * Added catalog search settings and a live catalog search status endpoint in the admin UI, including sync/rebuild controls.
  * Enabled Meilisearch-backed catalog search indexing with scheduled outbox syncing and full rebuild support (with fallback/circuit breaker).
* **Bug Fixes**
  * Strengthened Meilisearch settings normalization/validation (including semantic options) and improved Meilisearch connection-check behavior.
  * Improved catalog indexing event reliability for upserts/deletes/renames using transactional enqueueing; improved search pagination/total handling for interactive browsing.
* **UI/Performance**
  * Capped scan-related real-time snapshots and activity display; reduced live-search navigation debounce latency.


Quick104:
## 🔎 Code review — catalog search provider (max-effort pass)

_Automated multi-agent review (Claude Code `/code-review max`) against PR head `fa9ea0b3`: 10 independent finder angles → per-finding verification → gap sweep. AI-generated — please apply your own judgment. Links point at the reviewed commit._

The feature is well-architected — outbox + advisory locks, Postgres fallback, circuit breaker, schema versioning, opt-in via `profiles: [search]`. The items below are the sharp edges. **Config labels:** 🟥 **default (postgres)** = affects every deployment on merge, even ones that never touch Meilisearch; 🟦 **meili-only** = only once an admin enables Meilisearch.

---

### High severity

**1. 🟥 A best-effort cleanup now aborts the scan on a transient enqueue error**
[`internal/scanner/manga_scan.go#L399`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/scanner/manga_scan.go#L399) · same shape in [`internal/metadata/service.go#L4398`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/metadata/service.go#L4398)
The enqueue runs `INSERT … SELECT … WHERE EXISTS(server_settings…)` on every call — it inserts 0 rows under Postgres but still executes. Pre-change, the duplicate-manga cleanup logged and returned `(winner, nil)`; now a transient DB error (conn reset, serialization failure) on that statement propagates and **aborts `findOrCreateMangaSeries` and the scan** — on the *default* provider.
_Fix: keep this path best-effort (log, don't propagate), or gate the enqueue on an in-process "meili active" flag._

**2. 🟥 Per-item enqueue subquery on the scan hot path (performance-first)**
[`internal/catalog/item_repo.go#L542`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/item_repo.go#L542) → [`internal/catalog/search_index_repo.go#L74`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_index_repo.go#L74)
`upsert()` (every scanner's per-item write path) issues an extra `INSERT … WHERE EXISTS (SELECT 1 FROM server_settings WHERE key='catalog.search.provider' AND lower(value)='meilisearch')` per item. A 300k-item scan = 300k extra round-trips + a `server_settings` probe each, all inserting nothing when provider=postgres.
_Fix: cache the provider flag in-process (refresh on the existing settings-change path); make `EnqueueUpsert` a no-op when not Meilisearch, and drop the per-row `WHERE EXISTS` for an unconditional `INSERT` when it is._

**3. 🟦 Outbox poison-pill stalls all catalog-search indexing forever**
[`internal/catalog/search_indexer.go#L140`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_indexer.go#L140) (and the other `MarkFailed` paths)
On *any* per-batch error, `MarkFailed(ctx, ids, err)` is called on the **entire** 500-event batch and the function returns without `MarkProcessed`. There's no max-attempts and no dead-letter; `ListPending` is `ORDER BY id ASC LIMIT 500`, so one permanently-rejected document (lowest id, always in the first 500) re-fails the whole batch every drain and starves every later event indefinitely.
_Fix: per-event error isolation, an `attempts` cap that quarantines poison events, and partial-progress `MarkProcessed`._

**4. 🟦 Circuit breaker never trips on 5xx / malformed responses — every search eats the full timeout**
[`internal/catalog/search_meilisearch_provider.go#L271`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_meilisearch_provider.go#L271)
`shouldTripCircuit` only trips on deadline/cancel/`net.Error`/401/403. A reachable-but-broken Meili (503, 429, or a JSON-decode failure) returns `false`, so the breaker stays closed and **every** catalog search re-hits Meili and pays up to the 800 ms timeout before falling back to Postgres — sustained latency on a user-facing hot path, exactly what the breaker should prevent.
_Fix: trip on 5xx/429 and on decode failures (anything that isn't a clean 4xx query error)._

**5. 🟦 The 800 ms search timeout is reused for bulk indexing**
[`internal/catalog/search_meilisearch_client.go#L96`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_meilisearch_client.go#L96) (client built in `search_indexer.go` `loadClient`)
The indexer builds its client with `settings.Timeout` (default 800 ms, only clamped `>0`), and that single `http.Client.Timeout` applies to every request including `AddDocuments` of up to 500 documents. A large/slow batch POST exceeds 800 ms → `context deadline exceeded` → fails the batch, and via #3 the index can never drain or rebuild at the default timeout.
_Fix: a separate, generous timeout (or ctx-only) for indexing/`WaitTask` vs. the user-facing search query timeout._

**6. 🟦 `folder_repo`/`library_repo` enqueue a delete for *every* candidate id when the guarded DELETE spared some**
[`internal/catalog/folder_repo.go#L568`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/folder_repo.go#L568) · [`internal/catalog/library_repo.go#L520`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/library_repo.go#L520)
The DELETE has a `NOT EXISTS` re-check (a concurrent scan may have re-attached a content_id to another library — see the comment at folder_repo.go:547). But the enqueue loop fires for all `ids`/`orphanIDs` whenever `deleted > 0` (a *count*, not a per-id check), so a still-existing, still-visible item gets removed from the Meili index.
_Fix: `DELETE … RETURNING content_id` and enqueue only the returned ids (the pattern `manga_scan` now uses)._

---

### Medium severity

**7. 🟥 `Delete` & `UpdateMetadata` wrapped in transactions solely to carry the enqueue**
[`internal/catalog/item_repo.go#L831`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/item_repo.go#L831) · [`#L1507`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/item_repo.go#L1507)
Both were single `r.pool.Exec` calls; now each does `Begin`/`Commit` (+ the no-op enqueue) on every call even when Meilisearch is off — 2 extra round-trips each on the default config.
_Fix: keep the single-statement path when the cached provider isn't Meilisearch._

**8. 🟦 Non-atomic enqueues drop events on crash → permanently stale index**
[`internal/scanner/manga_scan.go#L206`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/scanner/manga_scan.go#L206) · [`internal/catalog/folder_repo.go#L570`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/folder_repo.go#L570) · [`internal/metadata/canonicalize.go#L105`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/metadata/canonicalize.go#L105)
Here the mutation commits and the enqueue runs *separately* on the bare pool — unlike the atomic sites ([item_repo.go:874](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/item_repo.go#L874), [library_repo.go:522](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/library_repo.go#L522), `provider_id_integrity.go`, `rebindItemToExistingItem`). A crash between commit and enqueue loses the delete/rename event forever; no periodic reconcile, only manual rebuild.
_Fix: enqueue inside the same transaction as the mutation at every site — the abstraction should own this, not each caller._

**9. 🟦 `Rebuild` never advances `last_processed_event_id`, so pre-rebuild events replay onto the fresh index**
[`internal/catalog/search_indexer.go#L265`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_indexer.go#L265)
`Rebuild` snapshots the DB into a new index but doesn't mark pending events processed. The next `SyncOutbox` replays them — mostly idempotent, except a stale `delete` event for a content_id recreated/indexed during the rebuild will, on replay, **delete the live document**.
_Fix: advance `last_processed_event_id` to the max event id captured by the snapshot (and mark those processed) on rebuild completion._

**10. 🟦 `Rebuild` leaks Meilisearch indexes unboundedly**
[`internal/catalog/search_indexer.go#L227`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_indexer.go#L227)
Each rebuild creates `<index>_rebuild_<unixts>` and flips `active_index_uid` to it, but the prior index is never dropped (the client has no `DeleteIndex`). Every manual rebuild leaves a full copy of the catalog behind in Meili.
_Fix: delete the previous active index after a successful swap (or sweep `*_rebuild_*` older than the active one)._

**11. `/items?search=` `total` silently becomes approximate with no `total_exact` signal**
[`internal/api/handlers/items.go#L257`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/api/handlers/items.go#L257)
`browseResponse` exposes `total`/`has_more` but no `total_exact`. When Meili is active, `result.Total` is the approximation `offset+page+1` (`TotalExact:false`), so the existing field's meaning changes for clients. `/catalog` already surfaces `total_exact` ([catalog.go:57](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/api/handlers/catalog.go#L57)); `/items` doesn't. Per the repo's v1 rule ("never … repurpose … an existing endpoint"), signal it rather than silently changing semantics.
_Fix: add `total_exact` to `browseResponse` (additive) from `result.TotalExact`._

**12. 🟦 Admin status shows "Healthy" while 100% of searches fall back; fallback reason is dropped**
[`internal/catalog/search_meilisearch_provider.go#L288`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_meilisearch_provider.go#L288) + `catalog_resolver.go`
`Status()` defaults `Healthy:true`/`closed` and only flips when the circuit is open — which (per #4) never happens on 5xx. So the Search Settings panel reads green while every query is silently served by Postgres. Compounding it, `fallbackSearch` sets `result.FallbackReason` but the resolver copies only `Items/Total/HasMore/TotalExact` and discards it — no log, metric, or field for silent degradation.
_Fix: derive health from recent call outcomes (not just circuit state), and surface/log `FallbackReason`._

**13. 🟦 Every manga-chapter upsert generates a wasted Meili delete**
[`internal/catalog/search_indexer.go#L147`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_indexer.go#L147)
Chapters are upserted as media_items (→ `EnqueueUpsert`), but `LoadDocumentsByIDs`/`LoadDocumentsAfter` exclude any `manga_chapters.chapter_content_id`. So each chapter's upsert is "not found" → appended to `deleteIDs` → a `DeleteDocuments` task for an id never indexed, every sync.
_Fix: skip enqueue for chapter rows (non-indexable by construction)._

---

### Low severity

**14. 🟦 `WaitTask`'s `taskUID == 0` sentinel collides with Meilisearch's first real task**
[`internal/catalog/search_meilisearch_client.go#L175`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_meilisearch_client.go#L175)
`0` means "no task created," but Meili task UIDs are 0-indexed, so on a fresh instance `Rebuild`'s first `CreateIndex` returns taskUid 0 and `WaitTask(0)` returns immediately without waiting. Currently masked by Meili's FIFO task processing, but fragile.
_Fix: use `-1` (or `*int64` / `(uid, ok)`) for "no task" instead of `0`._

**15. Setting validation in `admin.go` duplicates `CatalogSearchSettingsFromMap`**
[`internal/api/handlers/admin.go#L2143`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/api/handlers/admin.go#L2143)
The inline provider/URL/index/timeout/strategy validation re-encodes rules already in [`search_provider.go`](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_provider.go) `CatalogSearchSettingsFromMap` — which the connection-check path already calls ([admin_settings_checks.go:131](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/api/handlers/admin_settings_checks.go#L131)). Two copies will drift.
_Fix: validate by building the candidate map and calling `CatalogSearchSettingsFromMap`._

---

<details>
<summary><b>Lower-severity cleanups (confirmed, not itemized above)</b></summary>

- **Dead code:** `EnqueueSearchIndexUpsert` ([search_index_repo.go:90](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_index_repo.go#L90)) and `ErrSearchProviderFallback` ([search_provider.go:31](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/search_provider.go#L31)) are unused; the four `MeilisearchProviderConfig` knobs (`BatchSize/CandidateScanCap/DeepOffsetLimit/CircuitCooldown`) are never set by any caller; the `if provider == nil` fallback in `resolveDirectSearchSource` is unreachable.
- **Unused index:** `idx_catalog_search_index_events_content_id` is never queried (write overhead only); the pending index can't serve `ORDER BY id` past the `available_at` range predicate.
- **Reuse:** `compactNonEmptyStrings`, the JSON-over-HTTP `do()`, the advisory-lock helper, and the `MarkFailed` backoff SQL each duplicate existing patterns (`literaryworks` / `embeddings` / `scanqueue` / `movie_match_queue`); frontend `formatStatusDate` duplicates `lib/date`.
- **Altitude:** provider selection is decided per-row in SQL *and* startup-bound in `restart_keys.go` (two sources of truth); the manga-exclusion `NOT EXISTS` SQL and the default-status struct are each duplicated; the task-link triples are hand-written in three places.
- **Minor:** `WaitTask` polls a fixed 100 ms with no backoff; rebuild progress is pinned at 25% for the whole indexing loop.

**Refuted (checked, not a bug):** negative-offset slice panic in `searchMeilisearch` — `Offset` is clamped to `>=0` in [catalog_parser.go:54](https://github.com/Silo-Server/silo-server/blob/fa9ea0b3ed8fe252e15760185699ace88ecb4b69/internal/catalog/catalog_parser.go#L54).
</details>

---
_Suggested priority: the 🟥 default-config items (#1, #2, #7) land on every deployment the moment this merges; #1 + #3 + #4 + #6 are the highest-risk for anyone who enables Meilisearch._

PR #239, opened 2026-06-27
fix(jellycompat): trim episode list to StartItemId for play-from-here queue

Pukabyte:
## Summary
- Parse `StartItemId` query param in `itemsQuery`
- Trim the sorted episode list so it begins at the requested item
- Fixes Android TV "play from here" returning wrong episode (always S0E1/Specials)

## Test plan
- [ ] On Android TV, long-press mid-season episode → Play From Here → confirm correct episode plays
- [ ] Verify episode list without StartItemId is unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * Episode playback requests now start from the requested episode when a start ID is provided, instead of always returning the full sorted list.
  * The app now correctly reads and applies the start-episode parameter from incoming requests, improving “resume from here” behavior.

### DIFF
--- a/internal/jellycompat/handlers_items.go
+++ b/internal/jellycompat/handlers_items.go
@@ -1378,6 +1378,21 @@ func (h *ItemsHandler) writeSeriesEpisodesResponse(w http.ResponseWriter, r *htt
 		return leftSeason < rightSeason
 	})
 
+	// StartItemId: trim the list so it begins at the requested item.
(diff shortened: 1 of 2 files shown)


Quick104:
Closing in favor of a replacement PR that handles the StartItemId-missing case correctly and adds focused regression coverage for Jellycompat play-from-here queue behavior.

PR #264, opened 2026-07-02
feat(observability): persistent per-subsystem file logging via SILO_LOG_FILE

CoffeeKnyte:
Part of #265.

Adds opt-in durable file logging so a server's logs survive container recreation, with optional per-subsystem split files.

## Problem

**Logs vanish when the container is recreated.** Silo only ever wrote its logs to stderr. Under Docker that means the only record of a run lives in the daemon's `json-file` log, which is deleted every time the container is recreated (a redeploy, an image bump, a `docker compose up` after an edit). An operator investigating something that happened "before the last restart" has nothing to look at — the history is gone. There is also no way, short of hand-editing the container entrypoint, to keep logs on a mounted volume.

**No way to isolate one subsystem's logs.** When chasing a specific problem — a Jellyfin-compat client misbehaving, a transcode failing, a scan stalling — an operator has to grep the entire firehose. There is no focused, per-area log to tail.

## Solution

A new opt-in file sink, gated entirely on the `SILO_LOG_FILE` environment variable (empty = today's stderr-only behaviour, unchanged). File logging is operational infrastructure, so it is env-driven and configured before the DB-backed settings, matching the existing `SILO_NODE_NAME` / `SILO_PLUGIN_CACHE_DIR` pattern.

### feat(observability): persistent per-subsystem file logging via SILO_LOG_FILE

- **New `internal/logsink` package (`logsink.go`).** `logsink.Router` is an `slog.Handler` installed as the process base handler. It fans each record to stderr + a combined rotating file, and — with `SILO_LOG_SPLIT` — to a per-subsystem file (`silo-jellycompat.log`, `silo-playback.log`, `silo-api.log`, `silo-scanner.log`, `silo-metadata.log`, `silo-notifications.log`). Rotation uses `lumberjack` (already a dependency, same settings as the jellycompat debug logger). Records are classified by the same `component` signal `opslog` persists — an explicit `component` attr, else the `subsystem:` message prefix — reusing `opslog.InferComponent` (exported in `internal/opslog/handler.go`) so file and DB classification stay in lock-step.
- **Wiring (`cmd/silo/main.go:131`, `buildBaseHandler`).** Delegates to `logsink.New`, reads the `SILO_LOG_*` env knobs (`logSinkOptionsFromEnv`), and returns a closer deferred in `run()`. Chain is unchanged otherwise: `opslog.Handler → logfilter(quiet) → logsink.Router → {stderr, files}`.
- **Reliability: file writes are best-effort (`logsink.go`, `Router.Handle`).** Only the console destination's error may propagate. A file write failure (ENOSPC, read-only mount, rotation error) is swallowed. This matters because the wrapping `opslog.Handler` skips its DB/admin-stream capture whenever its inner handler returns an error — so without this, a full log volume would silently disable the *database* operational-log path too. The console's effectively-never-fails contract is preserved, so logging can't wedge the server.
- **Reliability: bad paths fail fast (`logsink.go`, `ensureWritable`).** `lumberjack` opens files lazily on first write, so a bad `SILO_LOG_FILE` would otherwise surface only at runtime (where slog discards `Handle` errors). `New` now `MkdirAll`s and probe-opens every target path up front and returns an error, so `buildBaseHandler`'s stderr fallback actually engages and logs the reason.
- **Config fidelity.** Explicit `SILO_LOG_MAX_BACKUPS=0` / `SILO_LOG_MAX_AGE_DAYS=0` are passed through verbatim (lumberjack's "keep all" / "never expire") rather than coalesced back to defaults. Multi-word inferred prefixes (`jellycompat autoscan:`) route via their leading token so they still land in the subsystem file. Inline `component` attrs override a static `With("component", …)`, matching `opslog`'s last-write-wins precedence.

This change went through a 7-angle adversarial review; the items above (best-effort file writes, eager path validation, config fidelity, routing fixes) are the confirmed findings from that pass, folded into the initial commit.

## Risk / follow-ups

- **`log_quiet` also suppresses file records.** The router sits below the quiet filter, so quieting a noisy subsystem also empties its split file. This keeps files an exact mirror of the console stream (documented in the package doc). If durable-capture-despite-quiet is wanted, the file sinks would need to tee in before the filter — deferred pending a product call.
- **Per-subsystem routing is best-effort, not exhaustive.** Subsystems that neither set a `component` attr nor use a `subsystem:` message prefix fall to the combined log only. Tagging more subsystem roots explicitly is a follow-up.
- **File records are not redacted.** Like stderr today (no regression), file output is raw; `opslog`'s DB redaction does not apply to console/file sinks. Worth a separate hardening pass if secrets-in-logs is a concern.
- **Runtime wiring not included.** The deployment `docker-compose.yml` still needs `SILO_LOG_FILE`/`SILO_LOG_SPLIT` set and the old tee/FIFO entrypoint removed; intentionally left out of this code-only PR.

## Verification

- `go build ./...` — clean.
- `go vet ./internal/logsink/ ./cmd/silo/ ./internal/opslog/` — clean.
- `gofmt -l` on all touched files — no output.
- `go test ./internal/logsink/ -race` — 6 tests pass (routing to correct files incl. inline-attr and multi-word paths, inline-overrides-static, rotation produces a backup, unwritable path returns an error, closer error propagation, console-only when no file set).
- `go test ./cmd/silo/` — env-parsing tests pass (`envIsTrue`, defaults, overrides incl. explicit-0 retention, invalid-number fallback).

## AI-use disclosure

Implemented with AI assistance (Claude Code): drafting, the multi-agent adversarial review, the resulting fixes, and this description. All changes were reviewed by the author.

## Summary by CodeRabbit

* **New Features**
  * Added optional file-based logging with automatic rotation and per-area log splitting.
  * Logging now falls back to console output if file logging can’t be initialized, so the app still starts.
  * Added broader log category detection for more consistent routing.
* **Bug Fixes**
  * Improved handling of environment settings, including default values, zero values, and invalid numeric inputs.
  * Added safeguards for log cleanup so log files are properly closed and flushed.

### DIFF
--- a/cmd/silo/logsink_env_test.go
+++ b/cmd/silo/logsink_env_test.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"log/slog"
+	"testing"
+)
+
+func TestEnvIsTrue(t *testing.T) {
+	cases := map[string]bool{
+		"1": true, "true": true, "TRUE": true, "yes": true, "  yes  ": true,
+		"": false, "0": false, "false": false, "no": false, "off": false,
+	}
+	for in, want := range cases {
+		t.Run(in, func(t *testing.T) {
+			t.Setenv("SILO_TEST_BOOL", in)
+			if got := envIsTrue("SILO_TEST_BOOL"); got != want {
+				t.Errorf("envIsTrue(%q) = %v, want %v", in, got, want)
+			}
+		})
+	}
+}
+
+func TestLogSinkOptionsFromEnvDefaults(t *testing.T) {
+	// With nothing set, file logging is off and rotation matches DefaultRotation.
+	for _, k := range []string{"SILO_LOG_FILE", "SILO_LOG_SPLIT", "SILO_LOG_MAX_SIZE_MB", "SILO_LOG_MAX_BACKUPS", "SILO_LOG_MAX_AGE_DAYS", "SILO_LOG_COMPRESS"} {
+		t.Setenv(k, "")
+	}
+	opts := logSinkOptionsFromEnv("text", slog.LevelInfo)
+	if opts.File != "" {
+		t.Errorf("File = %q, want empty when SILO_LOG_FILE unset", opts.File)
+	}
+	if opts.Split {
+		t.Error("Split = true, want false when unset")
+	}
+	if opts.Rotation.MaxSizeMB != 50 || opts.Rotation.MaxBackups != 100 || opts.Rotation.MaxAgeDays != 30 || !opts.Rotation.Compress {
+		t.Errorf("default rotation = %+v, want {50 100 30 true}", opts.Rotation)
+	}
+}
+
+func TestLogSinkOptionsFromEnvOverrides(t *testing.T) {
+	t.Setenv("SILO_LOG_FILE", "/var/log/silo/silo.log")
+	t.Setenv("SILO_LOG_SPLIT", "1")
+	t.Setenv("SILO_LOG_MAX_SIZE_MB", "10")
+	// Explicit 0 must be preserved (lumberjack: keep-all / never-expire), not
+	// coalesced back to the defaults.
+	t.Setenv("SILO_LOG_MAX_BACKUPS", "0")
+	t.Setenv("SILO_LOG_MAX_AGE_DAYS", "0")
+	t.Setenv("SILO_LOG_COMPRESS", "false")
+
+	opts := logSinkOptionsFromEnv("json", slog.LevelDebug)
+	if opts.File != "/var/log/silo/silo.log" {
+		t.Errorf("File = %q", opts.File)
+	}
+	if !opts.Split {
+		t.Error("Split = false, want true for SILO_LOG_SPLIT=1")
+	}
+	if opts.Rotation.MaxSizeMB != 10 {
+		t.Errorf("MaxSizeMB = %d, want 10", opts.Rotation.MaxSizeMB)
+	}
+	if opts.Rotation.MaxBackups != 0 {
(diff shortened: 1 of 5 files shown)


darthShadow:
Copied from https://github.com/Silo-Server/silo-server/issues/265#issuecomment-4865162461:

> I would have preferred some discussion regarding this before committing any effort for it. If we are touching the monitoring/observability, it would have been a lot better if we just went the OpenTelemetry ([opentelemetry.io/docs](https://opentelemetry.io/docs/)) way rather than custom implementations and sinks.

CoffeeKnyte:
@darthShadow — fair point, and I'm open to it. To be clear about scope: this PR is deliberately narrow — its only goal is to stop the transient Docker logs (lost on every container recreate) from disappearing, by persisting them to a mounted volume. I mainly just wanted to keep the logs.

Happy to close this PR for now — let's discuss in Discord what foundation we want to use for the logs (OTel or otherwise) before I take it further.

CoffeeKnyte:
Closing in favour of #290. Per @darthShadow's feedback (https://github.com/Silo-Server/silo-server/issues/265#issuecomment-4865162461) to build on OpenTelemetry rather than a custom file sink, we dropped the bespoke rotating-file approach here and reworked it: #290 adds opt-in OpenTelemetry logs + traces (default-off, stderr + opslog unchanged), moves durable retention/rotation to the collector/backend and the container runtime, and — beyond the original scope — adds secret redaction on every sink plus a repo-wide slog context standardization with a lint gate.

Durability (the original goal here) is now handled by OTLP export + Docker json-file rotation on a mounted volume, documented in docs/architecture/observability.md.

PR #35, opened 2026-06-01
fix(metadata): don't promote season/episode 404s to item-level stale provider IDs

fluxis:
## Summary

The main-refresh path in `MetadataService.processInternal` passes `provider404s` to `handleProvider404` for **season** and **episode** fetch errors. Since `handleProvider404` records the item's provider ID as stale whenever `provider404s != nil`, a 404 fetching seasons or episodes promotes the *item* itself to stale — even though the 404 is a child-resource failure (TMDB returns 404 for `/tv/{id}/season/0` when the series has no specials season; for episodes that haven't aired yet; etc.).

The equivalent **target**-refresh paths (`fetchTargetSeasonResults` and `fetchTargetEpisodeResults`) already pass `nil` for this exact reason. This PR brings the main-refresh path in line with the existing target-refresh pattern.

## Diff

Two call sites in `internal/metadata/service.go`:

| Line | Resource | Before | After |
|---|---|---|---|
| 1140 | Seasons (main refresh) | `handleProvider404(provider404s, ...)` | `handleProvider404(nil, ...)` |
| 1173 | Episodes (main refresh) | `handleProvider404(provider404s, ...)` | `handleProvider404(nil, ...)` |

For reference, the already-correct sites that this fix matches:
- `service.go:2305` — `fetchTargetSeasonResults`, already passes `nil`
- `service.go:2339` — `fetchTargetEpisodeResults`, already passes `nil`

## Observed impact

On our deployment, the bug caused **13,524 series items** to have their `media_items.tmdb_id` cleared and recorded as stale during a single library refresh sweep — every series whose TMDB entry doesn't have a "specials" season (most anime, many older shows) was affected. Once recorded, `suppressRecordedStaleProviderIDs` locked the items out of future TMDB enrichment, so subsequent refreshes couldn't self-heal even when TVDB supplied a working TMDB cross-reference.

The bug is fully reproducible. Live log capture during a refresh sweep:

```
provider=tmdb content_id=... season=0 provider_id=62913 error="...tmdb: HTTP 404..."
provider=tmdb content_id=... season=0 provider_id=72511 error="...tmdb: HTTP 404..."
provider=tmdb content_id=... season=0 provider_id=75778 error="...tmdb: HTTP 404..."
```

Every single 404 has `season=0`. Direct curl tests confirm:

```
id=68417: GET /tv/68417 → 200   /season/0 → 404   /season/1 → 200
id=76645: GET /tv/76645 → 200   /season/0 → 404   /season/1 → 200
[...]

PR #276, opened 2026-07-02
fix(metadata): stop manual rematch from resurrecting recorded stale IDs

### LINKED ISSUE #268: Stale External IDs: manual rematch to a different provider re-records the stale ID instead of clearing it
## Summary

**Summary:**  
When an item with a stale TMDB ID is manually rematched to a correct TVDB result, it does not disappear from “Stale External IDs”. Instead, `last_seen` is updated and the item moves to the top of the list again.

**Steps to Reproduce:**
1. Go to Admin Libraries → Stale External IDs.
2. Pick an item with a stale TMDB ID, e.g. `Formula 1` with `tmdb=324880`.
3. Click `Match`.
4. Select the correct TVDB result.
5. Click `Apply Match`.
6. Check the stale list again.

**Expected Result:**  
The item should disappear from the stale list after a successful rematch, especially when the corrected match only has a valid TVDB ID and no valid TMDB ID.

**Actual Result:**  
The item remains in the stale list. `last_seen` is updated to the current time, causing the row to move to the top.

**Likely Cause:**  
The newer `/admin/items/{id}/match/apply` flow does not appear to clear `stale_media_ids` the way the older stale-rematch endpoint does. Existing durable provider IDs also seem to be loaded during `ModeIdentify`, so the old stale TMDB ID may be carried forward even after selecting a TVDB-only match.

**Example:**  
`Formula 1` has a valid TVDB match, but the old `tmdb=324880` appears to be reused and marked stale again.

## Reply digest

When an item with a stale TMDB ID is manually rematched to a correct result from a different provider, it does not leave the "Stale External IDs" admin list; its `last_seen` is refreshed and it jumps back to the top.

Repro: Admin Libraries -> Stale External IDs -> pick an item with a stale TMDB id (e.g. `Formula 1`, tmdb=324880) -> Match -> select the correct TVDB result -> Apply Match -> the item is still in the stale list, now at the top.
Expected: after a successful rematch to a different provider the item should disappear from the stale list.

Root cause verified in code: the current "Apply Match" UI flow (`HandleApplyItemMatch`) runs metadata `Process` in `ModeIdentify`, which (a) re-injects the durable stale tmdb id via `prepareProcessRequest`, and (b) skips `suppressRecordedStaleProviderIDs` because that call is guarded `if req.Mode != ModeIdentify`. So the Phase-2 fetch re-queries the stale id, 404s again, and the post-persist `DeleteByContentID` + re-`Upsert(provider404s)` block (added in PR #38) re-inserts the stale row with a fresh `last_seen_at`. The stale list is ordered `last_seen_at DESC`, which is why the row resurfaces at the top.

## Code references

Verified STILL OPEN at origin/main (46a6e86d) as of 2026-07-02. The deprecated HandleRematchStaleID path clears correctly (wipes IDs + DeleteByContentID), but the live admin_match.go apply flow does not.
- internal/metadata/service.go:1094: `if req.Mode != ModeIdentify` excludes the manual apply/identify path from suppressRecordedStaleProviderIDs, so the recorded-stale provider id is not stripped before the Phase-2 fetch and 404s again
- internal/metadata/service.go:851-868 (prepareProcessRequest): re-injects the durable stale tmdb id into the identify request, re-arming the 404
- internal/metadata/service.go:1293-1310: after a successful identify, DeleteByContentID + re-Upsert(provider404s) re-creates the stale row with a fresh last_seen_at instead of leaving it deleted
- internal/metadata/stale_media_id_repo.go:120-131 (ListAll) + internal/api/handlers/libraries.go:2114 (HandleListStaleIDs): list ordered by last_seen_at DESC, so the refreshed row surfaces at the top

## Attachments

_None._

## Source

Imported from Discord:
- https://discord.com/channels/1487838737248292864/1522156068023566477

fluxis:
## Problem

When an item with a stale provider ID (e.g. `Formula 1`, tmdb=324880) is manually rematched to a correct result from a different provider via Admin → Match → Apply Match, it never leaves the Stale External IDs list — its `last_seen` refreshes and it jumps back to the top.

Mechanism: the Apply Match flow runs `Process` in `ModeIdentify`. `prepareProcessRequest` re-injects the item's durable provider IDs — including the known-dead tmdb ID — into the request without consulting `stale_media_ids`. The Phase-2 fetch 404s on it again, and the post-persist stale-row refresh (`DeleteByContentID` + re-`Upsert` of this run's 404s) re-creates the stale row with a fresh `last_seen_at`. The list is ordered `last_seen_at DESC`, hence the row resurfacing at the top.

## Approach

Filter recorded-stale IDs out of the **injected durable set** in `prepareProcessRequest` (reusing `suppressRecordedStaleProviderIDs`), before it merges into the request. Two properties fall out deliberately:

- Caller-supplied IDs in `req.ProviderIDs` are never filtered, so an admin explicitly re-selecting a previously-stale ID (e.g. the provider fixed it) still retries it. This is also why the existing `Mode != ModeIdentify` suppression guard in `processInternal` stays untouched.
- With no 404 recorded, the existing post-persist block now clears the item's stale rows and re-records nothing — the item drops off the stale list after a successful rematch, matching the (deprecated) `HandleRematchStaleID` behavior.

For non-identify modes the filtering is redundant-but-harmless: those paths already suppress recorded-stale IDs later in `processInternal`.

Two regression tests via `ProcessWithProviders`: one reproduces the full #268 scenario (stale tmdb + rematch to tvdb → request must not carry tmdb, stale rows must be empty after persist; failed before the fix with the stale row re-upserted), and one pins the exception (user re-selecting the recorded-stale ID keeps it).

## Risks / follow-ups

Behavior change is limited to durable-ID injection when a matching `stale_media_ids` row exists. Scheduled/manual refreshes already suppressed these IDs downstream, so no observable change there.

Fixes #268

## AI-use disclosure

Implemented with Claude Code (test-first); human-reviewed before submission.

### DIFF
--- a/internal/api/handlers/admin_match.go
+++ b/internal/api/handlers/admin_match.go
@@ -161,7 +161,11 @@ func (h *AdminMatchHandler) HandleApplyItemMatch(w http.ResponseWriter, r *http.
 		return
 	}
 
-	if len(req.ProviderIDs) == 0 {
+	// Normalize keys and values the same way the search endpoint does so a
+	// caller-supplied key like "TMDB" or " tmdb " cannot bypass downstream
+	// provider-id handling (e.g. stale-ID suppression in metadata.Process).
+	providerIDs := normalizeMatchProviderIDs(req.ProviderIDs)
+	if len(providerIDs) == 0 {
 		writeError(w, http.StatusBadRequest, "bad_request", "At least one provider ID is required")
 		return
 	}
@@ -191,7 +195,7 @@ func (h *AdminMatchHandler) HandleApplyItemMatch(w http.ResponseWriter, r *http.
 
 	result, err := h.metadata.Process(r.Context(), metadata.ProcessRequest{
 		ContentID:   contentID,
-		ProviderIDs: req.ProviderIDs,
+		ProviderIDs: providerIDs,
 		FolderID:    folderIDStr,
 		Mode:        metadata.ModeIdentify,
 	})
--- a/internal/api/handlers/admin_match_test.go
+++ b/internal/api/handlers/admin_match_test.go
@@ -326,6 +326,67 @@ func TestAdminMatchApply_RequiresProviderIDs(t *testing.T) {
 	}
 }
 
+func TestAdminMatchApply_NormalizesProviderIDKeys(t *testing.T) {
+	items := &fakeMatchItemLookup{
+		items: map[string]*models.MediaItem{
+			"item-5": {ContentID: "item-5", Title: "Test", Type: "movie"},
+		},
+	}
+	metaSvc := &fakeMatchMetadataService{
+		processResult: &metadata.ProcessResult{ContentID: "item-5", Updated: true},
+	}
+	h := NewAdminMatchHandler(items, nil, metaSvc)
+	router := buildMatchRouter(h)
+
+	body, _ := json.Marshal(matchApplyRequest{
+		ProviderIDs: map[string]string{
+			" TMDB ": " 157336 ",
+			"blank":  "   ",
+		},
+	})
+	req := httptest.NewRequest(http.MethodPost, "/admin/items/item-5/match/apply", bytes.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	rec := httptest.NewRecorder()
+
+	router.ServeHTTP(rec, req)
+
+	if rec.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", rec.Code, rec.Body.String())
+	}
+	if got := metaSvc.lastProcess.ProviderIDs["tmdb"]; got != "157336" {
+		t.Errorf("Process provider_ids[tmdb] = %q, want 157336 (keys/values must be normalized)", got)
+	}
+	if _, ok := metaSvc.lastProcess.ProviderIDs[" TMDB "]; ok {
+		t.Errorf("Process received raw key %q, want normalized keys only: %v", " TMDB ", metaSvc.lastProcess.ProviderIDs)
+	}
+	if _, ok := metaSvc.lastProcess.ProviderIDs["blank"]; ok {
+		t.Errorf("blank provider IDs should be dropped, got %v", metaSvc.lastProcess.ProviderIDs)
+	}
+}
+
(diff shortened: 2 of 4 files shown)


Quick104:
### Code review

Found 1 issue:

1. The new suppression can be defeated by provider-key casing/whitespace. `suppressRecordedStaleProviderIDs` lowercases and trims the stored stale row's provider before the map lookup, but the map keys it checks are used as-is. `HandleApplyItemMatch` passes `req.ProviderIDs` straight from the JSON body into `Process` without normalizing (only the search endpoint runs `normalizeMatchProviderIDs`), and durable IDs are stored with whatever casing they arrived with — so a key like `TMDB` or ` tmdb ` slips past the lookup and the stale ID is re-injected after all. This is the same gap flagged in PR #182's review ("Normalize provider IDs before applying matches — apply still passes `req.ProviderIDs` directly into `Process`"); normalizing the apply map (or normalizing keys inside `suppressRecordedStaleProviderIDs` itself) would close it. (prior PR feedback, PR #182)

https://github.com/Silo-Server/silo-server/blob/fa72747e31e4e6936fd697244c3b77dc88b57007/internal/metadata/service.go#L827-L831

https://github.com/Silo-Server/silo-server/blob/fa72747e31e4e6936fd697244c3b77dc88b57007/internal/api/handlers/admin_match.go#L192-L197

Otherwise the fix verified cleanly: it correctly filters only the injected durable set, preserves the deliberate `ModeIdentify` retry exception, doesn't disturb the `provider404s` gating from earlier stale-ID fixes (#35/#37/#38), and both regression tests pin the intended behavior.

Quick104:
Addressed the review finding in 515d24552299bf1f9975e1932edff5b7f2e5209a:
- `HandleApplyItemMatch` now normalizes `req.ProviderIDs` via the existing `normalizeMatchProviderIDs` (same semantics as the search endpoint) and returns the existing 400 when nothing non-blank remains — closing the PR #182 gap.
- `suppressRecordedStaleProviderIDs` now matches keys case-insensitively/trimmed regardless of caller.

New tests: stale-ID suppression despite `"TMDB "` key casing, apply-endpoint key normalization, and all-blank rejection. Both packages vet/test clean.

PR #278, opened 2026-07-02
fix(metadata): re-fetch titles/overviews when a library's metadata language changes

### LINKED ISSUE #211: Changing a library's metadata language doesn't re-fetch titles/overviews
## Summary

Goal:
  Change a library's metadata language (e.g. correct a library from Chinese to Danish) and have its existing items re-fetch their titles/overviews in the new language on the next metadata refresh.

  Steps:
  1. Have a library (media folder) whose items were matched at some point — items get a per-item metadata language stamped at first match.
  2. In Admin → Libraries, change that library's Metadata language to a new value (e.g. da).
  3. Trigger a metadata refresh on the library (quick or full) — or on an individual item via "Refresh metadata".
  4. View an item in that library.

  Expected:
  Items re-fetch from the provider in the library's current metadata language and their displayed title/overview update to it.

  Actual:
  The displayed title/overview stay in the old language. Re-matching and repeated refreshes never fix it. The only way to change it is editing the database directly. Newly-added items in the same library do get the new language, so the library shows a mix.

  What is slow/broken:
  Metadata language. Specifically: an item's stored language (media_items.default_metadata_language) is set once at initial match and is never updated afterward — not by a library metadata-language change, not by a quick or full refresh, not by a per-item refresh. A refresh in the new language writes into a side
  localization instead of replacing the visible title.

  Scope:
  - Happens every time, for every already-matched item; reproducible across all libraries.
  - Quick-mode library refresh makes it worse to notice: it skips items that already have an overview/poster, so a wrongly-languaged-but-otherwise-complete item is never even revisited.

  Version/branch:
  - Branch: main
  - Commit: 69c1e6c

## Reply digest

Changing a library's metadata language (e.g. Chinese→Danish) does not cause existing items to re-fetch titles/overviews in the new language on refresh.

## Code references

Verified against code 2026-06-25 (merged PRs + source). Verified not fixed (silo-server).
- internal/api/handlers/libraries.go: HandleUpdateLibrary only enqueues a re-scan when Paths change; no trigger when MetadataLanguage changes -> existing items aren't re-fetched

## Attachments

_None._

## Source

Imported from Discord:
- https://discord.com/channels/1487838737248292864/1512905656359718932

fluxis:
## Problem

An item's `default_metadata_language` was stamped once at first match and never updated. Changing a library's metadata language did nothing for existing items: the canonical-language pin in `mergeAndPersist` routed any refresh in a different language into the localization side-tables, the upserts' `COALESCE` kept the old stamp forever, quick-mode library refresh skipped complete items entirely, and `HandleUpdateLibrary` triggered nothing on a language change. Items stayed in the old language no matter how often the admin refreshed or re-matched (#211).

## Approach — four coupled changes

1. **Trigger**: `HandleUpdateLibrary` enqueues a quick library metadata refresh when the metadata language changes (mirrors the existing paths-change rescan trigger; conflict with an already-queued refresh is tolerated).
2. **Lister**: quick-mode refresh includes complete items whose stamped language differs from the library's configured language — previously they were never revisited.
3. **Adoption**: a folder-scoped **manual** refresh whose resolved library language differs from the item's stamp sets `ProcessRequest.AdoptLanguage`; `mergeAndPersist` then treats the library language as the new canonical — base-row title/overview are rewritten and the item restamped — instead of shunting the fetch into localization tables. Gated to `ModeManualRefresh` only: scheduled refreshes merge fill-empty and would restamp the language without rewriting the text.
4. **SQL pins**: the five `COALESCE` clauses that made `default_metadata_language` write-once (media_items, seasons ×2, episodes ×2) now prefer the incoming non-empty value. Every existing caller sends the unchanged stamp or an empty string, so behavior is unchanged outside adoption — and the restamp is atomic with the successful canonical write (no mislabeled rows if a provider fetch fails mid-refresh).

Viewers preferring the old language still get localized text via the existing `LocalizeItemModel` overlay; the base row simply follows the library setting — which is what the setting claims to mean.

## Testing

- Unit tests on the canonical pin (adopt rewrites+restamps; without the flag the old pin behavior is preserved) and the adoption decision table.
- DB-backed tests (`SILO_TEST_DATABASE_URL`) proving the upsert restamp semantics on all three tables and the quick-lister language-mismatch predicate.
- Handler test proving the language-change trigger fires exactly once (and not on rename or same-language updates).
- **Live-validated** on a production deployment: one-movie test library flipped en→da→de→en — each flip auto-enqueued the refresh job and restamped the item (`default_metadata_language` followed each change; this exact flow was a no-op before the fix).

## Risks / notes

- An item shared by two libraries with different languages will follow whichever library was refreshed last (previously it was pinned to first-match language forever). Deemed acceptable: the localization overlay still serves per-viewer languages.
- `ProcessRequest` gains a field — additive, internal type.

Fixes #211

## AI-use disclosure

## Summary by CodeRabbit

* **New Features**
  * Library updates can now trigger a quick metadata refresh when the library language changes.
  * Manual folder refreshes can now adopt the folder’s language as the item’s main language for cleaner updates.

* **Bug Fixes**
  * Quick refresh now includes items whose stored language differs from the folder language.
  * Language values are now preserved more consistently during item, season, and episode updates, avoiding unexpected resets.

### DIFF
--- a/internal/adminjob/library_refresh.go
+++ b/internal/adminjob/library_refresh.go
@@ -89,6 +89,7 @@ func (l *PGLibraryRefreshItemLister) ListLibraryItems(ctx context.Context, libra
 	query := `
 		SELECT mi.content_id, COALESCE(mi.tmdb_id, ''), COALESCE(mi.tvdb_id, ''), COALESCE(mi.imdb_id, '')
 		FROM media_item_libraries mil
+		JOIN media_folders f ON f.id = mil.media_folder_id
 		JOIN media_items mi ON mi.content_id = mil.content_id
 		WHERE mil.media_folder_id = $1`
 	args := []any{libraryID}
@@ -122,6 +123,11 @@ func (l *PGLibraryRefreshItemLister) ListLibraryItems(ctx context.Context, libra
 				SELECT 1
 				FROM stale_media_ids smi
 				WHERE smi.content_id = mi.content_id
+			)
+			OR (
+				COALESCE(mi.default_metadata_language, '') <> ''
+				AND LOWER(TRIM(mi.default_metadata_language))
+					<> LOWER(TRIM(COALESCE(NULLIF(f.metadata_language, ''), 'en')))
 			)
 		  )`
 	}
--- a/internal/adminjob/library_refresh_language_test.go
+++ b/internal/adminjob/library_refresh_language_test.go
@@ -1,0 +1,93 @@
+package adminjob
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// TestQuickRefreshListsLanguageMismatchedItems covers the listing half of
+// issue #211: quick-mode library refresh must include items whose stamped
+// default_metadata_language differs from the library's configured metadata
+// language, even when the item is otherwise complete (overview, poster,
+// backdrop all present) — otherwise a library language change never revisits
+// already-complete items.
+func TestQuickRefreshListsLanguageMismatchedItems(t *testing.T) {
+	dsn := os.Getenv("SILO_TEST_DATABASE_URL")
+	if dsn == "" {
+		t.Skip("SILO_TEST_DATABASE_URL is not set")
+	}
+	ctx := context.Background()
+	pool, err := pgxpool.New(ctx, dsn)
+	if err != nil {
+		t.Fatalf("connect test database: %v", err)
+	}
+	t.Cleanup(pool.Close)
+
+	suffix := time.Now().UnixNano()
+	mismatchID := fmt.Sprintf("lang-mismatch-%d", suffix)
+	matchedID := fmt.Sprintf("lang-matched-%d", suffix)
+
+	var folderID int
+	if err := pool.QueryRow(ctx, `
+		INSERT INTO media_folders (type, name, enabled, metadata_language)
+		VALUES ('movies', 'Lang Refresh Test', true, 'da')
+		RETURNING id
+	`).Scan(&folderID); err != nil {
+		t.Fatalf("seed folder: %v", err)
+	}
+	t.Cleanup(func() {
+		_, _ = pool.Exec(ctx, `DELETE FROM media_items WHERE content_id = ANY($1)`, []string{mismatchID, matchedID})
+		_, _ = pool.Exec(ctx, `DELETE FROM media_folders WHERE id = $1`, folderID)
+	})
+
+	for _, row := range []struct {
+		id, lang string
+	}{
+		{mismatchID, "zh"},
+		{matchedID, "da"},
+	} {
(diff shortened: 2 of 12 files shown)


Quick104:
### Code review

Found 2 issues:

1. Locked title/overview fields defeat the adoption restamp. In `mergeAndPersist`, `item.DefaultMetadataLanguage = canonicalLanguage` is applied unconditionally when `AdoptLanguage` is set, while the title/overview merges above are gated per-field by `isFieldLocked`. An item with a locked title keeps its old-language text but gets restamped to the new language, so the quick-refresh mismatch predicate in `library_refresh.go` never flags it again — it's permanently mislabeled. Consider only restamping when the language-bearing fields were actually adoptable, or accepting the restamp deliberately with a comment. (bug)

https://github.com/Silo-Server/silo-server/blob/d4e8a1a5bb45a6fb00da30778c64a49b0646d537/internal/metadata/service.go#L1509-L1516

2. Shared items flip-flop between libraries with different languages. `media_item_libraries` allows one item in multiple libraries, but the adoption gate (`adoptableFolderLanguage` + single-language resolution) only looks at the requesting folder's library. An item in library A (en) and library B (de) will have its canonical base-row title/overview rewritten to whichever library's refresh ran last, oscillating on every language-change refresh instead of settling on a stable canonical language. (bug)

https://github.com/Silo-Server/silo-server/blob/d4e8a1a5bb45a6fb00da30778c64a49b0646d537/internal/metadata/service.go#L528-L541

Also noting (minor, likely acceptable scoping): `refreshSeriesChildTarget` still pins season/episode-scoped manual refreshes to `series.DefaultMetadataLanguage` without the adoption logic, so a manual season/episode refresh racing ahead of the automatic library-wide refresh still writes the old language; it self-heals once the series is restamped.

The COALESCE inversions checked out: applied consistently across all three repos, no historical fix is being reverted (the old direction dates to the initial migration), and every existing caller sends either the unchanged stamp or empty, matching the commit message's claim.

Quick104:
Addressed both review findings in 4dbf2f36:
- **Locked fields**: `mergeAndPersist` now skips the language restamp when both title and overview are field-locked (a single locked field still adopts, since the other is genuinely rewritten). The skipped case falls back to the existing stamp, so `isCanonicalWrite` routes the fetch into the localization tables exactly like a non-adopting refresh does today.
- **Multi-library flip-flop**: adoption in `Process` is now additionally gated on `itemLibrariesAgreeOnLanguage`, reusing the existing `GetDistinctMetadataLanguagesForItem` query (which applies the same empty→`en` default as `resolveFolderLanguage`). Libraries that disagree, lookup errors, or a nil repo all skip adoption and keep the stable stamp.

New tests cover both-locked (stamp unchanged), title-only-locked (stamp adopted, locked title preserved), and a table test for the agreement gate. All metadata/adminjob/catalog tests pass, including the DB-backed restamp tests against a real migrated scratch Postgres.

PR #279, opened 2026-07-02
fix(playback): stop seek-restart churn and premature paused-session reaping

### LINKED ISSUE #243: Playback freezes / won't resume on player state changes (resume after long pause, intro-skip, manual seek)
## Summary

Silo tvOS 1.0.0
Build 1

After pausing tv show or movie for more then 5 mins
App freezes and doesn’t start back playing after hitting play/Pause on remote or on tvOS screen.

Causing you to force close app out and going back in to continue watching where you left off.

## Reply digest

Two reports of the player freezing when playback state changes. (1) After pausing a show/movie for >5 min, playback won't resume — play/pause via remote or on-screen does nothing; the app must be force-closed and reopened to continue (tvOS 1.0.0 build 1). A second user confirms it **also happens on the web client**, so this is not tvOS-only. (2) Pressing the Intro-Skip button lags and freezes the app, and manually seeking forward near the start buffers and freezes the entire app (tvOS TestFlight 1.0.0 build 1). Same reporter/timeframe; both are 'player freezes on a state change' and likely share a root cause. Confirmed cross-client (web + tvOS).

## Code references

Not yet localized in code as of 2026-06-29. Confirmed cross-client (web + tvOS), so most likely server-side playback/session resume handling; a tvOS client component is also possible. Merge of two Discord reports.

## Attachments

_None._

## Source

Imported from Discord:
- https://discord.com/channels/1487838737248292864/1520884395182002248
- https://discord.com/channels/1487838737248292864/1520636049494118451

fluxis:
## Problem

Cross-client playback freezes (web + tvOS) on three state changes: resume after a long pause, intro-skip, and manual seek near stream start (#243). Investigation found two independent latent defects, both present since the initial migration:

1. **Dueling transcode restarts.** `TranscodeSession.Restart` had no concurrency guard, and the segment-recovery decision told concurrent requests to restart again while a restart was already in flight — the restart window runs with `running=false`, so requests hit the restart-happy `transcode_not_running` branch before the `Restarting` case was ever reached. Pipelined HLS segment requests therefore spawned restarts that kept killing the ffmpeg the player was waiting on: a 30s stall, another restart, surfacing as buffer-then-freeze. Near stream start (tiny produced head) almost any forward seek entered this path, and intro-skip is the same code path.
2. **Paused sessions reaped after 2 minutes**, killing the transcode with no revival path — every subsequent manifest/segment/progress request 404s. Clients that stop progress-reporting while paused (tvOS; backgrounded/slept web tabs) froze on play after a >5 minute pause.

## Approach

- `Restart` is now **single-flight**: a second caller during an in-flight restart returns immediately as a no-op instead of killing the process the first restart just started.
- `SegmentRecoveryDecision` checks `Restarting` **before** `Running` and returns wait-not-restart (`WaitForSegment` already polls politely through a restart window).
- Paused grace raised **2m → 30m** in both reapers (in-memory `SessionManager` and the DB `SessionCleaner`), so an intentional pause survives; abandoned sessions still reap.
- Post-restart re-arming (throttler + exit monitor) moved into a **restart hook** installed at session creation and fired by `Restart` itself — previously the jellycompat seek path re-armed nothing and the audio-switch path re-armed only the throttler; now all three callers behave identically.

## Testing

- Unit tests: wait-during-restart decision, single-flight guard, restart hook firing, paused-session survival/reap boundaries.
- **Live-validated** on a production deployment (QSV transcode, real media):
  - 3 concurrent far-seek segment requests → exactly **one** ffmpeg restart (`restart_count=1`); the concurrent request hit the new `transcode_restarting` wait branch and got its segment in 1.4s. Old code restarted for each.
  - Session paused then silent for 2m50s (past the old 2m reap) → session row survived, segment fetch on resume returned 200, resume progress accepted. Old code deleted the session at 2:00 and every resume request 404ed.

## Risks / follow-up

- A paused transcode now holds its ffmpeg process and output segments for up to 30 minutes instead of 2 — bounded by max-streams admission limits.
- Follow-up (not in this PR): lazily revive a reaped session's transcode on resume instead of 404ing, so even multi-hour pauses recover. Worth its own issue.

Fixes #243

## AI-use disclosure

### DIFF
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -1778,10 +1778,9 @@ func (h *PlaybackHandler) HandleChangeAudioTrack(w http.ResponseWriter, r *http.
 			ts.SetAudioTrackIndex(req.AudioTrackIndex)
 			seekSeconds := req.Position
 			startSegment := computeStartSegment(seekSeconds, ts.Opts().SegmentDuration)
+			// Throttler + exit monitor re-arm via the session's restart hook.
 			if restartErr := ts.Restart(context.WithoutCancel(r.Context()), seekSeconds, startSegment); restartErr != nil {
 				slog.Error("failed to restart transcode for audio switch", "session", sessionID, "error", restartErr)
-			} else {
-				h.maybeStartThrottler(r.Context(), ts)
 			}
 		}
 	}
@@ -2211,6 +2210,14 @@ func (h *PlaybackHandler) HandleStartTranscode(w http.ResponseWriter, r *http.Re
 	h.transcodes[req.SessionID] = transcodeSession
 	h.transcodeMu.Unlock()
 
+	// Re-arm the throttler and exit monitor after every restart, no matter
+	// which caller triggered it (web segment recovery, audio switch, or
+	// jellycompat seek — the compat path has no handler access at all).
+	transcodeSession.SetRestartHook(func(ctx context.Context) {
+		h.maybeStartThrottler(ctx, transcodeSession)
+		h.monitorLocalTranscodeExit(req.SessionID, transcodeSession)
+	})
+
 	h.maybeStartThrottler(r.Context(), transcodeSession)
 	h.monitorLocalTranscodeExit(req.SessionID, transcodeSession)
 
@@ -2395,8 +2402,8 @@ func (h *PlaybackHandler) HandleGetTranscodeSegment(w http.ResponseWriter, r *ht
 						seekSeconds,
 						segNum,
 					); restartErr == nil {
-						h.maybeStartThrottler(r.Context(), transcodeSession)
-						h.monitorLocalTranscodeExit(sessionID, transcodeSession)
+						// Throttler + exit monitor re-arm via the session's
+						// restart hook.
 						segmentPath, err = transcodeSession.WaitForSegment(segmentName, 30*time.Second)
 						if err == nil && strings.EqualFold(transcodeSession.Opts().TargetCodecVideo, "copy") {
 							// Copy-mode seeks can resume as soon as the target segment
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -2210,9 +2210,11 @@ func (h *PlaybackHandler) HandleStartTranscode(w http.ResponseWriter, r *http.Re
 	h.transcodes[req.SessionID] = transcodeSession
 	h.transcodeMu.Unlock()
 
-	// Re-arm the throttler and exit monitor after every restart, no matter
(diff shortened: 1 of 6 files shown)


Quick104:
### Code review

Found 2 issues (both advisory — the core concurrency changes verified clean):

1. The new comment overclaims jellycompat coverage. The restart hook is installed only in `HandleStartTranscode`, but jellycompat creates its transcode sessions through its own `playback.StartTranscode` call in `internal/jellycompat/streams.go` and calls `.Restart()` directly — those sessions never receive the hook, so the compat seek path still re-arms nothing. Not a regression (that path never re-armed pre-PR either), but the comment asserting re-arm happens "no matter which caller triggered it (… or jellycompat seek …)" will mislead future maintainers; either wire the hook at the jellycompat creation site or scope the comment to handler-created sessions. (code comment guidance)

https://github.com/Silo-Server/silo-server/blob/1b16cc0aefdb993c06f3fd78c15035dff8731549/internal/api/handlers/playback.go#L2213-L2217

2. Coordination with #174: the open `feat/unified-playback-reconstruct` branch introduces a competing manager-level restart lock (`RestartSessionLocked`) over the same function and same three files this PR touches (`transcode.go`, `session.go`, handlers `playback.go`), including jellycompat call-site coverage. Neither PR references the other; whichever merges second faces a non-trivial reconcile, so worth sequencing deliberately. (historical git context)

Verified clean: the single-flight `restarting` guard is race-free (checked/set under `s.mu`, all exit paths reset it, hook fires after unlock), the `Restarting`-before-`!Running` reorder fixes a genuine latent bug present since the initial migration, all five `Restart` call sites still behave correctly, both grace constants agree at 30m, and no prior fix (#52's `RestartOnTimeout` gating, #262's session liveness) is regressed.

Quick104:
Addressed the review finding in 95327c90bd193d9711796c7a79115b8bdd299bc8 (comment-only, as suggested): verified that jellycompat keeps its own transcode registry (`internal/jellycompat` `PlaybackHandler.transcodes`) and its `Restart()` calls never touch handler-created sessions, and that the throttler/exit monitor exist only in the API handler — so wiring the hook there would re-arm nothing. The comment in `HandleStartTranscode` and the test doc comment now scope the guarantee to handler-created sessions and state explicitly that jellycompat-created sessions never had this wiring.

Bonus: fixed a portability bug the test run exposed — `transcode_restart_guard_test.go` hardcoded `/bin/true` (absent on macOS); it now resolves `true` via `exec.LookPath`.

The overlap with #174 remains a sequencing decision for the maintainer — no code change belongs in this PR for it.

PR #281, opened 2026-07-02
fix(web): exit the reader through history so back doesn't re-open it

### LINKED ISSUE #189: Pressing back on series page after exiting reader re-opens the reader
**Goal:** To have the "back" button on the series detail page correctly navigate to the previous view (e.g., Library or Home) after exiting the manga/ebook reader.

**Steps:**

Open a manga or ebook and enter the reading interface.

Click the back arrow (top left) within the reader to exit back to the series detail page.

Once on the series detail page, click the back arrow (top left) again.

**Expected:**
The application should navigate backward in the routing history, taking the user to the previous screen they were on before opening the series (like the main Manga library view or a search results page).

**Actual:**
The application navigates forward/reloads the previous state, throwing the user right back into the reader interface.

**What is slow/broken:**
The "back" action is incorrectly triggering a return to the reader.

**Scope:**
Affects navigation specifically after exiting the manga/ebook reader.

Version/branch:
e244&#8203;d4a8

Deployment:
Docker

https://github.com/user-attachments/assets/732f6404-875b-425b-80ee-0faa653202ad

fluxis:
## Problem

Exiting the manga/ebook reader to the series page, then pressing back on the series page, re-opened the reader. The reader's header back control was a plain `<Link>` push: history became `[…, series, reader, series]`, so back landed on the reader again — an inescapable loop unless the user force-navigated elsewhere.

## Approach

Plain left-clicks on the reader's back control now consume the reader's history entry (`navigate(-1)`) when in-app history exists, so history returns to wherever the reader was opened from. The `href` is preserved for modified clicks (open in new tab) and remains the fallback when the reader was opened as a direct deep link (no in-app history).

The history-availability predicate is extracted to `lib/backNavigation.ts` and shared with `PageBack`, which already implemented the same pattern — one definition instead of two copies.

## Testing

New regression test: history seeded as `[origin, reader]` with `backTo` pointing at a third URL — clicking Back must land on the origin entry, not a fresh push of `backTo` (fails on the old push behavior). Existing backTo-sanitization tests still pass (the href is unchanged). Full EbookReader + PageBack suites green; `pnpm run lint` + `format:check` clean. (Two unrelated test files fail identically on clean origin/main — pre-existing upstream breakage, verified by stash comparison.)

Fixes #189

## AI-use disclosure

Implemented with Claude Code (test-first); human-reviewed before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved back navigation in the ebook reader when browsing from existing history.
  * Readers without navigation history are now taken to the appropriate fallback page.
  * Next-chapter navigation now returns readers to the series page correctly when navigating back.
  * Preserved standard link behavior for modified clicks and direct navigation.
  * Added regression coverage for reader back navigation and chapter transitions.

### DIFF
--- a/web/src/components/PageBack.tsx
+++ b/web/src/components/PageBack.tsx
@@ -1,6 +1,8 @@
 import { ChevronLeft } from "lucide-react";
 import { type To, useNavigate } from "react-router";
 
+import { hasRouterHistory } from "@/lib/backNavigation";
+
 interface PageBackProps {
   label?: string;
   to?: To;
@@ -25,9 +27,7 @@ export default function PageBack({
     : "absolute top-4 left-2 sm:top-6";
 
   function goBack() {
-    const historyIndex = window.history.state?.idx;
-
-    if (preferHistory && typeof historyIndex === "number" && historyIndex > 0) {
+    if (preferHistory && hasRouterHistory()) {
       navigate(-1);
       return;
     }
--- a/web/src/lib/backNavigation.ts
+++ b/web/src/lib/backNavigation.ts
@@ -1,0 +1,8 @@
+// hasRouterHistory reports whether the current entry has in-app router
+// history behind it, i.e. navigate(-1) stays inside the app. React Router
+// stamps its entry index on window.history.state.idx; the first in-app entry
+// has idx 0.
+export function hasRouterHistory(): boolean {
+  const historyIndex = (window.history.state as { idx?: unknown } | null)?.idx;
+  return typeof historyIndex === "number" && historyIndex > 0;
+}
--- a/web/src/pages/EbookReader.test.tsx
+++ b/web/src/pages/EbookReader.test.tsx
@@ -387,9 +387,7 @@ describe("EbookReader", () => {
     const backTo = encodeURIComponent("/item/manga-series-1?libraryId=7");
     await act(async () => {
       root.render(
-        <MemoryRouter
-          initialEntries={[`/reader/ebook/ebook-1?libraryId=7&backTo=${backTo}`]}
-        >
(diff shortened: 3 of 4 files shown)


Quick104:
### Code review

Found 1 issue:

1. The new Back handler pops exactly one history entry (`navigate(-1)`), but the reader's "Next chapter" links are plain pushing `Link`s with no `replace`. After advancing through chapters (history: `[series, reader-ch1, reader-ch2, ...]`), clicking Back lands on the previous chapter's reader page instead of exiting the reader — the user must click Back once per chapter read, which contradicts the PR's goal of exiting the reader through history. Note `handleFileChange` already uses `navigate(..., { replace: true })` for the same reason; making the chapter-nav links `replace` navigations would keep the reader at a single history entry so the one-pop exit works. (This matches the unresolved Codex inline comment on this PR; verified independently.)

Back handler that pops a single entry:
https://github.com/Silo-Server/silo-server/blob/800873ce96c9d941eee1299fe9234cf8510ccc61/web/src/pages/EbookReader.tsx#L599-L622

Pushing next-chapter links (header and end-of-book CTA):
https://github.com/Silo-Server/silo-server/blob/800873ce96c9d941eee1299fe9234cf8510ccc61/web/src/pages/EbookReader.tsx#L637-L639
https://github.com/Silo-Server/silo-server/blob/800873ce96c9d941eee1299fe9234cf8510ccc61/web/src/pages/EbookReader.tsx#L1305-L1307

Contrast with the existing `replace: true` file switch:
https://github.com/Silo-Server/silo-server/blob/800873ce96c9d941eee1299fe9234cf8510ccc61/web/src/pages/EbookReader.tsx#L303-L305

PR #283, opened 2026-07-02
feat(catalog): Latest Episodes sort — order series by newest episode file

### LINKED ISSUE #202: [v1] 'Latest Episodes' sort/filter for sections and collections
## Summary

Problem: Recently Added in Series is Recently added Series, I'd like a Recently added Episodes option.
Why it matters: One of the things I personally do is sort the library by recently added episodes so I can see what came out. Yes, we have the calendar but its not the same as having a dedicated section or collection.
Proposed change: Add a filter for Latest Episodes.
Tradeoffs or constraints:,
Whether this feels core, plugin-related, or deployment-specific: Core

## User value / why v1

_Reporter's use case is described above; articulate the concrete user value during triage._

## Reply digest

Recently-Added for Series surfaces recently-added series, not episodes; requests a 'recently added episodes' / sort-by-last-episode-date option for sections and collections (7 replies refining the exact sort semantics).

## Attachments

_None._

## Source

Imported from Discord:
- https://discord.com/channels/1487838737248292864/1502054975000416257

fluxis:
## Problem

Every recently-added surface reflects when a SERIES was first added — linking a new episode file never bumps the series' `media_item_libraries.first_seen_at` (`ON CONFLICT DO NOTHING`), so a long-running show with a fresh episode sorts as stale. Users can't see which shows received new episodes (#202).

## Approach

**New denorm** `media_items.latest_episode_added_at` (migration + backfill + partial series index), mirroring the `last_air_date_at` precedent. Source of truth is `episode_libraries.first_seen_at`; all three insert paths (`UpdateEpisodeLink`, `BulkLinkEpisodesBySeries`, scanner folder restore) bump the parent series in the same SQL statement — monotonic via `GREATEST`, and only for genuinely new links (conflict no-ops return no rows). Like `last_air_date_at`, the value is global rather than access-scoped — same accepted trade-off for an O(1) sort key.

**Sort registration** in both frameworks: a `querySortDefs` entry (sections, smart collections, and `/v1/catalog` accept it automatically via `QuerySortFieldSet`) and a `buildOrderByPlan` case for the browse path. `NULLS LAST` so movies and episode-less series trail.

**Jellyfin compat**: `SortBy=DateLastContentAdded` — Jellyfin's standard sort behind the TV "Latest" shelf, which clients already send — now maps to the new sort instead of silently collapsing to series creation date. Existing Jellyfin/compat clients get correct behavior with **no client changes**. `DatePlayed` keeps its previous `created_at` mapping instead of piggybacking on the same case.

**Web picker**: "Latest Episode Added" (series scope), sibling of "Latest Episode Air Date".

## API contract

Additive-only per the v1 rules: a new accepted sort value; both frameworks already fall back gracefully on unknown values, so older servers/clients interop unchanged. The compat remap is a behavior change existing clients will feel — deliberately, since the previous mapping was wrong.

## Testing

- DB-backed test covering both link paths: first link stamps, newer link bumps, conflict re-link does not re-bump.
- Sort-plan tests for both frameworks (denorm column read, no joins/args, NULLS LAST).
- `mapSortBy` tests pinning the new mapping and the `DatePlayed` split.
- **Live-validated** on a production deployment (19,048/19,140 series backfilled in seconds): Jellyfin `/Items?SortBy=DateLastContentAdded` order matched the SQL denorm order exactly, tracking an in-flight library re-ingest in real time; series that received new episodes today surfaced above newly-added series.

## Follow-ups (out of scope)

- Optional built-in "Latest Episodes" home-row recipe (the sort is already usable in custom sections/collections).
- silo-android / silo-apple can expose the new sort in native pickers; they already benefit via the Jellyfin mapping.

Fixes #202

## AI-use disclosure

### DIFF
--- a/internal/catalog/browse.go
+++ b/internal/catalog/browse.go
@@ -1482,6 +1482,13 @@ func buildOrderByPlan(sort, order string, snapshot *time.Time, argIdx int, singl
 			direction,
 			nullsClause,
 		), nil
+	case "latest_episode_added":
+		// Latest Episodes (issue #202): denormalized newest-episode-file
+		// arrival; NULLS LAST so movies and episode-less series trail.
+		return fmt.Sprintf(
+			"ORDER BY mi.latest_episode_added_at %s NULLS LAST, mi.content_id ASC",
+			direction,
+		), nil
 	case "rating_imdb":
 		return fmt.Sprintf("ORDER BY mi.rating_imdb %s%s, mi.content_id ASC", direction, nullsClause), nil
 	case "rating_tmdb":
--- a/internal/catalog/episode_added_denorm.go
+++ b/internal/catalog/episode_added_denorm.go
@@ -1,0 +1,38 @@
+package catalog
+
+import (
+	"context"
+	"fmt"
+)
+
+// updateSeriesLatestEpisodeAddedSQL fully recomputes the denorm from
+// episode_libraries after membership changes; series with no memberships reset
+// to NULL.
+const updateSeriesLatestEpisodeAddedSQL = `
+	UPDATE media_items mi
+	SET latest_episode_added_at = sub.latest_added
+	FROM (
+		SELECT s.series_id, (
+			SELECT MAX(el.first_seen_at)
+			FROM episode_libraries el
+			JOIN episodes e ON e.content_id = el.episode_id
+			WHERE e.series_id = s.series_id
+		) AS latest_added
+		FROM unnest($1::text[]) AS s(series_id)
+	) sub
+	WHERE mi.content_id = sub.series_id
+	  AND mi.type = 'series'
+	  AND (mi.latest_episode_added_at IS DISTINCT FROM sub.latest_added)`
+
+// RecomputeSeriesLatestEpisodeAdded refreshes latest_episode_added_at for the
+// supplied series IDs after episode library membership changes.
+func RecomputeSeriesLatestEpisodeAdded(ctx context.Context, execer itemExecer, seriesIDs []string) error {
+	seriesIDs = compactNonEmptyStrings(seriesIDs)
+	if len(seriesIDs) == 0 {
+		return nil
+	}
+	if _, err := execer.Exec(ctx, updateSeriesLatestEpisodeAddedSQL, seriesIDs); err != nil {
+		return fmt.Errorf("recomputing latest episode added denorm: %w", err)
+	}
+	return nil
+}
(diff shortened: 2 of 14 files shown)


Quick104:
### Code review

Found 2 issues (both confirm the earlier Codex inline comments — I validated each against the code):

1. `latest_episode_added_at` is only ever bumped upward (`GREATEST`) but is never recomputed when `episode_libraries` rows are removed. The same `UpdateEpisodeLink` statement deletes the old episode's membership when a file is re-linked, so a series whose newest membership was just removed keeps a stale timestamp and continues sorting as recently updated. The sibling `last_air_date_at` denorm this PR cites as precedent uses a full `MAX()` recompute with `IS DISTINCT FROM` precisely so decreases are handled; the same pattern is needed here (and on other membership-removal paths).

https://github.com/Silo-Server/silo-server/blob/6acf3529a268780a1a1ef4c60bd6d2c109df4baf/internal/scanner/file_repo.go#L2940-L2967

2. `mapSortBy` is used for all jellycompat `/Items` requests, not just series browses, so `SortBy=DateLastContentAdded` with `IncludeItemTypes=Movie` now sorts by `mi.latest_episode_added_at`, which is NULL for every movie — movie libraries lose the previous `created_at` ordering and fall through to the tiebreaker. The remap needs to be scoped to series requests.

https://github.com/Silo-Server/silo-server/blob/6acf3529a268780a1a1ef4c60bd6d2c109df4baf/internal/jellycompat/query.go#L438-L444

I am pushing fixes for both to this branch.

Quick104:
Both issues addressed:

- ca72674d: all episode-membership removal paths (`UpdateEpisodeLink` re-link, path-prefix clears, folder reconciliation, metadata rebind) now run transactionally and finish with a shared full-MAX recompute (`catalog.RecomputeSeriesLatestEpisodeAdded`) that lowers or NULLs `latest_episode_added_at` from the `episode_libraries` source of truth, mirroring the `last_air_date_at` maintenance pattern. Splitting into sequential statements also fixes a latent no-op: the old same-statement `deleted` CTE in the path-prefix clear read the pre-update `media_files` snapshot, so its `NOT EXISTS` guard could never pass.
- 978b5e38: jellycompat `DateLastContentAdded` maps to `latest_episode_added` only when `IncludeItemTypes` is exactly `Series`; movie and untyped requests keep the previous `created_at` ordering.

DB-backed regression tests extended (re-link lowers/NULLs the old series, bumps the new one; reconciliation reset). Verified with `go build ./...`, `go vet`, and the scanner/catalog/jellycompat suites — the two `TestBeginWebOperation*` failures are pre-existing on the branch (fail identically without these changes; `/proc`-dependent). Diff written by GPT-5.5 (Codex) under review supervision; validated and committed by Claude.

Issue #265, opened 2026-07-02
Epic: clean, persistent monitoring & observability


CoffeeKnyte:
## Goal

Clean, persistent, operator-friendly monitoring of a Silo deployment. An operator should be able to answer "what happened, and when" after the fact — including across container recreations — and be able to focus on one subsystem without grepping the entire log firehose.

## Why

Today Silo logs only to stderr. Under Docker that means:

- The only record of a run lives in the daemon's `json-file` log and is **deleted on every container recreate** (redeploy, image bump, `compose up` after an edit). Post-incident investigation of anything before the last restart is impossible.
- There is **no persistent on-disk log** without hand-editing the container entrypoint.
- There is **no way to isolate one subsystem** (jellycompat / transcode / scan / api) — every investigation is a grep over everything.

`opslog` already persists structured records to Postgres for the admin log viewer, but that is a DB-backed, query-oriented surface — not a durable file an operator can `tail -f`, ship to a log collector, or retain independently of the database.

## Scope / deliverables

- [ ] **Persistent file logging** — opt-in `SILO_LOG_FILE` sink with rotation, plus optional per-subsystem split files (`SILO_LOG_SPLIT`). Replaces the tee/FIFO entrypoint workaround. _(PR #264)_
- [ ] **Deployment wiring** — set the env vars in the deployment `docker-compose.yml`, remove the tee/FIFO entrypoint hack, document the resulting log paths.
- [ ] **Broaden per-subsystem tagging** — ensure jellycompat / scanner / metadata / transcode roots carry a `component` attribute so split routing is exhaustive rather than best-effort.
- [ ] **Quiet vs. durable capture** — decide whether `log_quiet` should also suppress file records (current) or whether files should capture regardless (tee before the quiet filter).
- [ ] **Secret redaction for file/console sinks** — `opslog` redacts before DB persistence; console/file output is currently raw. Decide whether to extend redaction to these sinks.
- [ ] _(future)_ Structured log shipping / metrics endpoint for external monitoring stacks.

## Non-goals (for now)

- Replacing the DB-backed `opslog` admin viewer.
- A full metrics/tracing stack — this epic is about logs first.

---

Capability-level tracking issue. Proposed for v1 scope lock; awaiting triage.

darthShadow:
[...]

PR #290, opened 2026-07-03
feat(observability): OpenTelemetry logs+traces with secret redaction and slog standardization

CoffeeKnyte:
Part of #265.

Supersedes #264. Opened in response to @darthShadow's feedback (https://github.com/Silo-Server/silo-server/issues/265#issuecomment-4865162461) to adopt OpenTelemetry rather than a custom file sink.

Two commits: the OpenTelemetry logs+traces foundation with secret redaction, and the repo-wide `slog` call-site standardization plus a lint gate that keeps it from regressing.

## Problem

**Logs disappear when the container is recreated.** Silo only ever wrote logs to stderr, so under Docker the only record lives in the daemon's `json-file` log, which is wiped on every redeploy or image bump. An operator investigating something that happened "before the last restart" has nothing to look at.

**There was no standard, tool-agnostic way to ship logs and traces off the box.** The earlier attempt (#264) solved durability with a bespoke rotating file sink. On review the maintainer asked us to build observability on OpenTelemetry instead of custom sinks, so a collector/backend of the operator's choice can own retention and correlation. This PR follows that direction.

**Secrets could reach the logs unmasked.** Redaction only ran on the operational-log database path. Anything written to stderr — and, once telemetry exists, to a collector — emitted API keys, tokens, and cookies verbatim.

**Log lines were inconsistent and uncorrelated.** Most call sites used the non-context `slog` form, so a log line could not be tied to the request/trace it belongs to, and nothing stopped new code from staying that way.

## Solution

Design and rationale are captured in `docs/architecture/observability.md` and the plan under `docs/superpowers/plans/`.

### feat(observability): OpenTelemetry logs+traces with secret redaction

Opt-in, gated entirely on `SILO_OTEL_ENABLED` / `OTEL_EXPORTER_OTLP_ENDPOINT`. With no config the behaviour is byte-for-byte the same as today (stderr + `opslog`).

- **New `internal/telemetry` package.** `Setup` builds one shared resource, a `TracerProvider` (parent-based trace-id-ratio sampler), a `LoggerProvider`, and the W3C `TraceContext`+`Baggage` propagator from `OTEL_*` env. It deliberately installs **no `MeterProvider`** — metrics stay on Prometheus, and leaving the global meter as the built-in no-op stops the trace instrumentation libraries from double-emitting (`internal/telemetry/telemetry.go`).
- **Logs are bridged, not rerouted.** An `otelslog` handler is added to the existing chain via fan-out (`slog.MultiHandler`), so records still reach stderr and the `opslog` DB/admin pipeline unchanged. The OTel branch is **level-gated** by the shared `LevelVar` (otherwise `MultiHandler.Enabled` would OR the branches and export Debug at `log_level=info`) and **best-effort** (a failing collector can't break the console/DB branches) — `cmd/silo/main.go` `buildBaseHandler`, `internal/telemetry/loghandler.go`.
- **Secret redaction on all sinks (`internal/logredact`).** A `slog.Handler` masks secret-keyed attributes (`password`, `token`, `api_key`, `authorization`, `cookie`, …) — including `.With`-bound attrs, nested groups, secret-keyed group subtrees, and values behind a `LogValuer` (resolved before inspection) — with a zero-alloc fast path when a record has no secret keys. `opslog.shouldRedact` now delegates to `logredact.SecretKey`, so the DB path and the console/OTLP sinks share one marker list and can't drift.
- **Rotation is infra-managed** (no custom in-process file sink, per the maintainer's steer). Retention is owned per sink: the container runtime for stderr (documented Docker `json-file` `max-size`/`max-file` + mounted volume), the collector/backend for OTLP, and `opslog`'s existing daily-partition pruning for the DB path.

### refactor(observability): slog context+component sweep, sloglint gate (phase 3)

- **Call-site sweep.** Converted the remaining `slog.<Level>(...)` calls to `slog.<Level>Context(ctx, …)` wherever a `context.Context` is in scope (background/init calls with no ctx are left as-is), across 183 files, via a type-aware AST codemod. Log levels and message strings are preserved verbatim; a `component` attr (canonical per-package name) is added to direct package-level calls. `main.go` and the telemetry package's own conversions ride with their file in the first commit so no file is split across commits.
- **Enforcement (`.golangci.yml`).** Enabled `sloglint` with `context=scope`, `static-msg`, `key-naming-case=snake`, `no-mixed-args`. After the sweep all four report zero violations repo-wide, so `make lint` / CI now blocks any regression to the non-context form. The gate ships with the sweep because it can't be green until the legacy sites are converted.

## Risk / follow-ups

- Redaction is key-based: a secret embedded in a free-text message or under an unrecognized key is not caught.
- `log_quiet` also suppresses OTel logs — the fan-out mirrors the console stream, so quieting a subsystem empties its OTLP stream too.
- Early-boot logs (DB connect, migrations, tuning) reach stderr only, matching existing `opslog` behaviour.
- Distributed tracing spans at the seams (HTTP middleware, pgx, redis, plugin gRPC, scanner/playback) are planned as follow-up phases; this PR wires the providers and propagation but does not yet start spans.
- Per-subsystem trace propagation into plugin subprocesses is owned by `silo-plugin-sdk` and is a coordination follow-up.
- Metrics intentionally remain on Prometheus; `/metrics` and Grafana are untouched.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` — all clean.
- `sloglint` (all four rules) — 0 violations repo-wide, tests included.
- `go test ./internal/telemetry/ ./internal/logredact/ -race` — pass (bootstrap enable/disable + no-op MeterProvider guard; redaction of record attrs, `.With`-bound, nested groups, secret group keys, `LogValuer`, `WithGroup`, clean pass-through, and `slogtest` conformance).
- Default-off path confirmed identical to today; log levels verified unchanged across the sweep (Info/Warn/Error/Debug counts identical pre/post).
- Each commit builds independently; every file is contained in a single commit.

## AI-use disclosure

Implemented with AI assistance (Claude Code): drafting, the AST codemod for the sweep, two rounds of adversarial review (which hardened the SDK bootstrap and found/fixed two secret-leak paths in the redaction handler), and this description. All changes were reviewed by the author.


CoffeeKnyte:
Addressed the two Codex findings in be5ad6e5:

- **Per-signal OTLP protocol** — `OTEL_EXPORTER_OTLP_{TRACES,LOGS}_PROTOCOL` now override the generic `OTEL_EXPORTER_OTLP_PROTOCOL` per signal, so mixed collector setups (e.g. HTTP logs + gRPC traces) build the right exporter (`telemetry/config.go`, `telemetry.go`).
- **Secret `WithGroup` names** — entering a group whose name is secret-bearing (e.g. `WithGroup("authorization")`) now masks every leaf in the subtree, matching `slog.Group("authorization", ...)` (`logredact/redact.go`).

build / vet / tests pass.

Quick104:
Pushed two commits addressing the review and resolving the conflict with main:

**ec2c73af — review feedback**
- Telemetry setup failure is now non-fatal: `Setup` returns usable no-op providers alongside the error, and `main` logs and continues with telemetry disabled (best-effort contract; a typo'd `OTEL_*` env can no longer crash-loop the server).
- `OTEL_TRACES_SAMPLER` is honored (`always_on`/`always_off`/`traceidratio`/`parentbased_*`); unsupported values fall back to `parentbased_traceidratio`.
- Node identity moved from the non-semconv `node.name` to `service.instance.id`.
- opslog retention-scope attrs renamed to `target_component`/`target_level` (they collided with the canonical `component` routing key); those lines now carry `component=opslog`.
- Nits: stale `levelGated` comment, `WarnContext` in the shutdown defer, LogValuer double-resolve documented.

**f14a949a — merge main**
- Conflict resolution keeps main's logic (audiobooks provider-error handling, autoscan transient-resolve stats) with the sweep's context+component form applied on top; go.mod/go.sum rebuilt from main plus the OTel requires via `go mod tidy`.
- Converted the 23 slog call sites that landed on main after the sweep, so the sloglint gate is green repo-wide again. Added `policy` to the canonical component list.

Verification: `go build ./...` clean, sloglint (uncapped) 0 issues, `go test` green across telemetry/logredact and all conflict-touched packages (the two `TestBeginWebOperation*` jellycompat failures are the known pre-existing flaky baseline).

PR #294, opened 2026-07-03
fix(metadata): stop specialist providers hijacking new library chains

fluxis:
## Problem

When creating a new **TV series** library, the provider chain defaulted to the **Sportarr** plugin at position 1 (enabled). More broadly, the chain-seeding logic had two flaws:

1. **Specialists could hijack the top slot.** `seedDefaultChain` enabled *every* provider whose declared `default_priority > 0` and ordered purely by that number. Sportarr declares `series/season/episode = 1`, which out-ranks TVDB (2) and TMDB (3), so it landed at position 1 enabled on every new series library.
2. **Single-purpose providers cluttered unrelated libraries.** Providers that declare only their own level (audiobook / ebook / manga metadata) were still attached as *disabled* rows to series and movie libraries.

The same happened to **existing** libraries whenever a plugin was installed, via `AppendProviderToAllChains`.

## Approach

- Add a `default_enabled` capability-metadata flag. **Defaults to `true`**, so every existing plugin behaves exactly as before. A provider sets it `false` to be seeded *installed-but-disabled* while keeping its declared priority — so a user can opt in per-library and it slots in where the manifest intends rather than jumping to the top.
- Seeding and append now **drop providers that don't declare a content level**, reusing the same `providerSupportsLevel` rule as the chain-less fallback (#106). Audiobook/ebook/manga no longer appear in series/movie chains.
- `LookupSeedPlacement` resolves support/priority/enabled in a single metadata fetch; `buildSeededChainEntries` is extracted as a pure, unit-tested helper.

Paired with a Sportarr manifest change (`default_enabled: false`, priority `50`) in `Silo-Server/silo-plugin-sportarr`.

## Frontend

Second commit standardizes the provider **slug casing** in the library chain editor: defaulted chains showed the display name (`TMDB`) while server-loaded chains showed the capability id (`tmdb`). Now the capability id is used everywhere (matches the API's `provider_slug` and the mono styling).

## Testing

- `go test ./internal/metadata/ ./internal/api/handlers/` — pass (new `TestExtractDefaultEnabled`, `TestBuildSeededChainEntries_*`).
- `pnpm vitest run` on `LibraryForm.test.tsx` — pass (new `metadataProvidersFromInstallations` tests).
- `pnpm run lint` (0 errors), `pnpm run format:check`, `tsc --noEmit` clean.

Behavior change affects **new** library creation and **new** plugin installs; existing seeded chains are unchanged (operator DB cleanup can follow separately).

## AI use

Implemented with Claude Code (Opus 4.8), test-driven. Reviewed by the author.

## Summary by CodeRabbit

* **New Features**
  * Library provider seeding now considers per-level support, default priority, and default enabled state, with deterministic ordering.
  * Admin library form now uses a standardized provider identifier when building provider entries.
* **Bug Fixes**
  * Providers that don’t support a selected content level are excluded.
  * Opted-out providers (`default_enabled:false`) are added but kept disabled; legacy providers are parked last and disabled.
  * Default enablement now reliably “fails open” when metadata is missing or malformed.
* **Tests**
  * Added unit tests covering default enabled/priority seeding and admin form provider mapping.

### DIFF
--- a/internal/api/handlers/libraries.go
+++ b/internal/api/handlers/libraries.go
@@ -1988,6 +1988,44 @@ func (h *LibraryHandler) HandleGetLibraryProviders(w http.ResponseWriter, r *htt
 	writeJSON(w, http.StatusOK, map[string]any{"levels": levels})
 }
 
+// HandleGetLibraryProviderDefaults handles GET /libraries/provider-defaults.
+// It returns the provider chain that would be seeded for a new library of the
+// given type, grouped by content level and in seeded order. The admin UI
+// renders this while creating a library instead of re-deriving the chain from
+// plugin manifests client-side, so the displayed defaults and the chain the
+// server seeds on create can never disagree.
+func (h *LibraryHandler) HandleGetLibraryProviderDefaults(w http.ResponseWriter, r *http.Request) {
+	libraryType := r.URL.Query().Get("library_type")
+	levels := metadataContentLevelsForLibraryType(libraryType)
+	if len(levels) == 0 {
+		// A type the server doesn't seed chains for (unknown, or one like
+		// podcasts with no metadata content levels) simply has no defaults.
+		writeJSON(w, http.StatusOK, map[string]any{"levels": map[string][]chainLevelEntry{}})
+		return
+	}
+
+	if h.ChainRepo == nil {
+		writeError(w, http.StatusServiceUnavailable, "unavailable", "Provider chain management is not configured")
+		return
+	}
+
+	out := make(map[string][]chainLevelEntry, len(levels))
+	for _, level := range levels {
+		out[level] = []chainLevelEntry{}
+	}
+	for _, e := range h.seedDefaultChain(r.Context(), libraryType) {
+		out[e.ContentLevel] = append(out[e.ContentLevel], chainLevelEntry{
+			PluginInstallationID: e.PluginInstallationID,
+			CapabilityID:         e.CapabilityID,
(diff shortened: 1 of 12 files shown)


Quick104:
Pushed `15756f08` extending the fix to the web UI create path: the form's client-built default chain (`buildDefaultLevelChains`) replaces the server-seeded chain whenever the chain is dirty — and changing the library type on create (the normal path for a series library) marks it dirty. It now mirrors `buildSeededChainEntries`: providers that don't declare the level are dropped, `default_enabled: false` providers are seeded disabled at their declared priority, and legacy catch-alls are parked last, disabled. Covered by new `buildDefaultLevelChains` / `default_enabled`-parsing vitest cases mirroring the server's `TestBuildSeededChainEntries_*`.

Quick104:
Follow-up push `87832ddc`: replaced the client-side mirror from `15756f08` with the server-driven approach. New additive endpoint `GET /api/v1/libraries/provider-defaults?library_type=X` returns exactly what `seedDefaultChain` would write; the create form renders that instead of re-deriving defaults from plugin manifests. Changing the library type now just refetches defaults (no longer dirties the chain), so an untouched create lets the server-seeded chain stand — no chain PUT at all. `buildDefaultLevelChains` / `metadataProvidersFromInstallations` and all frontend `default_priority`/`default_enabled` parsing are gone; one source of truth for ordering and enablement. Types without metadata levels (podcasts) get an empty `levels` map, matching the fact the server never seeds or resolves chains for them.

PR #314, opened 2026-07-06
fix(autoscan): advance marker past paths outside any library folder

uffcd:
A filesystem-watcher scan source (e.g. the CephFS plugin) observes the entire volume, so it legitimately returns changes for paths that live in folders which are not registered as Silo libraries (unregistered decade buckets, 4k-dv, anime-dub, download/recycle dirs, etc.).

Previously, when a poll returned paths but none resolved to a library folder, PollOnce recorded an "unresolved" error and refused to advance the marker, on the assumption this only happens for a freshly-enabled source with unconfigured path rewrites. For a whole-volume watcher that assumption is wrong: a single change under an unregistered folder pins the journal marker in place, so every subsequent poll re-reads the same window and autoscan stalls permanently. Real imports pile up behind the stuck marker and are never scanned until the next full library scan.

Treat "paths returned but none resolved" as a benign, expected condition: log a warning (so an operator can still investigate a genuinely misconfigured source) and advance the marker so the queue keeps moving. Partial resolution was already handled correctly — the unresolved subset is simply outside Silo's libraries.

Observed in production: one file added under movies/4k-dv (not a library path) froze the CephFS source for ~18 days; ~42k journal changes had accumulated behind it. With this change the marker advances and the backlog drains on the next poll.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined autoscan polling marker behavior and event outcomes when provider-reported changes don’t match any library folder.
  * Autoscan now advances with an unresolved warning for non-library-related misses, while holding the marker and reporting an error when failures appear transient/internal, reducing unintended stalls and improving status accuracy.
* **Tests**
  * Expanded autoscan poll tests to cover marker-advance vs marker-hold for transient/internal resolution failures and updated existing expectations accordingly.

### DIFF
--- a/internal/autoscan/repository.go
+++ b/internal/autoscan/repository.go
@@ -458,7 +458,9 @@ func (r *Repository) DeleteSource(ctx context.Context, id string) error {
 }
 
 // AdvanceMarker stores the opaque next marker for a source, stamps last_run_at,
-// and clears any prior error. Called only after a successful enqueue.
+// and clears any prior error. Called once a poll window's work is consumed —
+// after a successful enqueue, or when the window's paths all resolved outside
+// Silo's libraries and were advanced past.
 func (r *Repository) AdvanceMarker(ctx context.Context, sourceID, marker string) error {
 	tag, err := r.pool.Exec(ctx, `
 		UPDATE autoscan_sources
--- a/internal/autoscan/service.go
+++ b/internal/autoscan/service.go
@@ -50,6 +50,10 @@ type resolveStats struct {
 	ChangesResolved int
 	TargetsClaimed  int
 	Suppressed      int
+	// TransientErrors counts resolve attempts that failed with an internal
+	// (non-RequestError) error — resolver/database faults that may clear on a
+	// later poll, as opposed to paths that are simply outside Silo's libraries.
+	TransientErrors int
 }
 
 // connectionResolver resolves a stored connection to concrete credentials.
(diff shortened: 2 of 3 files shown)


Quick104:
Maintainer follow-up pushed in 730b8805 after review. The stall diagnosis and the direction here are right — thanks for the detailed writeup. Two things the flat "always advance" flip gave up, now restored:

**1. Transient resolver failures no longer advance the marker.** `resolvedAny == false` conflated two very different states: "every path is outside Silo's libraries" (`scantrigger.RequestError` — advancing is correct, this is the CephFS whole-volume case) and "resolve attempts failed internally" (DB timeout / resolver fault — advancing silently skips real imports until the next full library scan). `resolveAndClaim` already distinguished these internally, so it now counts internal failures (`resolveStats.TransientErrors`), and `PollOnce` holds the marker + records the error only in that case. The window is retried once the fault clears.

**2. The all-outside case stays observable.** It still advances (the fix this PR is for), but the poll event now finishes as `unresolved` with an explaining message instead of `success`. That keeps the condition visible in poll history and keeps the existing `unresolved` status filter in the API and admin Activity panel meaningful — after the original change nothing wrote that status anymore. No source-level `last_error` is recorded, so the source itself still reads healthy.

Also refreshed the stale doc comments on `PollOnce` and `AdvanceMarker` ("Called only after a successful enqueue") that described the pre-PR contract, and added a regression test (`TestPollOnceHoldsMarkerOnTransientResolveFailure`) alongside the updated advance test.

AI-use disclosure: review and follow-up commit were done with Claude Code (Claude Fable 5), maintainer-verified; `go test ./internal/autoscan/`, `go vet`, and `gofmt` are clean.

PR #317, opened 2026-07-06
fix(audiobooks): return all authors/series for non-paginated ABS requests

Quick104:
## Problem

The ABS Authors page in client apps only shows the first ~50 authors and won't scroll further. Clients fetch `GET /api/libraries/{id}/authors` with no `limit`/`page` (real ABS returns the complete list for that form and clients scroll it locally), but Silo applied a default limit of 50 (25 for `/series`) to the non-paginated form — a bare `{authors:[...]}` truncated to the first page, with no `total` field to signal more exist.

## Fix

Non-paginated `/authors` and `/series` requests now use `limit=0` ("return all" — the contract the store layer already implements). Explicitly paginated requests are unchanged, as are the capped filter-data/search author buckets.

Response-size note: this returns the full author set (~100k objects on a large library) exactly as real ABS does; the MV-backed query serves it in one indexed read.

## Verification

- `go build` / `go test ./internal/audiobooks/...` pass.
- Reproduced on a dev deployment via debug request logging: the reporting client sends the non-paginated form; details in internal notes.

## Disclosure

Implemented with Claude Code (AI-assisted).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-paginated library author and series endpoints to avoid unintended default limits, while enforcing maximum caps for very large result sets.
  * Added warnings for non-paginated requests when results are truncated.
  * Fixed author image handling by using the correct photo field for poster display.
* **Security / Logging**
  * Updated access logging to redact credential-related query parameters and to record sanitized query details and User-Agent when available.

### DIFF
--- a/internal/audiobooks/abs/access_log.go
+++ b/internal/audiobooks/abs/access_log.go
@@ -14,7 +14,26 @@ import (
 // per request. The 2xx/3xx path logs at Debug so a default-Info runtime
 // stays quiet during normal playback; non-2xx escalates to Warn so
 // failures still surface without an explicit log-level flip. Path is
-// captured query-less so ?token= and refresh tokens never land in logs.
+// captured query-less; the query is logged separately with
+// credential-bearing params redacted (see sanitizedQuery) so ?token=
+// and refresh tokens never land in logs.
+// sanitizedQuery renders the request query string with credential-bearing
+// params redacted, so pagination/sort/filter params are visible in debug
+// logs without ever landing tokens there.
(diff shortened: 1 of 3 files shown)


Quick104:
Follow-up pushed: verified on a large dev library (250k audiobooks, ~98k authors) that true return-all (15.5MB) crashes a non-paginating client, so the return-all form is now clamped to 10k authors / 2k series with a Warn log on truncation. Paginated clients still reach the full set via the paged envelope.

Issue #318, opened 2026-07-06
Wrongly merged "versions" cannot be split, and tagged folders with different provider IDs merge into one item


Quick104:
## Summary

Two related defects in version grouping, validated against a user report (internal Discord thread, reporting user asked "Is there a way to split apart a movie/show that was combined and matched incorrectly?"):

1. **Tagged folders with *different* provider IDs can merge into one item.** Reproduced with `Passenger (2026) [imdb-tt33763941][tmdb-1368314]` and `The Passenger (2026) [imdb-tt32298956][tmdb-1285959]`: `makeContentGroupKey` is title+year only, title normalization strips the leading article, so both folders derive `v1|movie|passenger|2026` and stack as four "versions" of one item — with state `resolved`, not even flagged ambiguous. The explicit tags are parsed but never enter the group key.
2. **There is no in-app repair for a wrong merge.** Item-level rematch re-identifies *all* files; root overrides refuse ambiguous roots with a 409 that points at a "split" operation that does not exist. The only workaround is renaming folders on disk — and watch state accrued under the merged item is never reattributed. Relatedly, item merges (`rebindItemToExistingItem`) orphan all per-user watch state today.

## Expected behavior

- Files carrying explicit provider tags never merge across different IDs, and always group across folders with the same ID.
- An admin can split a subset of an item's files to the correct (possibly new) item, durably across rescans, with watch progress/history following the files where evidence allows.
- Merging duplicate items preserves user watch state.

## Scope

Owning project: Silo Server (scanner group inference, catalog, admin API, web admin).

Design spec: `docs/superpowers/specs/2026-07-06-split-versions-reassign-design.md`.

## Acceptance criteria

- Structured provider tags anchor content-group keys (same tag groups, different tags never merge; untagged libraries keep existing keys).
- `POST /admin/items/{id}/split` with dry-run preview, path-scoped identity overrides, and evidence-based watch-state reattribution; `POST /admin/items/{id}/merge`.
- Admin UI: Split Versions dialog on item pages; Resolve path from the ambiguous-roots diagnostics.

PR #319, opened 2026-07-06
feat(matching): split wrongly merged versions with watch-state reattribution; anchor group keys on provider tags

Quick104:
Part of #318. Design spec: `docs/superpowers/specs/2026-07-06-split-versions-reassign-design.md` (includes an addendum for the field-validated root cause).

## Problem

Version grouping buckets files by a normalized title+year key. A user report (internal Discord thread) showed the worst case: `Passenger (2026) [tmdb-1368314]` and `The Passenger (2026) [tmdb-1285959]` — two different films, each in a correctly tagged folder — merged into one item as four "versions" with state `resolved`, because title normalization strips the leading article and the explicit provider tags never enter the group key. There was no in-app repair: item rematch moves *all* files, root overrides 409 on ambiguous roots pointing at a "split" flow that didn't exist, and the only workaround (rename + rescan) strands the watch state accrued under the merged item. Separately, item merges (`rebindItemToExistingItem`) have always orphaned per-user watch state.

## What changed

**Prevention — anchored group keys** (`internal/naming`): when a file/folder carries structured provider tags, the content-group key is derived from the ID anchor (`v1|movie|anchor|tmdb-…`, precedence mirroring `internal/contentid`) instead of title+year. Same tag always groups (cross-folder 1080p/4K version stacking keeps working — covered by test); different tags can never merge. Untagged files keep title+year keys, so untagged libraries see zero key churn. Tagged libraries re-key on next scan; content ids are deterministic from the same tags, so this is scanner-internal state churn only.

**Durable repair — path-scoped identity overrides** (`media_identity_overrides` migration, `internal/scanner`): operator-forced identities bound to a root folder or single file, applied per file during group inference before bucketing. File scope beats root scope; forced provider IDs anchor the key exactly like natural tags. Folders intentionally split across groups by overrides are no longer flagged `ambiguous` forever.

**Watch-state reattribution** (`internal/catalog/reattribute`): shared engine for splits and merges. Rows with a file reference (progress via `last_file_id`, downloads, per-session playback log) move exactly. `user_watch_history` has no file link, so rows are classified per (user, profile) against the `playback_history_admin` session log — moved only when the evidence is unanimous, with `evidence`/`keep`/`move_all` modes; ambiguous rows are surfaced, never guessed. Progress PK conflicts resolve by newest `updated_at`. Intent rows (favorites, ratings, watchlist, collections) stay put unless the whole item moves. Wired into `rebindItemToExistingItem`, fixing the existing merge-orphaning bug, including old→new episode id mapping by season/episode number for series merges.

**Endpoints**: `POST /admin/items/{id}/split` — resolves the target via deterministic content ids (existing item / new skeleton / `local-` unmatched), re-points files (episode links re-derive deterministically), persists overrides, reattributes state, all in one transaction. **Dry-run executes the full transaction and rolls back**, so preview counts are exact. Post-commit: identify the target, refresh the source, `ScanSubtree` the affected roots. Also `POST /admin/items/{id}/merge` and `GET /admin/items/{id}/files`. Admin-gated, additive-only; the `ambiguous_root` 409s now point at the split flow.

**Web admin**: Split Versions dialog (files grouped by folder → match-candidate search → dry-run preview → confirm) on movie/series pages; the ambiguous-roots diagnostics table gains a Resolve link via the new `content_id` on the roots listing.

## Risks / notes for review

- The Passenger case is reproduced as a regression test (`TestInferGroupAssignments_StructuredTagsAnchorGroupKeys`); pre-fix it produced one `resolved` group.
- Items merged **before** this fix don't auto-heal on rescan (files already point at the merged id) — the split flow is the repair path for existing damage.
- Reattribute DB tests are skip-gated on `SILO_TEST_DATABASE_URL` (no CI DB suite); they should be run against a scratch database before merge.
- History attribution quality is bounded by `playback_history_admin` retention; deferred follow-up is a nullable `user_watch_history.media_file_id` written at playback-stop for exact attribution.
- Existing `media_group_overrides` (whole-group forcing) is left in place; migrating it into the new table is deliberately out of scope.

## Verification

`go build ./...`; scanner/naming/metadata/handlers suites green (new tests: key-collision splits, scope precedence, rescan convergence, cross-folder same-tag grouping, untagged key stability, target resolution, episode-pair derivation); `tsc`, ESLint, Prettier, affected vitest suites green; `golangci-lint --new-from-rev=origin/main` clean apart from known local goconst test-file false positives; goose migration validated.

## AI-use disclosure

Implemented with Claude Code (Claude Fable 5), including spec authoring, implementation, tests, and the Discord report triage/reproduction; human-directed throughout.

## Summary by CodeRabbit

* **New Features**
  * Added an admin “Split Versions…” workflow on movie/series pages to move selected files into a resolved target (or detach as unmatched).
  * Added split/merge repair endpoints, plus a dry-run preview with reattribution reporting.
  * Library roots now optionally show the matched catalog item and provide clearer resolution links for ambiguous roots.
* **Bug Fixes**
  * Improved watch progress/history/downloads and favorites handling during split/merge, including configurable history modes and better evidence-based attribution.
  * Strengthened scan-time identity override behavior to keep grouping stable across rescans and avoid incorrect merges.
* **Tests**
  * Added unit and integration coverage for deterministic target resolution and reattribution behavior.

### DIFF
--- a/docs/superpowers/specs/2026-07-06-split-versions-reassign-design.md
+++ b/docs/superpowers/specs/2026-07-06-split-versions-reassign-design.md
@@ -1,0 +1,355 @@
+# Split Versions / Per-Folder Reassign with History Reattribution — Design Spec
+
+**Date:** 2026-07-06
+**Status:** Proposed
+**Scope:** scanner / catalog / metadata / admin API / web admin
+**Related:** `docs/architecture/deterministic-content-id.md`,
+`migrations/sql/078_content_groups_and_locations.sql`,
+`migrations/sql/20260614120000_content_id_online_reid.sql`
+
+Commands and paths assume the repository root is the cwd.
+
+## Problem
+
+The scanner groups files into logical items by `ContentGroupKey`
+(normalized title + year, `internal/naming/group_identity.go`) unless a
+structured provider tag (`{tmdb-…}`, `[tvdbid-…]`, `tt…`) anchors the
+identity. When two different titles parse to the same key (remakes,
+sequels with sloppy names, "Movie (2019)" vs "Movie (2019) Directors Cut"
+folders that are actually different films), their files are merged into a
+single item as "versions".
+
+Today there is **no way to fix a wrong merge in-app**:
+
+- Item-level rematch (`POST /admin/items/{id}/match/search|apply`,
+  `internal/api/handlers/admin_match.go`) re-identifies the *whole* item —
+  every file follows.
+- Group/root overrides (`internal/scanner/group_override_repo.go`,
+  `HandleUpsertRootOverride` in `internal/api/handlers/libraries.go`) force
+  an identity for an *entire* group, and the root-override endpoints
+  explicitly refuse ambiguous roots with 409 `ambiguous_root` — telling the
+  admin to "override the group after splitting", an operation that does not
+  exist.
+- The only real fix is renaming folders on disk with provider tags and
+  rescanning — and even then, watch state accrued under the merged item is
+  not reattributed.
+
+Symmetrically, when a **wrong split** is merged back (two items that are
+one title), watch state on the losing item is silently orphaned.
+
+## Goal
+
+An admin operation that:
+
+1. **Splits** a selected subset of an item's files (typically one folder)
+   into a different logical item — an existing item, a newly identified
+   one, or an unmatched `local-` placeholder.
+2. Makes the assignment **durable across rescans** (path-scoped identity
+   overrides, not just a one-time row edit).
+3. **Reattributes user watch state** (progress, history, downloads,
+   session records) to the item that actually owns the plays, using
+   per-file evidence where it exists.
+4. Provides the same reattribution when items are **merged**, so the
+   existing `rebindItemToExistingItem` path stops orphaning state.
+
+Non-goals: automatic detection of wrong merges (ambiguity flagging already
+exists and stays), cross-server state sync, file-level dedup.
+
+## Existing machinery this builds on
+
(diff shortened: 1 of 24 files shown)


Quick104:
Addressed all three review findings in c766c58 — each was verified by reproduction before fixing, this time against a real migrated PostgreSQL scratch database (the previously skip-gated reattribute DB suite now ran for real):

- **[P1] Evidence erasure**: confirmed — history classification now runs *before* the session log is re-pointed. The pre-fix code demonstrably fails `TestRun_HistoryEvidenceClassification` (`HistoryMoved:0`, unanimous profile misclassified ambiguous); post-fix it passes.
- **[P1] Managed downloads**: confirmed — `downloads.content_id` now moves per file on splits and per id pair on merges; `downloads.episode_id` is swept by the episode-pair remap. Covered by a new FK-backed seed in `TestRun_FileSubsetExactAndProgress`.
- **[P2] Series-scoped preferences**: confirmed — `user_audio_preferences`, `user_subtitle_preferences`, `user_series_playback_preferences` (all `series_id`-keyed) and the denormalized `user_home_item_dismissals.series_id` now move with merges/`move_all`, mirroring the provider-merge remap. Covered in `TestRun_WholeItemPairsAndConflicts`.

All five DB tests are green against PostgreSQL 🎉. One unrelated discovery while running the DB suites: `UpdateEpisodeLink` in `internal/scanner/file_repo.go` contains an invalid CTE (`inserted` referenced without a `RETURNING` clause) that errors on every call — pre-existing on main, tracked separately.

PR #336, opened 2026-07-07
feat(images): add larger TV image variants

raiden202:
# feat(images): add larger TV image variants

## Problem

Large TV detail surfaces can upscale the existing small still/logo variants, which is especially noticeable on 4K displays. Episode stills also reuse a smaller card thumbnail path where a TV-sized still variant is more appropriate.

## Approach

Add larger generated variants for TV-sized still and logo surfaces, use a still-specific episode image variant where needed, and preserve fallback behavior for cached images whose larger variants have not been generated yet. Fallback presigned URLs are cached only for the short missing-variant window, so a refresh-generated larger object is picked up promptly. Focused tests cover variant selection, generated variant ladders, and S3 resolver fallback from new keys to older cached variants.

## Verification

```sh
go test ./internal/api/handlers ./internal/catalog ./internal/imagecache ./internal/metadata
git diff --check
```

## Risks & follow-up

Moderate risk. The change is additive and keeps existing small card surfaces on smaller variants, but operators should expect extra S3/storage and cache-warm bandwidth once images refresh: logos add `w1280` alongside existing logo variants, and stills add `w780` alongside existing still variants. Existing cached media can continue to fall back when larger variants are unavailable, and missing larger variants are filled only when refresh/upload runs.

## AI-use disclosure

## Summary by CodeRabbit

* **Improvements**
  * Episode still images now use larger, dedicated variants (including consistent w780 scaling), while backdrops follow safer cached limits.
  * Episode artwork fallback now selects the most appropriate card-sized variant (poster/backdrop/card-thumbnail behavior remains correct).
  * Image resizing has updated variant ladders for logo, still, poster, and profile for better consistency.
  * When an image variant isn’t available, the system automatically falls back to the nearest supported lower-size version and refreshes sooner for missing variants.
  * Logos can now be delivered in higher-resolution variants.

### DIFF
--- a/internal/api/handlers/episode_image_variant_test.go
+++ b/internal/api/handlers/episode_image_variant_test.go
@@ -1,0 +1,33 @@
+package handlers
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/Silo-Server/silo-server/internal/models"
+)
+
+// The episode's own still generates w780/w500 variants, but the series-artwork
+// fallback (backdrop/poster) does not — applying w780 (or its w500 resolver
+// fallback) to a backdrop 404s. The card path must keep the fallback on w300.
+func TestEpisodeResponseShellImageVariant(t *testing.T) {
+	tests := []struct {
+		name       string
+		still      string
+		fallback   string
+		wantSuffix string
+	}{
+		{"own still uses w780", "tvdb/series/1/seasons/2/episodes/3/still/original.webp", "", "/still/w780.webp"},
+		{"backdrop fallback keeps w300", "", "tmdb/shows/1/backdrop/original.webp", "/backdrop/w300.webp"},
+		{"poster fallback keeps w300", "", "tmdb/shows/1/poster/original.webp", "/poster/w300.webp"},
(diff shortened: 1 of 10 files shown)


raiden202:
Addressed the fallback caching review in c2f220a8. Missing new-variant S3 fallbacks now cap their resolved-URL cache window to the same short missing-variant TTL, so refreshed w780/w1280 objects are picked up promptly instead of waiting on the full presign cache window. Added a regression test for the fallback URL cache TTL and reran the relevant Go package tests.

Quick104:
Thanks @raiden202 — this PR correctly identified both the missing large TV variants and the need for an S3 existence-check fallback while older libraries catch up, and that design is carried forward with credit in #742, which supersedes this PR. #742 extends the same idea to posters (the most visible blur on 4K TVs), adds a client-facing `image_size` request parameter with a capability endpoint, and an automatic one-shot backfill so existing libraries regenerate the new rungs without operator action. Closing in favor of #742; your groundwork here is appreciated.

PR #337, opened 2026-07-07
feat(scanner): capture audio stream profile metadata

raiden202:
# feat(scanner): capture audio stream profile metadata

## Problem

The scanner preserves video track profile metadata but discards ffprobe's audio stream `profile`. Downstream surfaces therefore cannot display or reason about audio profile details such as Dolby Digital Plus + Dolby Atmos when ffprobe exposes them.

## Approach

Carry the ffprobe audio stream `profile` through scanner probe data and store it in the existing audio track JSON metadata. This is an additive JSON field, with a focused scanner test that verifies profile capture.

## Verification

```sh
go test ./internal/scanner
git diff --check
```

## Risks & follow-up

Low risk. Existing fields and scanner decisions are unchanged. Consumers that do not know about `profile` can ignore it.

## AI-use disclosure

## Linked scope

Companion web PR that displays the captured field in Media Info: #339.

## Review follow-ups applied

- `catalogseed.AudioTrackRecord` now mirrors the audio `profile` field (parity with `VideoTrackRecord.Profile`), so seed exports do not drop it.

## Known limitations

- Existing libraries only gain `profile` once files are re-probed (new/changed files, probe repair, or a metadata refresh); there is no automatic backfill.
- jellycompat does not yet expose the audio profile on `MediaStream.Profile` / `AudioSpatialFormat`; tracked as a follow-up.

### DIFF
--- a/internal/catalogseed/service.go
+++ b/internal/catalogseed/service.go
@@ -2599,6 +2599,7 @@ func toAudioTrackRecords(tracks []models.AudioTrack) []AudioTrackRecord {
 			EmbeddedTitle: track.EmbeddedTitle,
 			Language:      track.Language,
 			Codec:         track.Codec,
+			Profile:       track.Profile,
 			Layout:        track.Layout,
 			Channels:      track.Channels,
 			Bitrate:       track.Bitrate,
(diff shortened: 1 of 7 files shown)


raiden202:
Resolved the stale Codex thread: current main includes `profile` on `AudioTrackRecord` and copies `track.Profile` in `toAudioTrackRecords`, so catalogseed export preserves the captured audio profile metadata.

PR #339, opened 2026-07-07
feat(web): show audio stream profile in Media Info

raiden202:
# feat(web): show audio stream profile in Media Info

## Problem

The Media Info dialog already surfaces video profile metadata, but it drops the equivalent audio stream profile even when the server provides one. That makes useful audio details, including Dolby Atmos-specific profile strings, invisible in the web UI.

## Approach

Add optional `profile` support to the web `VersionAudioTrack` type and render a `Profile` row in the Media Info dialog only when the value is present. Existing servers that do not send audio profiles keep the same output. Formatter test coverage was extended for the new optional row.

## Verification

```sh
pnpm exec vitest run src/pages/ItemDetail/components/mediaSpecSections.test.ts
pnpm exec eslint src/api/types.ts src/pages/ItemDetail/components/mediaSpecSections.ts src/pages/ItemDetail/components/mediaSpecSections.test.ts
pnpm exec prettier --check src/api/types.ts src/pages/ItemDetail/components/mediaSpecSections.ts src/pages/ItemDetail/components/mediaSpecSections.test.ts
git diff --check
```

## Risks & follow-up

Low risk. This is display-only and additive. No API contract is required from older servers because the field is optional.

## AI-use disclosure

### DIFF
--- a/web/src/api/types.ts
+++ b/web/src/api/types.ts
@@ -982,6 +982,7 @@ export interface VersionAudioTrack {
   embedded_title?: string;
   language?: string;
   codec?: string;
+  profile?: string;
   layout?: string;
   channels?: number;
   bitrate?: number;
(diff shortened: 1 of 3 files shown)


raiden202:
Resolved the stale Codex thread: the paired backend scanner/model PR has merged, and current main now exposes `profile` on audio tracks, so the Media Info Profile row has data to render when present.

Issue #376, opened 2026-07-11
Unify roaming and device-scoped user settings across web, Apple, and Android


Quick104:
# Unify roaming and device-scoped user settings across web, Apple, and Android

## Summary

User settings currently have good backend primitives—profile defaults, library overrides, per-series choices, and profile+device overrides—but the three clients do not use those scopes consistently. The result is that controls with the same visible meaning can roam in web, remain device-specific in native clients, silently fail to sync, or have no playback effect.

This was verified against:

- `silo-server` `3fd0912cb3fe15cc364f3dd04095c2e39db0bef0`
- `silo-apple` `120f493593119e71dfb1247dde0f89c55d46c1d0`
- `silo-android` `5c6439cebe753103c3a12cca7d1d152c5d6e35ab`

## Expected portability model

The product should have one documented precedence model:

1. Profile-wide defaults roam to every client and new browser session.
2. Optional library defaults override the profile for one library.
3. Explicit series/item track choices override both.
4. Device overrides apply only to one profile on one stable device/browser identity.
5. OS- or hardware-specific settings remain local-only.

An incognito browser intentionally receives a new device ID. Profile/library/series preferences should therefore carry over, while device overrides should fall back cleanly. We should not weaken private-browsing isolation by reusing the regular browser's device ID.

### P1: Apple `Audio Language` is saved but never consumed by playback

- `silo-apple/iosApp/iosApp/Screens/Settings/PlaybackSettingsView.swift:46-62` exposes the picker.
- `silo-apple/iosApp/iosApp/Screens/Settings/SettingsViewModel.swift:123-126` forwards it to `PlayerSettings.setAudioLanguage`.
- `silo-apple/iosApp/iosApp/Screens/Player/PlayerSettings.swift:372-375` writes `playback.audio_language` as a device setting.
- Repository-wide Apple usage shows `PlayerSettings.audioLanguage` is only loaded, cached, displayed, and written in the settings layer; no detail, session-start, route-planning, or player-selection path consumes it.
- Server initial audio selection instead reads `Profile.Language` in `internal/catalog/detail.go:2869-2907` and `internal/api/handlers/playback.go:1398-1428`.
- Apple decodes the full profile `language`, but `iosApp/iosApp/Networking/WireFormatModels.swift:36-49` deliberately drops it when converting to `UserProfile`; `UpdateProfileBody` also cannot write it.

[...]

PR #377, opened 2026-07-11
docs: define cross-platform user settings contract

Quick104:
## Summary

Design-only draft for standardizing user settings across the server, bundled web client, Apple clients, and Android clients.

This turns the findings in #376 into a typed, decision-complete contract covering ownership, value types, persistence scopes, resolution order, storage, API shape, client synchronization, UX language, migration, and release gates.

No runtime behavior changes in this PR.

## Core decisions for review

1. Every production user-facing setting requires a server-owned manifest entry, including settings whose values remain client-local.
2. Only private `local.<client>.*` implementation, diagnostics, or experimental knobs may bypass the server contract; they must never reach the API or normal production Settings UI.
3. **One coordinated breaking release.** The manifest, typed API, canonical storage, migration, and removal of the legacy settings surface ship together; server, bundled web, Apple, and Android update at the same time. Mixed-version operation is an explicit non-goal — no shim, no projection, no client fallback. **After the cutover, no future setting requires coordination**; manifest revisions move independently per repository.
4. `internal/policy` stays authoritative for restrictions. Definitions declare `constrained_by`, the effective response reports the permitted value alongside the user's stored preference, and a restriction filters what a preference *does* rather than validating what it *is*.
5. Account/profile values roam into a new browser or incognito session; profile-device overrides do not. Appearance and date/time settings live at **profile** scope, not account.
6. jellycompat (third-party Jellyfin clients) carries no contract negotiation; its DisplayPreferences seed repoints to the canonical resolver at profile scope and its blobs move off the legacy `user_settings` store.

## Changed in the latest revision

**The `426` upgrade gate is withdrawn** — but the coordinated cutover it was meant to enforce is not. Removing the legacy routes already produces the break: an old client gets `404` and its settings screens fail, which is what "update both" means. Adding a version header to every request plus a check in the authenticated middleware is strictly more code, in four repositories, for the same user-visible outcome — and it permanently couples every endpoint in the product to one subsystem's versioning for a one-time event. Route removal has no such tail.

**Preferences versus restrictions** is new. `internal/policy` already resolves `max_playback_quality` and metadata-language restrictions over the same controls this contract resolves preferences for. Without an explicit seam, a child profile capped at 720p gets an effective value of `2160p`, renders 4K as selectable, and playback silently disagrees. Adds `constrained_by`, `requested_value`/`permitted_values` reporting, and the rule that mutations exceeding a restriction are **stored, not rejected** — a stored 4K preference should take effect the day the cap lifts, not be destroyed by it.

**Compatibility rules split widening from narrowing.** Blanket definition immutability was simultaneously too strict (scope widening is backward-safe) and too weak (defaults and enum widening were already permitted, which broke the safety argument). `introduced_in` is now a manifest revision, and enum members, scopes, and widened bounds carry their own so a newer client never offers a choice an older server will reject.

**Appearance and date/time move from account to profile scope.** Account scope was an artifact of pre-profile `user_settings` storage; under the old rules a household shared one theme and text size, and any non-child profile could restyle everyone else permanently. Account writes now require the primary profile.

**A read path section**: batched context-resolution endpoint (`profile_series` resolution is per-item, so a season view would otherwise issue one request per episode), index requirements, session-snapshot rule, and a no-regression benchmark against the specialized tables being replaced.

Also: degraded settings mode instead of account lockout when a per-user SQLite database is quarantined; a separate `user_settings` event channel rather than overloading the admin-only one; an idempotency sweeper; advisory `platforms`; and clearer device-scope UX copy ("This device, for this profile").

Finally, a phase that **ships the #376 P1 fixes ahead of the contract** — the Apple audio-language no-op, Android's key/clamp/default drift, the durable outboxes, and web cache owner-tagging are all independent of the manifest and shouldn't wait on the migration.

## Review focus

- Is one coordinated switchover right, given it means no old client works until both sides update?
- Is `constrained_by` the right seam between this contract and `internal/policy`, particularly the rule that mutations exceeding a restriction are stored rather than rejected?
- Is moving appearance and date/time to profile scope correct, given it costs a fan-out migration rule now to avoid a new-key migration later?
- Is the widening-vs-narrowing compatibility table complete?
- Is retaining superseded columns for one release (rather than dropping them in the migration) worth the follow-up migration?
- Is the standardized scope copy clear enough for web, Apple, and Android UX?

## Verification

- Reconciled against the current server, Apple, and Android implementations documented in #376.
- Review pass (2026-07-11) re-verified every drift claim against silo-server, silo-apple, silo-android at the pinned baselines, plus the jellycompat surface.
- Revision pass (2026-07-25) re-checked the design against `internal/api/handlers/settings.go`, `web/src/lib/settingsManifest.ts`, `internal/policy`, `internal/events`, `internal/jellycompat/handlers_displayprefs.go`, `internal/userdb`, `internal/userstore/pgstore`, and `docs/architecture/v1-scope.md`. Confirmed: the unknown-key extension bag (`keyUsesUserScope` accepts any unregistered key), the web manifest's 18 definitions with zero user-scope entries against ~24 server keys, the `settings` event channel being admin-granted with no publishers, and v1 scope reading **NOT LOCKED** — which is why the removal needs no amendment if it lands before lock.
- `make verify-local-paths` passes.
- Spec-only change; no source tests were required.

Related to #376. This PR intentionally does not close the implementation issue.

## AI-use disclosure

The original design and its revisions were drafted with AI assistance (Claude Opus 5) under maintainer direction. All claims about current behavior were verified against the cited source files; the rollout model, policy seam, and scope assignments were reviewed and decided by the maintainer.

### DIFF
--- a/docs/superpowers/specs/2026-07-10-cross-platform-user-settings-contract-design.md
+++ b/docs/superpowers/specs/2026-07-10-cross-platform-user-settings-contract-design.md
@@ -1,0 +1,782 @@
+# Cross-platform user settings contract
+
+**Date:** 2026-07-10
+
+**Status:** Draft — coordinated breaking-release design for issue #376
+
+**Scope:** `silo-server`, `silo-apple`, `silo-android`, and the Silo web client
+
+**Tracking:** https://github.com/Silo-Server/silo-server/issues/376
+
+> Commands and paths in this document are repository-relative; assume the relevant repository root
+> is the cwd. Cross-repository references are prefixed with the repository name.
+
+## Decision
+
+The server repository owns the canonical contract for every **production, user-facing setting**.
+That is true even when the value is intentionally stored only on one client. A client PR must not
+invent a production setting key, type, default, range, or scope independently.
+
+There is one narrow exception: a client may add a private implementation, diagnostics, or
+experimental knob without a server PR when all of the following are true:
+
+1. Its key is in `local.<client>.<domain>.<name>` (for example,
+   `local.apple.player.decoder_logging`).
+2. It is not shown as a normal production setting.
+3. It is never sent to any Silo API.
+4. It is not expected to roam, survive reinstall, appear in admin UI, or have shared semantics with
+   another client.
+5. Promoting it to a production feature requires adding it to the shared contract first.
+
+This gives clients freedom for genuine local implementation details without allowing the public
+settings model to drift again.
+
+## User-visible behavior
+
+The contract makes persistence visible and predictable:
+
+| Setting scope | New browser/incognito session | Another signed-in client | Reinstall | Admin-visible |
+|---|---:|---:|---:|---:|
+| Account | Yes | Yes | Yes | Yes |
+| Profile | Yes | Yes | Yes | Yes |
+| Profile + device override | Profile default only | Profile default only | Profile default only unless the device identity is restored | Yes |
+| Profile-device only | No; a new browser is a new device | No | No unless the device identity is restored | Yes |
+| Client-local | No | No | No unless the client explicitly uses OS-backed backup | No |
+
+Therefore, signing into an incognito window must carry profile language, subtitle behavior, and any
+profile-level subtitle appearance. It must not copy ordinary-browser device overrides. The
+incognito window gets a new device identity and resolves those settings from the profile fallback.
+
+The UI must use these exact scope descriptions:
+
+- **All devices for this profile** — profile value that roams after sign-in.
+- **This device/browser** — override tied to the active profile and device identity.
+- **Only this app/device** — client-local value that is never uploaded.
+
+Avoid ambiguous labels such as “global,” “default,” or “remember this” without naming what the
+value follows.
+
+## Why this is needed
+
+The current implementation has three partial contracts:
+
+- `silo-server: internal/api/handlers/settings.go` owns validation, defaults, and a `user` versus
+  `device` registry, but unknown user keys are accepted and values are strings.
+- `silo-server: web/src/lib/settingsManifest.ts` independently owns labels, controls, defaults,
+  enum options, and numeric ranges.
+- Apple and Android independently own raw key constants, defaults, parsing, and local migration
+  behavior.
+
+That duplication has produced verified drift:
+
+- Apple writes `playback.audio_language`, but playback selection reads the profile language; the
+  device value currently has no effect.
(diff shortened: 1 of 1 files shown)


Quick104:
Applied the design-review findings in 9314df18 (review pass verified every claim against silo-server, silo-apple, and silo-android code plus the jellycompat surface — all drift claims held):

**Design changes**
- **Post-cutover gate softened to protocol version only.** The cutover release is still built and verified as one exact version+revision set, but afterwards manifest revisions are monotone-compatible within protocol 1: the server accepts any revision, and clients hide definitions newer than the connected server. Rationale: definition immutability already prevents split-brain for additive revisions, and exact-match would have made every new setting a four-platform lockstep release — recreating the pressure to smuggle knobs through `local.*`. Also documents the store-auto-update vs operator-upgrade window and requires both upgrade screens to name which side is behind.
- **New "Jellyfin compatibility surface" section.** jellycompat runs on its own router/listener and is exempt from the handshake (third-party clients can never be gated). Its DisplayPreferences handler was the one real breakage found: it seeds from profile `subtitle_language`/`subtitle_mode`/`auto_skip_credits` columns this migration removes and persists via the legacy `user_settings` store. The spec now moves those blobs to dedicated jellycompat storage and repoints the seed at the canonical resolver; the `user_settings` table is then dropped explicitly.

**Corrections & hardening**
- Foreign-key claim corrected: the per-user SQLite store declares no FKs and Postgres has none on library/series/device columns, so delete behavior is application-enforced and covered by the store conformance suite.
- Migration failure semantics: recognized keys with unnormalizable values are quarantined to the rejects table (reported, resolve to default) instead of bricking startup; hard abort is reserved for structural failures. Per-user SQLite databases migrate atomically and independently with completion markers — one damaged user DB cannot block the whole server.
- Account-scope authorization specified: child profiles cannot mutate account-scope values.
- Native outboxes treat 426 as a hold state (queue kept, upgrade surfaced), not a failure.
- Last-write-wins concurrency stated explicitly; manifest digest defined over RFC 8785 canonical JSON; settings-changed events ride the existing `internal/events` hub (its `settings` channel re-scoped from admin-only to per-user routing).
- Baseline note: the silo-apple commit sits on `feature/tvos-manual-up-next` (merge base `169e4917` with `main`); all cited files are identical on both, so findings hold on `main`.

PR body updated to match (core decision 3, review focus).

PR #380, opened 2026-07-11
fix(userstore): batch allowed-library and profile lookups when listing

JeremySNR:
Listing profiles was costing 1 + P queries: one allowed-libraries lookup fired per profile row while the cursor was still open, which on Postgres also checks out a second pooled connection mid-scan. The admin sessions dashboard makes it worse by calling ListProfiles once per streaming user just to resolve names. The SQLite store had the same pattern for both profiles and collections, even though the Postgres collections path was already written with array_agg to avoid exactly this.

Now the rows are collected first and the child lists come back in one batched query. No behaviour change, just fewer round trips.

Tested locally: profile and collection listings return the same data in the same order as before, now with one batched child query instead of one per row.

Built with AI assistance. I've read through the whole diff and tested it myself, and I'm happy to own it.

### DIFF
--- a/internal/userdb/collections.go
+++ b/internal/userdb/collections.go
@@ -3,7 +3,6 @@ package userdb
 import (
 	"database/sql"
 	"fmt"
-	"strings"
 
 	"github.com/Silo-Server/silo-server/internal/userstore"
 )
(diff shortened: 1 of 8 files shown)


JeremySNR:
@coderabbitai review

PR #391, opened 2026-07-13
perf(literaryworks): rewrite match-candidate OR into indexable UNION …

uffcd:
…branches

listMatchCandidateIDs combined its title, provider-id, and series predicates in one OR. Mixing a column predicate with EXISTS subqueries in an OR defeats index selection, so the planner walked every ebook/audiobook row (~250k on a large library) and applied the OR as a filter: ~220ms and ~209k buffer hits per call, measured live. The audiobook scanner calls this once per unlinked book via AutoLinkContent, turning a full library scan into ~15 core-hours of Postgres time.

Rewriting the OR as a UNION of independently indexable branches (title via idx_media_items_books_title_lower, provider ids via their unique index, series via the *_series_name_lower indexes) resolves the same candidates in ~0.07ms and 9 buffer hits. A source with no predicates keeps the old all-opposite-format fallback. Result set, ordering, ignored-decision exclusion, and LIMIT semantics are unchanged, covered by new DB-backed tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved matching candidate selection for literary works using title, series, and provider information.
  * Preserved exclusions for ignored matches, same-format items, source items, and unrelated content.
  * Added fallback matching to include opposite-format items when no matching details are available.
  * Maintained candidate ordering and result limits for consistent results.

### DIFF
--- a/internal/literaryworks/match_candidates_test.go
+++ b/internal/literaryworks/match_candidates_test.go
@@ -1,0 +1,148 @@
+package literaryworks
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+)
+
+// TestListMatchCandidatesBranches covers the UNION-branch candidate query:
+// title, provider-id, and series matches must each surface candidates through
+// their own indexable branch, ignored decisions must exclude, and same-format
+// items must never match.
+func TestListMatchCandidatesBranches(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


Quick104:
Closing this duplicate in favor of #1288, which addressed the same OR-to-UNION candidate query linked to issue #1262. #1288 has since merged, so the optimization is now in main.

Issue #333, opened 2026-07-07
playback: restart resilience misses control-plane endpoints — progress 404 kills recoverable streams within 10s


Quick104:
## Summary

A server restart still visibly interrupts playback on every native client (black screen + spinner), even when the client has a large forward cache and could keep playing untouched. The restart-resilient playback work in #174 already solves this for the **data plane** — `/stream/{session_id}` reconstructs a missing session from the `?st=` recipe token, and clients verifiably preserve that token. But the **control plane** was never wired in: session-scoped endpoints that don't carry the token still do a plain `GetSession` and return the `playback_session_not_found` 404.

The killer is `HandleUpdateProgress`. Clients POST progress every ~10 seconds and treat that exact 404 sentinel as "session dead → tear down and renew playback" (Apple: `PlayerViewModel.startProgressReporting` → `attemptStaleSessionRenewal`; Android: equivalent logic in `PlaybackSessionLifecycle.kt` / `PlaybackSessionManager.kt`). So after every restart the client kills an otherwise perfectly recoverable stream within ≤10s — usually before the stream path ever gets a chance to demonstrate that reconstruct works.

## Affected endpoints (plain `GetSession` → `playback_session_not_found`, no reconstruct)

| Endpoint | Location | Post-restart consequence |
|---|---|---|
| `HandleUpdateProgress` | `internal/api/handlers/playback.go` (~1761) | Clients declare session dead within one 10s heartbeat → visible playback teardown |
| `HandleChangeAudioTrack` | `internal/api/handlers/playback.go` (~1873) | Audio switch fails after restart |
| `HandleSubtitle` | `internal/api/handlers/stream.go` (~195) | Subtitle track fetches 404 after restart |
| Session websocket attach | `internal/api/handlers/session_ws.go` (~72) | Realtime (AI subtitles, admin commands) can't reattach |
| `HandleStartTranscode` | `internal/api/handlers/playback.go` (~2348) | Post-restart fallback/upgrade to transcode fails |

`HandleStopPlayback` 404ing is fine (stop of an unknown session is a no-op from the client's perspective).

## Why the existing token can't cover these

The `?st=` stream token rides only on the serve URL, and by explicit design (`LoadOrReconstructSession` doc comment) it is the *sole* reconstruction descriptor — "there is no shared per-session store to fall back on." Progress/audio/websocket requests never carry it. The durable Redis recipe store (`noderecipe`) exists only for the remote transcode-node path.

## Proposed implementation

[...]

PR #398, opened 2026-07-14
feat(playback): add protocol v3 planning and recovery

Quick104:
## Summary

- add the server-side Playback Protocol V3 capability, decision, session, replan, seek-reanchor, and observability contract
- persist deterministic plans, attempts, routes, and revision-aware replans for idempotent recovery across retries and rolling deploys
- centralize HDR/Dolby Vision, subtitle, remux/transcode, and device-quirk decisions in the server planner
- propagate required video bitstream filters through recipe cards, stream tokens, and transcode-node execution
- add focused protocol, route-planning, subtitle-policy, device-quirk, Dolby Vision, migration, and session tests
- harden P1 recovery paths with current network evidence, canonical route events, bounded track selection, remote-node capability/readiness checks, and atomic live-session replacement

## Why

The Android Media3 V3 player needs one authoritative server plan instead of duplicating container, codec, HDR, subtitle, and recovery heuristics in the client. This also gives retries and failure recovery stable attempt identities, explicit diagnostics, and compare-and-swap replan semantics.

## Client and rollout impact

- Paired client work: Silo-Server/silo-android#67 (merged).
- Existing clients remain on the current playback paths unless they advertise and call the V3 contract.
- Apple does not advertise V3 yet; adopting this contract there remains follow-up work.
- The new attempt-revision migration includes backfill and synchronization triggers for rolling-deploy and rollback compatibility.
- Dolby Vision base-layer extraction requires the `dovi_split` FFmpeg filter; the planner checks availability before selecting that route.

## Validation

- `go test ./internal/playback ./internal/api/handlers ./internal/streamtoken ./internal/transcodenode ./internal/proxy`
- `go test ./cmd/silo ./web`
- `make migrate-validate`
- `make verify-local-paths`
- `pnpm --dir web run lint` (0 errors; existing warnings only)
- `pnpm --dir web run format:check`
- `pnpm --dir web run build`
- Focused P1 regression tests pass across `internal/playback`, `internal/api/handlers`, `internal/nodepool`, `internal/transcodenode`, and `migrations`, including `-race` coverage for replacement-state and remote-node paths.
- `go test -p 1 ./...` was also exercised. It reached the existing `internal/jellycompat` process-lock failures and one timing-sensitive fake-FFmpeg NVENC probe under suite load; that playback probe passes when rerun alone.
- The paired Android V3 client was exercised against this server worktree across representative 1080p HEVC Main10, 4K HDR10, Dolby Vision profiles 5/7/8, EAC3/FLAC, ASS/SRT, HLS, direct-play, remux, transcode, seek, and recovery paths.

## Tracking

Part of #333 for the broader restart/recovery control-plane work.

## AI use

Implemented and reviewed with OpenAI Codex assistance. Tests and behavior were validated locally, including comparison of the unrelated full-suite failures against clean `main`.

## Summary by CodeRabbit

* **New Features**
  * Added Silo Playback Protocol v3 with capability discovery, durable idempotent playback start, replanning (seek reanchor + recovery), and bounded route diagnostics.
  * Added authenticated endpoints for v3 capability probing, session replanning, and route-event ingestion, with feature-flag rollout.
  * Enhanced Dolby Vision remux/transcode handling and improved playback transport identity persistence.
* **Bug Fixes**
  * Improved subtitle handling: VTT requests are enforced for text subtitles and unsupported bitmap subtitle VTT now returns an error.
  * Fixed streaming/remux routing and reconstruction to keep DV/remux behavior consistent.
* **Tests**
  * Expanded Protocol v3, subtitle, Dolby Vision/remux, and idempotency/replacement test coverage.
* **Documentation**
  * Published the end-to-end Protocol v3 server specification.


Quick104:
# Deep review — protocol v3 planning and recovery

**Verdict: request changes.** The architecture is genuinely strong — deterministic plan keys, digest-checked idempotency, prepared-transport commit/rollback, complete old-token back-compat — but the PR ships one critical feature-killing defect (a nonexistent FFmpeg filter name), several high-severity reliability bugs at the durable-vs-live-state seam, and a schema/code drift that silently guts the new telemetry. Line numbers refer to PR head `03fc9de8`. Follow-up commits addressing these findings are being pushed to this branch.

## Critical

**1. The Dolby Vision strip filter was renamed to one that doesn't exist — kills the headline DV7→HDR10 feature and regresses main's working path.**
`main` probes and uses `dovi_rpu=strip=1` (real FFmpeg 7.1+ bitstream filter). The PR renames every probe, argument, constant, and test to `dovi_split=mode=bl` (`internal/playback/remux.go:45,140,207`, `transcode.go:65`, `transformations_v3.go:38`) — verified against FFmpeg 8.0.1 and the Dockerfile's jellyfin-ffmpeg7: only `dovi_rpu` exists, and it has no `mode` option. The stale comment at `remux.go:55` still says `dovi_rpu`. On every real deployment the probe fails closed: `server_dv7_to_hdr10` is permanently unavailable, and the **previously working v2 DV strip remux is disabled** (`remux.go:207` errors "requires dovi_split"). Note `dovi_rpu=strip=1` (strip metadata) and "mode=bl" (extract base layer) are not even semantically equivalent — this needs a deliberate decision, not just a rename back.

**2. Route-event DB CHECK constraint rejects events the API accepts and the server itself emits.** *(fix incoming)*
`migrations/sql/20260712103000_add_playback_protocol_v3.sql:66` allows 6 event names; the handler allowlist (`playback_v3.go:1520-1524`) accepts 9, and every successful seek reanchor emits `runtime_correction_succeeded` (`playback_v3.go:895`). The async writer logs a warning and drops the row — so 100% of runtime-correction telemetry never lands, while clients get 202 Accepted. Additionally the Android client's `seek_reanchor_requested`/`seek_reanchored` events aren't in the handler allowlist at all and are rejected.

## High

**3. Client-triggerable panic: negative `selected_tracks.subtitle.index` on replan.** *(fix incoming)*
`ReplanRequestV3.Validate()` never bounds `SelectedTracks` (start requests are protected by `validateTrackPairV3`; replans are not). `applySelectedTracksToStartV3` copies the index verbatim and `remapSubtitleSelectionV3` (`playback_v3.go:1286-1290`) evaluates `index < len(...)` — true for −1 — then indexes the slice and panics, *before* `NormalizeAndValidate` runs. Any authenticated user whose attempt had a start-time edition fallback can 500 the recovery endpoint at will.

**4. Everything after the durable commit is fire-and-forget — durable plan and live session can permanently diverge.** *(fix incoming)*
Start: `SaveAttempt` → `transport.commit()` → warn-only `updateV3SessionState` (`playback_v3.go:275-277,478-480`). Replan: after `CompleteReplan`, the commit closure does `_ = SetEffectiveMediaFileID`, `_ = UpdateAudioTrack`, warn-only state/progress updates across three separate lock acquisitions (`:876-899`). If the session idle-expires during transport prep, the durable record points at the new plan while the live session serves the old media file; for a remote transport, `TranscodeNodeURL`/`TranscodeTransportID` are never set, so manifest requests spawn a duplicate local ffmpeg beside the running remote job, and stop paths can't kill the node's process.

**5. Replan pool self-starvation under exactly the recovery storm this feature exists for.**
The handler holds `AcquireSessionLock` — a dedicated pooled connection with an open transaction (`planstore/postgres.go:22-54`) — across `GetAttempt`, `BeginReplan`, file loads, a local ffmpeg spawn or a 10s remote-node HTTP call, a synchronous 5s `StopRemoteTranscode`, and `CompleteReplan`, each of which needs a *second* connection from the same pool (default `MaxConns=20`). A transcode-node death triggering ~20 concurrent replans makes every conn a lock-holder waiting on the pool — replans deadlock until client contexts cancel, and auth/catalog queries starve server-wide.

**6. Mis-claimed "native" DV Profile 7 plans (compounds #1).**
When the client advertises DV profile 7 and strip is unavailable, `plan_v3.go:169-204` emits a progressive-remux plan claiming `DynamicRange: "dolby_vision"`, `Claims.Video.DolbyVision: true, Reason: "native_profile_supported"` — while the remux drops the EL and leaves dangling RPUs (`remux.go:53-59,90-94`). This contradicts the planner's own guard ("a decoder profile claim alone is not proof of native dual-layer output") on the protocol whose entire premise is validated claims. Related: DV7 "preserve" mode never emits `-tag:v dvhe` (`remux.go:212-213`), so Media3 can't select the DV decoder even when preservation is intended.

## Medium

**Contract/gating inconsistencies**
- `deviceQuirkProtocolAvailableV3` (`device_quirks_v3.go:87-89`) requires the feature flag in **both** `ClientFeatures` **and** `ClientPlaybackContext.Features`, while `detailedVideoEligibleV3`/`audioEligibilityV3` accept **either**. A client advertising in one location gets detailed-decode treatment but silently loses all device quirks (e.g. the Fire TV DV8+HDR10+ sanitizer).
- Subtitle policy always evaluates renderability against `Engines["media3_direct"].Subtitles` (`subtitle_policy_v3.go:52`) regardless of which engine the plan selects — an HLS-transcode plan can carry `Mode: "render"` the executing engine can't honor.
- ID-only subtitle selections (`subtitle_track_id` without index) skip remap entirely on edition fallback (`playback_v3.go:1283` early-return), leaving a stale track ID pointed at the wrong file.
- `remapAudioIndexV3` runs on every intent-change replan with no same-file short-circuit and picks the first (codec, language, channels) match (`playback_v3.go:1269-1280`) — a user listening to an eng/ac3/5.1 commentary track gets silently switched to the main track on a quality change, and the wrong track is then durably pinned.

**Rollout/deploy risks** (the rolling-deploy claims cover only the DB)
- New-server + old-worker fleets: an old proxy routes by `SessionID` (no `tid`), so a new node 404s reconstructs; an old node spawns a duplicate ffmpeg and silently drops `video_bitstream_filter`, producing DV7 output with dangling EL/RPUs. No capability negotiation gates v3 remote routing on worker version (`proxy/server.go:194-202`, `transcodenode/server.go:420-426`).
- Unconditional `-tag:v dvhe` for DV5/8 now applies to *all* remux consumers including browsers (`remux.go:141-147`) — previously-playing `hev1` DV8 files may be rejected by non-DV demuxers. The `empty_moov`→`delay_moov` change (`remux.go:172`) likewise changes startup latency for every legacy remux client in the same release.
- Replan-orphaned stream tokens: after a replan deletes the predecessor transport, a still-valid old token re-presented by an in-flight HLS request resurrects a full end-of-file ffmpeg on the node, which has no idle reaper (`transcodenode/server.go:404-464`).

**State/liveness edges**
- `RemuxDVMode` can never be cleared once set (`session.go:691-693` only writes non-empty values): an edition-fallback replan from DV strip-remux to SDR leaves `strip_to_hdr10`, and every subsequent `/stream` remux 500s until the client exhausts replans into transcode.
- `CheckReplacementAllowed`'s "−1 for self" arithmetic (`session.go:572-579`) assumes the current session is in the active counts, but sessions idle >45s (exactly a failed session awaiting replan) are excluded — the decrement frees someone else's slot and lets a third transcode past `MaxTranscodes=2`.
- Idempotent start replay skips the session-liveness check the replan replay performs (`playback_v3.go:143-149` vs `:625-628`) — retries return 201 plans for dead sessions until the attempt row expires.
- An expired-but-uncleaned attempt row (up to 1h before hourly cleanup) makes `SaveAttempt` conflict, but the recovery lookup filters expired rows → every retry hard-fails `internal_error` and kills the fresh session (`planstore/postgres.go:65-107`, `playback_v3.go:264-273`).
- Transient store failures map to terminal client outcomes: any `GetAttempt` error → 404 "session not found", route-event ownership lookup error → 403 (`playback_v3.go:577-581`, `:1133-1145`) — a DB blip makes clients tear down playback instead of retrying.

**Planner policy gaps**
- Any quality reduction on HDR/DV sources (device `max_resolution` below source, bandwidth estimate under "auto") routes into transcode and terminals `hdr_transcode_unsupported` instead of degrading to a source-preserving route with a warning — 4K HDR + a 1080p-capped device is simply refused (`plan_v3.go:102-104,264-266,375-391`).
- Legacy rows with `HDR == true` but no per-track range data become `hdr_unknown` → permanently unplayable under v3 until re-probed (`capabilities_v3.go:66-72,145-147`).
- `StartRequestV3.Metered` is accepted but entirely dead; `bandwidth_cap_kbps` is honored only under "auto" quality; replans drop network evidence altogether *(replan fields: fix incoming)*.
- Failed-route exclusion trusts client-supplied `attempted_plan_keys` exclusively for ordinary failure recovery; `AttemptCount` is validated but never read, and no server-side replan budget exists — a client with unreproducible local-mutation keys can be re-served the same failed plan indefinitely (`plan_v3.go:449-452`).

## Low (compact)

- `playback_route_events` retention DELETE has no usable index — hourly full scan; 30-day retention hardcoded.
- `CompleteReplan`'s attempts UPDATE has no revision predicate — correctness rests entirely on the advisory lock; add `AND current_replan_request_id = $base` as defense-in-depth.
- All persistence tests run against the in-memory store, whose semantics diverge from Postgres (overwrite vs conflict, creates-missing vs not-found) — precisely why #2 survived. The Postgres store has zero tests despite `SILO_TEST_DATABASE_URL` precedent.
- Legacy `/playback/start` now enforces a 256KB body cap and pre-parses `protocol_version` — a mild existing-endpoint behavior change under the v1 additive-only rule.
- Route-event rate limiting runs after the ownership DB lookups (read-side unbounded); `sanitizeDiagnosticsV3` keeps an arbitrary 16 of 32 allowed keys via random map iteration; the diagnostics whitelist is missing the network keys Android sends *(fix incoming)*.
- Per-request uncached `server_settings` SELECTs for the v3/shadow flags on every start/replan/event; the revision-sync trigger double-updates the attempts row per replan on new binaries (WAL churn only).
- Remote `/transcode/start` timeout after the node actually started leaves an orphan job (no rollback until response parse); `.vtt`-suffix override sends PGS/DVD tracks into a guaranteed-failing bitmap→text ffmpeg spawn after headers are written.
- `video_to_h264` availability requires `libx264` specifically, terminal-failing hardware-only ffmpeg builds the pipeline could serve; legacy bps-valued bitrates in 1–10M are misread and force transcodes; attempt keys are compared case-insensitively (`containsFoldV3`) despite being case-sensitive hashes; the quirk-identity hash pre-image is variable-arity (fixtures only pin the no-quirk case — needs a with-quirk fixture so the Kotlin client stays in lockstep).

## What's good

Deterministic key machinery is genuinely deterministic (sorted slices everywhere, charset-restricted client inputs, no map iteration feeding hashes); the fallback ladder provably terminates; the lease/digest idempotency model with completed-response replay is the right design; `AcquireSessionLock` handles the uncertain-rollback case correctly; the attempt-revision migration's trigger story for rolling deploys is actually correct; old stream tokens decode to exact legacy behavior at every consumer; BSF strings are strictly allowlisted before reaching ffmpeg; path traversal on reconstruction dirs is blocked and tested; scanner/capabilities legacy-data handling means no library rescan is required; the shadow handler is clean (read-only, no store writes, no resource acquisition).

## Test coverage

Good on happy paths, seek-pinning invariants, idempotent replay, and the rate limiter. The systematic gap: **everything runs against the in-memory store and fake registries/probes**. No test exercises the Postgres planstore against real schema (would have caught #2 immediately), real ffmpeg probing (#1), transport rollback paths, cross-user ownership rejection, hostile track indexes, reservation leaks under concurrency, or DV-mode transitions. One DB-backed test inserting each allowed event name is the single highest-value test to add.

---

*Review method: 4 parallel domain reviews (handlers/routing, planner core, sessions/transcode/tokens, persistence/migrations) with adversarial verification of every finding against the PR-state code at `03fc9de8`; CodeRabbit/Codex bot findings cross-checked and incorporated or superseded. The two pre-existing jellycompat process-lock test failures reproduce on clean main and are unrelated.*

Quick104:
# Review remediation — pushed

Four commits address the findings from the deep review above (on top of `03f9b57b`, which had already fixed the CHECK-constraint drift, the replan validation panic, and the non-atomic session replacement):

- **`930ecfa9` — restore `dovi_rpu` strip filter** *(critical #1, #6-related, F3, F4, F11-proxy)*: `dovi_split=mode=bl` does not exist in any deployable FFmpeg; every probe, arg, constant, and test is back on `dovi_rpu=strip=1`. Remux DV modes are now validated for every profile, P7 preserve fails fast instead of emitting dangling-RPU output, the `dvhe` sample-entry tag is opt-in for the v3 preserve recipe only (legacy web/jellycompat keep `hev1`), and the proxy honors the token-frozen DV mode.
- **`ec0fb57f` — planner policy & contract validation** *(#6, quirk gate, hdr_unknown, metered/cap, engine-aware subtitles, bounds, exact keys, h264 probe, aliases, strict track IDs, with-quirks key fixture)*: P7 can no longer take the non-strip remux route while claiming native DV; the full list is in the commit message.
- **`355fdbf7` — control-plane reliability** *(#5 pool starvation via replan slot semaphore, CompleteReplan CAS + `ErrReplanSupersededV3`, start request digests + replay liveness, expired-attempt recovery, error-code mapping, RemuxDVMode clearing, replacement-admission grace arithmetic, reservation survival, durable-key merge into replan exclusions, same-file track-remap guard, ID-only subtitle remap, flag caching, route-event identity projection + rate-limit ordering, deterministic diagnostics, `received_at` index, trigger churn guard, legacy audio-change re-mint carrying `tid`/`dvm`)*.
- **`515f8f26` — node/stream hygiene** *(F5 idle-job reaper on the transcode node, PGS→VTT 415 gate before headers, response-body drain, cleanup-boundary regression test — the prefix-matching concern turned out already correct and is now pinned)*.

**New test surface**: DB-backed planstore tests (`SILO_TEST_DATABASE_URL`-gated) exercise the real schema — including a regression test inserting every route-event name against the actual CHECK constraint, the SaveAttempt digest/expiry semantics, the replan lease lifecycle, CompleteReplan CAS, and advisory-lock serialization. 7/7 pass locally against a migrated Postgres; they skip cleanly when the env var is unset.

**Verification**: `go build ./...`, `go vet ./internal/...`, and the test suites for `internal/playback` (+`planstore`), `internal/api/...`, `internal/transcodenode`, `internal/proxy`, and `migrations` all pass. The NVENC/HWAccel probe tests flake only under parallel package load (pre-existing; pass consecutively in isolation). gofmt/golangci findings in touched files match the pre-existing branch baseline.

## Known items deliberately not addressed here

- **Mixed-fleet proxies**: node capability gating (from `03f9b57b`) covers new-server/old-node, but an **old proxy binary** still routes by session ID — sequence proxy upgrades with the server, or v3 remote transcodes can 404/duplicate until proxies restart on the new build.
- **Swap-window segment aliasing** (F7): in-flight old-plan segment requests during a quality-switch commit can still poke the successor's seek-recovery; transient stall only, needs a generation-tagged URL space to fix properly — follow-up.
- **`empty_moov` → `delay_moov`** stays global (correctness for copied E-AC-3/Atmos outweighs ~one fragment of startup latency); worth a line in release notes.
- **Replan budget**: instead of a hard per-session count (which would break long sessions where quality changes are replans), the server now always merges its durable plan key into the exclusion set, which kills the re-serve-the-failed-plan loop; `attempt_count` stays advisory.
- **Migration caveat**: `20260712103000` and `20260713170000` were edited in place (branch-unreleased on main, but any dev host that already applied them needs manual reconciliation — the usual goose-row dance).
- Legacy `/playback/start` now has a 256KB body cap and a `protocol_version` pre-parse (pre-existing on this branch) — mild existing-endpoint behavior change worth a sentence in the PR description.

Quick104:
# Second remediation round — pushed

Every open review comment was re-validated against `515f8f26` before acting. Four commits address the confirmed findings; the rest are answered below.

## Fixed

- **`1c3e73ff` — v3 planner policy gaps**
  - *Bandwidth caps below 1500 kbps* (codex, `plan_v3.go`): confirmed — the ladder bottoms out at 480p/1500 kbps and the final bitrate was never clamped, so caps in [100, 1499] were silently exceeded. The encode target is now clamped to the cap (the 480p resolution floor stays).
  - *Silent sources* (codex, `capabilities_v3.go`): confirmed — video-only files fell through audio eligibility as `audio_codec_unsupported`. Now treated as audio-compatible, but only when the file genuinely has no audio track; a track whose codec failed to probe still converts rather than copies.
  - *Bitmap subtitle sidecars* (codex ×2, `subtitle_policy_v3.go`): both confirmed and share one root cause. The stream handler raw-serves exactly one bitmap shape — embedded PGS as `.sup`; external/downloaded bitmap goes through WebVTT conversion (impossible) and embedded DVD/DVB is rejected outright. The policy now promises a sidecar only for embedded PGS with `EmbeddedBitmap`; everything else falls to burn-in or its terminal instead of publishing an artifact that always fails at fetch.
  - *Top-level client transform feature* (codex, `protocol_v3.go:739`): confirmed inconsistency with the planner's dual-source read; validation now accepts `client_video_transformations_v1` from either `client_features` or the nested context.

- **`5bb3300f` — one ffmpeg binary for probe and execution** (codex, `remux.go`): confirmed. The registry probed the configured `playback.ffmpeg_path` while progressive remux execution resolved the process-global discovery path, so `server_dv7_to_hdr10` could be advertised and then fail at stream time. Both now resolve through a shared `ResolveFFmpegPath` (the same rule the transcode pipeline already used), the `dovi_rpu` probe is cached per binary, and the stream handler and proxy worker pass their configured path through.

- **`9ad92974` — replan identity + control-plane limits**
  - *Replans defaulting to current tracks* (codex): confirmed — after an alternate-version fallback, `NormalizedRequest` still carries requested-edition track IDs, so a failure replan omitting unchanged tracks failed `NormalizeAndValidate` as a track/file mismatch. Failure replans now seed selections from `record.CurrentPlan.SelectedTracks` before overlaying the request (the seek path already did exactly this). Intent detection also now compares against what is actually playing.
  - *ID-only audio selections* (codex): confirmed — the remap guard skipped ID-only selections; a new `remapAudioSelectionV3` mirrors the subtitle remap (parse ID → index → remap → rebind).
  - *Reservation leak on remote rollback* (codex): confirmed — the rollback closure stopped the node job but left the `PlanSession` reservation to age out (up to 90 s of pinned max-job/bandwidth budget per failed start). Rollback now releases it, matching the prepare-error paths.
  - *Replan semaphore vs. DB pool* (codex): confirmed — with `MaxConns ≤ 8`, advisory-lock holders could starve their own inner store queries. The store now advises a capacity (half the pool, min 1) and the handler sizes the semaphore to the smaller of that and the fixed bound.
  - *Shadow goroutine recover* (CodeRabbit ×2): confirmed — added a recover boundary with a stack log inside `shadowLegacyPlaybackV3`.

- **`a05067e3` — transcode-node teardown races** (CodeRabbit + codex ×2, all confirmed, one design): manifest/segment lookup and idle-clock touch are now one critical section; the reaper re-validates each victim under the per-session lifecycle lock before `Close()` (which removes the output dir a concurrent reconstruct may be writing into); and `handleStop` takes the lifecycle lock so a stop racing a `RequireReady` start's readiness wait blocks until registration and tears the job down instead of 404ing and orphaning the ffmpeg.

## Already fixed before these comments were filed

- *Subtitle ID-only remap before fallback* (codex, 11:17 UTC): `remapSubtitleSelectionV3` has parsed ID-only selections to an index since the previous remediation push.
- *Runtime-correction events vs. CHECK constraint* (codex, 11:17 UTC): the migration has allowed all three `runtime_correction_*` events since `03f9b57b` (lines 69–70).
- *Stale `RemuxDVMode`* (codex, 11:17 UTC): `applySessionStreamStateLocked` assigns unconditionally when `TranscodeRouteSet` is true, with a regression test (`TestUpdateStreamStateClearsRemuxDVModeOnRouteSet`).

## Rejected after validation

- *VideoTrack JSONB legacy keys* (CodeRabbit, critical): the snake_case tags are **unchanged from main** — the PR diff only adds `dv_enhancement_layer` (additive, with an explicit legacy-rows fallback in `SourceDescriptorFromFileV3`) and realigns struct whitespace. Existing rows hydrate exactly as before.
- *Substring capability probes* (CodeRabbit): `" aac "` is already space-guarded, `-bsfs` output is a bare name list (no descriptions to false-match `dovi_rpu`), and `libx264rgb` cannot exist in a build without `libx264` — no realistic misclassification for any probed name.
- *Memory store `AcquireSessionLock` no-op* (CodeRabbit): sound by design — `HandleReplanPlaybackV3` takes the in-process per-session `lockReplanV3` before the store lock, and the memory store only ever backs single-process DB-less deployments; the store lock exists for cross-process serialization over shared PostgreSQL. Documented on the method.

## Deferred (follow-up)

- *Plan v3 routes from remote capabilities* (codex): real limitation, but architectural — fixing it needs node enumeration + capability union at planning time **plus** a new local-capability validation on the fallback path (a naive union would let local fallback execute transformations the local binary lacks). Today such deployments get retryable terminals and clients fall back to legacy routes, which remote-only toolchains already serve. Tracking as a follow-up rather than growing this PR.

Quick104:
The deferred remote-capability planning item (codex comment 3579105380) is now implemented in #408, stacked on this branch: planning draws on a per-executor-pool registry split (local for progressive remux routes, local ∪ pooled node capabilities for HLS routes), with negative-cached capability fetches and local-fallback validation so no executor is ever handed a recipe it cannot run.

PR #408, opened 2026-07-14
feat(playback): plan v3 routes from pooled node capabilities

Quick104:
## Problem

Protocol v3 planning gated every server transformation on the API host's local ffmpeg probe (`ProbeTransformationRegistryV3`). In transcode-node deployments where the toolchain lives on the nodes — libx264/aac/dovi_rpu on nodes, a minimal binary locally — `PlanPlaybackV3` returned `conversion_tool_unavailable` / `audio_conversion_unsupported` terminals before `prepareTransportV3` ever fetched the selected node's capabilities, even though the per-node validation path was already built to check them. Flagged in #398 review (comment 3579105380) and deferred as architectural.

## Why this shape

A naive union registry would be **wrong**: progressive remux routes (DV strip, remux audio conversion) execute on the API host's ffmpeg and can never offload — planning them from node capabilities would produce plans that fail at stream time. So availability is split by executor pool:

- **`Registry`** (unchanged) — the local probe; still gates progressive remux routes.
- **`HLSRegistry`** (new, optional) — local ∪ pooled node capabilities; gates HLS remux/transcode routes, which `prepareTransportV3` can place on a node.

The union only widens **availability of locally-defined specs** at the exact local recipe version — advertisements never introduce new transformations — so any plan built from it is guaranteed to pass the existing per-node advertisement validation when that node is selected.

## What changed

- `PlannerInputV3.HLSRegistry` + per-branch gating: the remux family now skips the progressive remux and ships the same recipe on the HLS remux delivery when only nodes carry a required transformation; HLS transcode eligibility (`videoTranscodeExecutableV3`, `planVideoTranscodeV3`) and the HLS EAC3 quirk gate on the union.
- `nodepool.Planner.TranscodeNodeURLs()` enumerates pooled nodes (enabled, healthy or not — unreachable nodes exclude themselves when their fetch fails).
- `hlsPlanningRegistryV3` builds the union at planning time: concurrent fetches for stale cache entries under a 3s planning deadline, riding the existing per-node TTL cache. Fetch **failures are now negatively cached** (15s) so a dead node costs one timeout per window, not one per start.
- `prepareTransportV3`'s local fallback validates the plan against the local registry's `Advertised()` set (same validator as the node path, renamed `validateAdvertisedTransformationsV3`) and returns the existing retryable `transcode_node_capability_unavailable` terminal instead of spawning a doomed ffmpeg. Transformation-free plans skip the check entirely, so nothing changes for plain copy routes.
- Shadow planning uses the same union for fidelity with production decisions.

When no nodes are pooled (or the planner doesn't enumerate), the union is exactly the local registry — single-host deployments are byte-for-byte unaffected.

## Terminals (requirement 3)

No node selected + local fallback disabled → existing `capacity_unavailable` (retryable). Selected node fails validation → existing `transcode_node_capability_unavailable` (retryable). Local fallback can't execute → same retryable reason; a capable node freeing up satisfies the same plan.

## Not in scope

`HandlePlaybackCapabilityV3` still advertises only locally-probed transformations (clients don't gate requests on it; under-advertising is the safe direction). Can follow up if client feature detection wants the union.

## Testing

- New planner tests: HLS transcode offload, remux audio-conversion offload (progressive skipped, HLS remux selected; local-capable keeps progressive priority), P7 DV-strip offload to HLS remux, union registry semantics (version pinning, no new specs, client-executor advertisements ignored, receiver immutability).
- New handler tests: union building against an httptest node, non-enumerating planner passthrough, negative caching (one fetch per window), local-fallback rejection with the retryable terminal.
- `go build ./...`, targeted package tests green (`internal/playback`, `internal/api/handlers`, `internal/nodepool`; the NVENC probe and one jellycompat test remain the known under-load flakes, passing in isolation). Lint on new lines: only test-file goconst (CI-excluded).

Part of the v3 playback protocol work in #398; stacked on `codex/playback-protocol-v3-plan`.

## AI-use disclosure

Implemented by Claude (Fable 5) under human direction, as a follow-up flagged during #398's bot review.

## Summary by CodeRabbit

* **New Features**
  * Improved playback planning by considering transformation capabilities advertised by available transcoding nodes.
  * Selects compatible transcoding nodes for server-side transformations.
  * Supports fallback to local playback capabilities when remote capability checks fail.
  * Improves HLS transcoding, audio adaptation, and HDR playback planning across varied device capabilities.

* **Bug Fixes**
  * Prevents unavailable transformations from being selected for local playback.
  * Preserves retryable playback outcomes for progressive-only clients.

* **Tests**
  * Added coverage for capability-aware planning, node selection, fallback behavior, and HLS transformation support.

### DIFF
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -35,10 +35,18 @@ const (
 	maxPlaybackV3EventBodyBytes = 32 << 10
 	replanLeaseDurationV3       = 15 * time.Second
 	v3NodeCapabilityTTL         = time.Minute
+	// Failed capability fetches are memoized briefly so an unreachable node
+	// costs one timeout per window instead of one per planning request.
+	v3NodeCapabilityErrorTTL = 15 * time.Second
+	// Capability fetches on the planning path run under a deadline well below
+	// the fetch helper's own 10s timeout: planning happens on the start
+	// request path, where a slow node must degrade the union, not the user.
+	v3NodeCapabilityPlanTimeout = 3 * time.Second
 )
 
 type v3NodeCapabilityCache struct {
 	transformations []playback.TransformationV3
+	err             error
 	expiresAt       time.Time
 }
 
@@ -145,11 +153,23 @@ func (h *PlaybackHandler) remoteTransformationsV3(ctx context.Context, nodeURL s
 	entry, ok := h.v3NodeCapabilities[nodeURL]
 	h.v3NodeCapabilitiesMu.Unlock()
 	if ok && now.Before(entry.expiresAt) {
+		if entry.err != nil {
+			return nil, entry.err
+		}
 		return append([]playback.TransformationV3(nil), entry.transformations...), nil
 	}
 
 	info, err := fetchRemoteTranscodeCapabilities(ctx, nodeURL, h.JWTSecret)
 	if err != nil {
+		// Memoize the failure briefly: with capability-union planning every
+		// start consults the pool, and an unreachable node must not add its
+		// fetch timeout to each of them.
+		h.v3NodeCapabilitiesMu.Lock()
+		if h.v3NodeCapabilities == nil {
+			h.v3NodeCapabilities = make(map[string]v3NodeCapabilityCache)
+		}
+		h.v3NodeCapabilities[nodeURL] = v3NodeCapabilityCache{err: err, expiresAt: now.Add(v3NodeCapabilityErrorTTL)}
+		h.v3NodeCapabilitiesMu.Unlock()
 		return nil, err
 	}
 	entry = v3NodeCapabilityCache{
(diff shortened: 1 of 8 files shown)


Quick104:
# Review remediation — pushed

All four findings validated as real and fixed in `31ad2e78`. Point-by-point:

- **Choose a capability-matching node for union plans (P1)** — confirmed: `prepareTransportV3` made one load/affinity `PlanSession` call and validated only that node, so a heterogeneous pool could land an offload-only recipe on an incapable node while a capable sibling sat idle. `nodepool.Planner` gains `PlanSessionWith(..., eligible func(*Node) bool)` (group health still computed over the full pool; the predicate is a set lookup under the planner lock, never a network call), and the new `planNodeSessionV3` restricts selection to nodes whose cached advertisements validate against the plan — including breaking soft affinity to a current node that is no longer eligible. Transformation-free plans and non-enumerating planners keep pure load-based selection. Covered by `TestPlanSessionWithRestrictsEligibleTranscodeNodes` and an end-to-end heterogeneous-pool test (`TestPlanNodeSessionV3PrefersCapabilityMatchingNode`: busy capable node wins over idle incapable one).

- **Don't cache planning timeouts for transport validation (P1)** — confirmed: the negative cache memoized `context.DeadlineExceeded` from the 3s planning fetch and the transport path read the same entries, so a healthy node with a 3–10s capability endpoint would be rejected after selection. The cache is now split by consumer: `remoteTransformationsPlanningV3` honors memoized failures (dead nodes cost one timeout per 15s window), while the transport-time `remoteTransformationsV3` fetches through them with its own 10s budget — restoring exactly the pre-PR transport behavior. A transport-path success also replaces the failure entry for subsequent planning. Covered by `TestRemoteTransformationsV3FailureCacheSplit`.

- **Require HLS support before using node-only remux tools (P2)** — confirmed regression: a progressive-only client needing audio conversion fell through to non-retryable `adaptation_unavailable` instead of the retryable `audio_conversion_unsupported`, because the union suppressed the terminal for a delivery the client can't run. The audio gate is now `local || (hlsEngineOK && union)`, and the DV-strip union flag is gated on the HLS engine the same way (the transcode-path gates already checked the engine first). Covered by `TestPlanPlaybackV3NodeToolchainDoesNotMaskTerminalForProgressiveOnlyClient`.

- **Avoid fetching node capabilities for direct-play starts (P2)** — confirmed: the widened registry was built eagerly per planning call, so a dead pooled node added its 3s deadline to every start each negative-cache window, including pure direct play. `PlannerInputV3.HLSRegistry` is now a lazy producer (memoized per request in the handler), and every consulting gate sits behind short-circuits — non-DV sources never reach the DV gate's registry argument, locally-satisfiable audio conversion never consults it, and the HLS-transcode gates check engine/settings first. Direct-play and source-preserving starts never touch the pool. Covered by `TestPlanPlaybackV3DirectPlayNeverConsultsNodeCapabilities` (the producer calls `t.Fatal` if invoked).

Verification: build green; `internal/playback`, `internal/api/handlers`, `internal/nodepool` tests pass (the two NVENC probe failures in the parallel run are the known under-load flakes, passing in isolation). Lint on new lines: only the pre-existing non-blocking goconst literal pattern.

PR #416, opened 2026-07-16
fix(scanner): recover malformed video durations

RXWatcher:
## Summary

- stop treating large ffprobe durations as microseconds; ffprobe reports these values in seconds
- recover malformed video durations from a sane video stream, an absolute timestamp offset, or a streaming packet timestamp/frame-count scan
- reprobe the narrow legacy signature of videos larger than 500 MiB that were stored as ten seconds or less
- preserve legitimate long-form audio durations, including audiobooks longer than 27.8 hours

## Root cause

Malformed container, audio, or subtitle timestamps can make ffprobe report a very large duration. The scanner divided every duration over 100,000 by 1,000,000, turning affected movies and episodes into 1-4 second database values. The player then correctly clamped seeking to that incorrect backend duration. Some malformed files instead report a directly collapsed duration; for those, packet count and frame rate provide the usable timeline.

## Validation

Production samples covered all fallback paths: a valid video stream duration, absolute end-minus-start timestamps, a malformed subtitle timeline, and collapsed packet timestamps.

- `go test -race ./internal/scanner`
- `go vet ./internal/scanner`
- `git diff --check`

### DIFF
--- a/internal/scanner/probe.go
+++ b/internal/scanner/probe.go
@@ -1,9 +1,12 @@
 package scanner
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"math"
 	"os/exec"
 	"slices"
 	"strconv"
@@ -50,6 +53,7 @@ type ffprobeFormat struct {
 	Filename       string            `json:"filename"`
 	FormatName     string            `json:"format_name"`
 	FormatLongName string            `json:"format_long_name"`
+	StartTime      string            `json:"start_time"`
 	Duration       string            `json:"duration"`
 	Size           string            `json:"size"`
 	BitRate        string            `json:"bit_rate"`
@@ -69,6 +73,8 @@ type ffprobeStream struct {
 	DisplayAspectRatio string              `json:"display_aspect_ratio"`
 	FieldOrder         string              `json:"field_order"`
 	AvgFrameRate       string              `json:"avg_frame_rate"`
+	StartTime          string              `json:"start_time"`
+	Duration           string              `json:"duration"`
 	BitRate            string              `json:"bit_rate"`
 	ColorTransfer      string              `json:"color_transfer"`
 	ColorPrimaries     string              `json:"color_primaries"`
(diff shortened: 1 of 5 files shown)


RXWatcher:
@coderabbitai review

Quick104:
Ran a deep review of this PR (8 finder angles, adversarial verification per finding) and pushed `fec48391` addressing the confirmed findings:

**Correctness**
- **Cover art counted as video**: ffprobe reports `attached_pic` streams as `codec_type: video`, so audiobooks/music with embedded artwork skipped the audio duration path, and long books either hard-failed `ProbeFile` (aborting the whole audiobook/podcast import) or persisted `Duration=1` from the cover's frame rate. Now filtered via `disposition.attached_pic`.
- **ProbeFile hard-fail discarded good metadata**: a failed packet fallback returned `nil, err`, so the scanner upserted rows with *no* probe data (codecs/tracks/resolution lost) where pre-PR they imported with Duration=0. Now the partial probe survives and the repair layer retries the duration.
- **Library scans never repaired legacy durations**: the scan path uses `needsCriticalProbeRepairScanState`, which wasn't updated — the legacy reprobe only fired lazily at detail/playback time (blocking the request up to 60s). The scan-state predicate now knows the legacy signature.
- **Non-converging repair loop**: a genuinely short ≥500MiB clip re-matched `needsLegacyDurationRepair` after every reprobe, re-running the full-file packet scan on every detail load/playback forever (the anti-pattern the ebook comment in the same file warns about). The repair is now one-shot: rows with `ProbeUpdatedAt` after `legacyProbeDurationFixTime` are authoritative. ⚠️ that cutoff is set to 2026-07-18 — bump it if this merges later.
- **Timeout mismatch**: the 60s grant only covered the legacy shape; `Duration<=0` video rows ran the new whole-file packet scan under the default 5-10s timeout, guaranteed to die and retry on every request. The grant now covers any reprobe likely to hit the fallback.
- **end-minus-start bypassed the plausibility guard**: a collapsed absolute-timestamp span (e.g. 3s for a 2GB file) was persisted instead of reaching the packet fallback. Both `durationAfterStart` paths now apply `durationLooksImplausiblyShort`.
- **Audio path lost its cap entirely**: preserving >27.8h audiobooks needed a raised ceiling, not none — malformed audio containers could persist multi-year durations with no repair path able to fix them (Duration>0, no video tracks). Capped at ~277h.

**Maintainability**
- The implausibly-short heuristic existed twice with divergent thresholds (10s/100MiB in probe.go vs 10s/500MiB in probe_repair.go), leaving 100-500MiB legacy rows unrepairable. Now one shared predicate + constants used by probe, repair, and scan-state layers.
- `parseFrameRate` was a new in-file duplicate of `normalizeFrameRate`'s fraction parsing; consolidated within the scanner package. (Follow-up: `parseFrameRateV3` in playback and the jellycompat copy still exist — cross-package consolidation left out of scope here.)

**Noted, not changed**: the packet scan intentionally demuxes the whole file — the timestamps being repaired are the ones interval seeking would rely on, so sampling can't be trusted; the one-shot repair bounds the cost. Documented in a comment.

Validation: `go test -race ./internal/scanner`, `go vet`, `gofmt` clean; remaining golangci hits on touched files are the known pre-existing `goconst` baseline.

PR #425, opened 2026-07-18
feat(playback): balance transcode sessions across multiple GPUs

RXWatcher:
Part of #449.

## Summary

`playback.hw_device` accepts a comma-separated render-device list (e.g. `/dev/dri/renderD128,/dev/dri/renderD129`). Every GPU workload — streaming transcode sessions, prepared downloads, chapter thumbnails — resolves the list to exactly one concrete device through a single acquisition boundary (`AcquireHWDevice` in `internal/playback/hwdevice.go`) immediately before ffmpeg launch: the present device with the fewest active workloads (ties keep list order). The reservation is held until that ffmpeg process has exited — sessions release after the process is reaped (seek/audio-switch restarts reuse the device), downloads and thumbnails release when their synchronous run returns.

Backward compatibility is exact: a single configured value keeps the historical pass-through contract for every accelerator, and an empty value still auto-detects.

## Supported topology / compatibility assumptions

- **Balancing is QSV/VAAPI-only.** NVENC identifies GPUs by CUDA index/UUID rather than render-node path; a multi-entry list with NVENC logs a warning and uses the first entry (covered by tests). Accelerator-specific device identities are out of scope for this version (#449).
- **The setting is cluster-wide.** Every transcode node reads the same `playback.hw_device`, so multi-node deployments must expose the listed devices at identical paths on every node (documented on the config field). `/admin/system/hw-accel` now probes **every** healthy node and returns per-node inventories; the admin UI warns when node inventories diverge and marks devices missing on specific nodes.
- Devices absent on a host fall out of that host's rotation; if none of the listed devices exist the first entry is used so the failure mode stays deterministic.

## Review follow-ups addressed (from the changes-requested review)

1. Fresh `/transcode/start` node sessions now feed the node-local configured value into the same resolution path as reconstruction — no more empty-device bypass.
2. Streaming (`StartTranscode`), downloads (`PrepareFile`), and chapter thumbnails (`ExtractFrame`) all acquire through the one shared path with load accounting; thumbnails no longer pass a raw CSV to ffmpeg as a device path.
3. Compatibility modeled explicitly: v1 constrained to homogeneous QSV/VAAPI render devices; NVENC multi-list behavior is explicit, warned, and tested (the old presence-filter collapse of `0,1` → `0` is gone because NVENC never enters the balancer).
4. Sessions release the GPU reservation only after ffmpeg has been reaped (`shutdown` waits on the monitor goroutine first).
5. Node-aware inventory endpoint + documented homogeneous-path contract (see above).
6. The device picker renders the union of configured ∪ detected devices, so configured-but-missing devices stay visible when detection is empty or an older node omits `render_device_details` (falls back to plain `render_devices` paths).
7. Typed `HWDeviceSet` parsed once; the allocator file (`hwdevice.go`) owns only selection/reservation; sysfs inventory and operator-facing descriptions live with GPU detection (`gpudetect.go`).
8. Integration-level coverage: fresh node start resolution, thumbnail arg construction from a list, NVENC multi-list behavior, reservation release after process exit (and on spawn failure), all-devices-missing UI rows, heterogeneous node inventories (aggregation endpoint + divergence detection).
9. `pnpm run format:check`, `pnpm run lint`, `tsc -b`, `gofmt`, `go vet` clean on touched files.

## Testing

- `internal/playback/hwdevice_test.go`: parse/acquire semantics incl. NVENC no-reserve, all-missing fallback, idempotent release, exact concurrent split under `-race`.
- `internal/playback/transcode_hwdevice_test.go`: reservation held until `CloseProcess` reaps a real (fake-binary) ffmpeg; released on spawn failure.
- `internal/playback/prepare_file_test.go`, `internal/chapterthumbs/service_test.go`: one concrete device in built args, never the raw list; release after run.
- `internal/transcodenode/server_test.go`: fresh `/transcode/start` resolves the configured list.
- `internal/api/handlers/system_test.go`: aggregation across healthy nodes, failed-probe entries, all-probes-fail local fallback.
- `web`: 14 vitest cases for row union/fallback/divergence; `pnpm run format:check` green.
- `go test -race` green for `playback`, `chapterthumbs`, `transcodenode`, `api/handlers` (the four NVENC ffmpeg-probe tests and two jellycompat process-lock tests fail on unmodified `main` on macOS; unrelated).

Admin UI screenshot: to follow.

## Summary by CodeRabbit

* **New Features**
  * Added per-render-device selection UI for `playback.hw_device`, showing per-node availability and a cluster-safety warning.
  * Enhanced `GET /admin/system/hw-accel` to aggregate results across all healthy transcode nodes, including per-node render-device inventory/details.
* **Bug Fixes**
  * Ensure transcoding and frame extraction resolve multi-device lists to a single concrete device before invoking FFmpeg.
  * Improved GPU reservation lifecycle, including correct release on early failures and proper handling during restarts.
* **Documentation / UX**
  * Updated admin payload/UI to surface richer render-device information and NVENC multi-GPU limitations.

---

### Maintainer Remediation AI Disclosure
- Tool(s): Codex, T3 Code
- Model(s): gpt-5.6-sol
- Involvement: AI-assisted
- Adversarial review: Found that GPU reservations survived a natural FFmpeg exit until session shutdown, which could permanently skew load balancing. Commit c474a6c2 releases each reservation immediately after process reaping, reacquires the same concrete GPU for a restart, and releases it on restart spawn failure. Final review retained three report-only items: device eligibility checks existence rather than openability; the picker omits GPUs detected only on secondary responding nodes; and the required admin UI screenshot is still absent because no T3 preview automation host was available.
- Verification: go test -race passed for internal/playback, internal/chapterthumbs, and internal/transcodenode; the two PR-added internal/api/handlers hardware-acceleration tests passed under -race. go vet on affected packages, changed-code golangci-lint, make build, make verify-local-paths, frontend ESLint and Prettier checks, 14 targeted Vitest cases, and git diff --check passed. The full handler package still has the unrelated pre-existing TestHandleReplanPlaybackV3SeekFailureRecoveryNeverChangesMediaVersion failure. Under the shared backend lock, two sequential real QSV transcodes selected /dev/dri/renderD128 and exited cleanly, demonstrating release and reuse; probe sessions and the temporary device setting were removed, the backend was redeployed, and /api/v1/ready returned status ok.


RXWatcher:
Added the admin UI on top: the hw-accel detection endpoint now reports `render_device_details` (per-device sysfs vendor/model labels), and Playback settings renders GPU devices as per-device toggles — none selected = auto, one = pin, several = balance. The stored `playback.hw_device` stays the comma list underneath, written in stable detection order. Utils covered by vitest; the three failing web test files on this branch fail identically on unmodified `main`.

RXWatcher:
Addressed the atomicity finding in b8595cc6: presence checks moved outside the lock, selection + reservation now share one critical section, and a new `-race`-clean concurrency test asserts an exact 4/4 split for eight simultaneous session starts across two devices — which the previous two-step selection could not guarantee.

Quick104 (changes requested):
The architectural direction is sound: a transcode node should be able to schedule work across the compatible GPUs local to that host, rather than requiring one daemon per GPU. Multiple nodes should remain available for host/process isolation and heterogeneous policies.

I’m requesting changes because the current implementation does not apply that model consistently and is not safe across the full advertised surface.

Please address all of the following before this is approved:

1. **Make normal transcode-node sessions use the device pool.** Fresh `/transcode/start` requests currently construct `TranscodeOpts` with `HWDevice: ""`, so they auto-detect the first GPU and bypass the configured list. Reconstruction uses `cfg.Playback.HWDevice`, which makes fresh and reconstructed sessions inconsistent.

2. **Use one shared acquisition path for every GPU workload.** Chapter thumbnails currently pass a non-empty comma-separated value directly to FFmpeg as one device path. Downloads are only best-effort and are not represented in the same active-load accounting. Streaming, prepared downloads, and thumbnails should all resolve exactly one compatible device immediately before process launch and release it at the correct lifecycle boundary.

3. **Model accelerator compatibility and device identity correctly.** Presence plus session count is not enough. The UI exposes every DRM render node for QSV, VAAPI, NVENC, and auto, so mixed-vendor selections can pair an accelerator with an incompatible GPU. CUDA/NVENC device selection uses an index or UUID, not a `/dev/dri/renderD*` path; the current filesystem-presence filter also makes a manual `0,1` list collapse to device 0. Either explicitly constrain this first version to homogeneous compatible QSV/VAAPI render devices or introduce accelerator-specific device identities/capabilities.

4. **Hold reservations until FFmpeg has actually stopped.** `shutdown` releases the GPU before cancelling the process and waiting for `done`, so another start can select the supposedly free device while the old FFmpeg process is still running.

5. **Fix the multi-node configuration model.** The admin endpoint reports whichever healthy node currently has the fewest jobs, but saving writes one cluster-wide `playback.hw_device` value consumed by every node. This is only correct under an undocumented identical-device-path assumption. Make inventory/configuration node-aware, or explicitly design and validate the homogeneous-path contract.

6. **Keep configured devices visible when detection is empty or unavailable.** The picker is gated on `renderDevices.length > 0`, so all configured-but-missing devices disappear when every device is absent or an older node does not return `render_device_details`.

7. **Simplify ownership.** Keeping CSV as backward-compatible storage is reasonable, but parse it once into a typed internal device set. The allocator/reservation logic should have one small ownership boundary; sysfs inventory and operator-facing descriptions should remain with GPU detection rather than sharing `hwdevicebalance.go`.

8. **Add integration-level coverage for the real call paths.** Please cover fresh transcode-node starts, chapter-thumbnail argument construction from a list, accelerator/device compatibility (including NVENC behavior or an explicit rejection), reservation release after process exit, all-devices-missing UI behavior, and heterogeneous node inventories. The current allocator unit tests and race test pass but do not exercise the paths where these failures occur.

9. **Complete the repository gates and PR requirements.** `pnpm run format:check` currently fails on the three changed TS/TSX files. Please also link the relevant scope issue/capability item, add a screenshot or recording of the admin UI, and document the supported topology/compatibility assumptions.

Once these are addressed, the result should be both simpler operationally than one process per GPU and consistent with the existing node-pool architecture.

RXWatcher:
All nine items addressed in 8bee418a…d08b162e; the PR description now carries the supported-topology section and a per-item map. The load-bearing changes:

- **One acquisition boundary** (`internal/playback/hwdevice.go`): `AcquireHWDevice(configured, resolvedAccel)` is the only way any workload turns the configured value into a device. Streaming sessions, `PrepareFile` downloads, and chapter-thumbnail extraction all resolve exactly one device at launch and participate in the same load accounting; thumbnails no longer hand ffmpeg a raw CSV. `PickRenderDevice` is auto-detection-only again.
- **Release boundary** (item 4): `shutdown` releases after waiting on the reap channel, so a device can't be re-selected while the old process is alive. Covered by a fake-binary test that asserts the count across `CloseProcess`.
- **Compatibility** (item 3): took the constrain-v1 option — balancing is QSV/VAAPI render devices on a homogeneous list; NVENC multi-lists warn and use the first entry (tested), and the UI hides the multi-select for NVENC. CUDA index/UUID identities are deferred to #449.
- **Cluster contract** (item 5): kept the single cluster-wide value, made it honest — `/admin/system/hw-accel` probes every healthy node and returns per-node inventories (flat fields unchanged for compatibility), the config field documents the identical-paths requirement, and the UI warns on divergence and marks per-row which nodes lack a device.
- **Item 1**: fresh node starts now pass the node-local config value into the same resolution path as reconstruction.
- **Item 6**: picker rows are configured ∪ detected, with a `render_devices`-only fallback for older nodes.

Scope issue: #449. Formatting/lint/vet green; test inventory is in the PR body. UI screenshot to follow.

PR #430, opened 2026-07-19
docs(design): Client Diagnostics — crash reports and opt-in debug log upload (cross-repo spec)

Quick104:
## What

**Spec only — no code.** This draft PR adds `docs/design/2026-07-19-client-diagnostics.md`: a cross-repo design + rollout plan for **Client Diagnostics** — opt-in crash reporting and debug-log upload from the Silo clients (Android phone/TV, iOS, tvOS) to the user's **own** Silo server, plus the server-side ingest, storage, admin, and retention work that lands in this repo.

It's opened here because the server is the hub: the payload contract (versioned JSON Schema + fixtures), the feature gate, the ingest endpoint, storage, retention, and the admin surface all live in `silo-server`. Matching client work in `silo-android` and `silo-apple` is phased in the spec.

## Why

We currently have zero visibility into client-side failures:

- No crash reporting exists in either client repo (no handler, no MetricKit, no SDK).
- Diagnostics are ephemeral — logcat on Android, `print()`/`[CMP-DIAG]` console streams on Apple. Nothing survives to the moment a user reports a problem.
- Our hardest bugs are device-specific (Shield HLG, Fire TV passthrough, HDMI mode switches, TV focus traps) and users can't report panel/audio-chain/codec facts.
- The server already records `playback_session_id` in `operational_logs` — client bundles carrying session IDs let admins join both halves of a playback incident.

## Why this route (short version — full rationale in the doc)

- **User's own server as destination, not a third-party crash service**: self-hosted product, self-hosted diagnostics; the recipient is the admin who can actually act; no vendor accounts/keys.
- **Consent-first and off by default everywhere**: server gate (`diagnostics.uploads_enabled`, storage-validated) + client settings; crash prompt asks *"Send a diagnostic report to your server?"* (Send / Don't send / Always send). Consent and pending reports are **bound to a specific server + account** so multi-server users can't leak one server's data to another.
- **Bundle-on-event, not streaming**: a compressed tar.gz archive (structured JSONL logs, crash artifacts, hardware snapshot) captured at failure or via a user-driven "start capture → reproduce → stop & send" flow.
- **Hybrid storage**: queryable manifest row in Postgres, payload blob in the existing S3-compatible private bucket; retention + orphan reconciliation via a taskmanager cleanup task. Raw client log lines never enter `operational_logs`.
[...]

PR #445, opened 2026-07-21
feat(diagnostics): client diagnostics server — ingest pipeline, R2 storage, admin API + viewer

Quick104:
## Problem

Clients (Android, Apple) have no way to deliver crash reports or opt-in debug logs to the user's server, so playback/browsing issues reported from the field are undiagnosable. The design for this capability is in #430 (cross-repo spec, included here as the first commit).

## What this adds

Server side of the client diagnostics capability, per the spec:

- `POST /api/v1/diagnostics/reports` — authenticated multipart ingest (manifest + tar.gz bundle) with schema validation, server-instance binding, consent gating, per-user/day and byte quotas, and streaming bundle validation (entry allowlist, size/ratio limits, no trailing data).
- Reports land in `client_diagnostic_reports` (`receiving` → `ready`, with `failed` compensation on any post-insert error) and bundles in the private S3/R2 bucket.
- `GET /api/v1/diagnostics/status` capability endpoint (feature detection per v1 API rules; additive only).
- Admin API: list/get/download/delete reports, download via presigned URL.
- Admin web viewer for browsing and downloading reports.

## Fixes found by end-to-end verification on dev

The final three commits are bugs found by exercising the full pipeline against the dev deployment (Cloudflare R2 storage) rather than fakes:

- `fix(s3client)`: R2 rejects unsized `PutObject` streams with 411; streaming uploads now go through the SDK multipart manager (8 MiB sequential parts, bounded memory, sizes stay untrusted).
- `fix(diagnostics)`: pgx binds nil slices as SQL NULL, violating the `playback_session_ids` NOT NULL default; ingest also swallowed underlying insert/validation errors, which made every failure a generic 500.
- `fix(diagnostics)`: the bundle contract was unimplementable with standard tar writers — end-of-archive zero padding (emitted by GNU tar, Python tarfile, Commons Compress) was rejected, and `uncompressed_bytes` used a sum-of-payloads definition no client observes. Padding (zeros only, ≤64 KiB) is now accepted and `uncompressed_bytes` is defined as the full decompressed tar stream; both documented in the spec doc and contract.

## Verification

- Unit + contract tests for ingest, quotas, bundle validation (incl. new padding/trailing-data cases), s3client.
- Full E2E on dev: upload accepted end-to-end, DB row `ready`, failed attempts compensated to `failed`, admin list/detail correct, and the bundle downloaded via the admin presigned URL is byte-identical (SHA-256) to the uploaded artifact.

## Risks / follow-ups

- `go.mod` picks up `aws-sdk-go-v2/feature/s3/manager` (bumps the s3 service module). Full test suite passes apart from pre-existing environment-dependent failures (playback NVENC probes, jellycompat process locks) that reproduce on the unmodified base commit.
- Client-side capture/upload (silo-android, silo-apple) is follow-up work per the spec.
- Admin viewer screenshots can be added on request.

Part of the client diagnostics capability — spec: #430.

AI-use disclosure: implemented and verified with Claude Code.

https://claude.ai/code/session_012e3QjbPo96ed9Mn2qRiUkh

## Summary by CodeRabbit

* **New Features**
  * Added opt-in client diagnostics uploads (consent-gated) with schema-validated manifests and gzip diagnostic bundles covering crash, abnormal exit, device, and logline data.
  * Added `/api/v1/diagnostics` status and report upload endpoints.
  * Added admin UI and endpoints to list/filter/paginate reports, download only “ready” reports, and delete reports.
  * Added background cleanup for expired, stale, and orphaned diagnostic reports/bundles.
* **Bug Fixes**
  * Improved upload/download robustness with stricter bundle parsing, clearer validation, and quota-aware `Retry-After` behavior.
* **Documentation**
  * Added versioned design and schema documentation for the client diagnostics contract (v1).


Quick104:
All 11 review findings addressed in a6348b3d — thanks Codex + CodeRabbit, several were real catches:

- **Padded tar names**: now rejected instead of trimmed (`manifest.json ` can no longer masquerade as an allowlisted entry).
- **Byte quota**: `receiving` rows now reserve their expected size in `blob_bytes` and the quota sums `receiving`+`ready`, closing the concurrent/multi-node overshoot.
- **Crash metadata**: the contract now requires the `crash` object for crash/anr/native_crash/hang/abnormal_exit and rejects it on manual reports (all valid fixtures already complied; verified the Apple client sends it for abnormal exits).
- **Instance-ID seeding**: atomic via `SetIfAbsent` + re-read-winner, with a plain-`Set` fallback for stores without it.
- **Embedded manifest**: `ValidateBundle` now captures the first entry (bounded by the manifest size limit) and `Ingest` deep-compares it against the multipart manifest minus `archive` — a stored archive can no longer disagree with its DB row.
- **Delete ordering**: DB row first, then blob; blob-delete failures log bucket/key for the reaper instead of stranding a visible-but-undownloadable report.
- **PAX/GNU smuggling**: non-USTAR formats and extension records are rejected outright.
- **Migration**: CHECK constraints for the state/report_type/platform closed sets.
- **Cleanup**: per-report errors now log-and-continue (aggregated via `errors.Join`) instead of aborting the run; diagnostics got its own `diagnostics.cleanup_interval_minutes` key instead of borrowing opslog's, and the startup settings lookup is bounded by a timeout.
- **Docs**: fence language identifiers added.

Quick104:
Round-2 findings all addressed in 2d5d4980:

- **Cleanup interval**: parsed minutes capped (1 min–7 days) so absurd values can't overflow `int64` nanoseconds.
- **Admin download**: the web admin now requests `?proxy=1` and streams through the server — presigned URLs broke when the private bucket is only server-reachable, and the separate-window navigation had no way to fall back.
- **Settings reads**: real `store.Get` failures now propagate (missing/empty still defaults) — a DB blip can no longer report the feature as disabled or silently use default quotas.
- **Bundle content validation**: non-manifest entries are structurally validated while streaming — `device.json`/`crash/*.json` must be a single JSON object (token-streamed, bounded depth), `logs.jsonl`/`breadcrumbs.jsonl` must be NDJSON objects with a 64 KiB per-line cap; `crash/stack.txt` and `crash/tombstone.pb` stay opaque. Memory stays bounded; size/ratio sentinels keep precedence.
- **Upload read deadline**: `HandleUpload` extends the per-request read deadline to 10 minutes via `http.ResponseController` — the server-wide 30s `ReadTimeout` was cutting off slow mobile uploads mid-stream.

Quick104:
Round-3 findings addressed in 93b851fe:

- **Write deadline**: `HandleUpload` now extends the write deadline alongside the read deadline (the global 120s `WriteTimeout` could still cut off slow uploads after the read-side fix).
- **Child-profile attribution**: diagnostics attribution now rejects child profiles server-side via `NewProfileAttributionValidator` (profile must belong to the user *and* not be a child) — mirrors the client-side gating, so a nonconforming client can't record reports against a child profile.
- **Download test**: the hook test now captures the clicked anchor and asserts the blob href and `silo-diagnostics-<SHORTID>.tar.gz` filename.

Quick104:
Round-4 findings addressed in 1f9bd99:

- **Schema/validator alignment**: the canonical `manifest.schema.json` now carries the crash-object conditionals (required for crash/anr/native_crash/hang/abnormal_exit, forbidden for manual) and `prefixItems` requiring `manifest.json` as the first archive entry — matching what the Go validator and upload path enforce.
- **Byte vs character limits**: JSON Schema can't express UTF-8 byte lengths, so `maxLength` stays as the character sanity bound with a top-level description stating the server enforces byte lengths, plus per-field notes on the free-text fields where the divergence realistically bites.
- **Profile mismatch**: when both `X-Profile-Id` and `manifest.report.profile_id` are present and differ, ingest now rejects with 400 `profile_mismatch` before any row/blob is written (previously the header silently won). Client-side handling is being aligned in the Apple PR.

Quick104:
Round-5 review findings addressed in dee46f93:

- **service.go child profile**: FIXED — a supplied child-profile attribution is now rejected with a distinct `ErrChildProfileForbidden` (`403 child_profile_forbidden`) instead of being silently dropped like a not-found profile. A profile that simply isn't the user's still drops attribution and proceeds unattributed.
- **repo.go manifest projection**: FIXED — added a manifest-free projection (`reportListSelectSQL` / `scanReportSummary`) for admin list and retention/stale cleanup queries; the full manifest projection is reserved for `GetByID`/`DeleteByID`, and `Report.Manifest` is now `omitempty`.
- **cleanup.go ordering**: FIXED — retention and stale loops now delete/mark the DB row before the blob; a blob-delete failure after the row is gone is logged with bucket/keys for orphan cleanup to reap rather than aborting the run (shared helper with the admin `DeleteReport` path).
- **admin.go settings cross-field**: FIXED — reject `max_bytes_per_user` below `max_bundle_bytes` and the reciprocal, so `/diagnostics/status` can't advertise a bundle size every upload would reject as `quota_exceeded`.
- **router.go demo guard**: FIXED — `POST /api/v1/diagnostics/reports` now routes through `DemoGuard`, and the reports prefix is blocked in demo mode; `GET /api/v1/diagnostics/status` stays available.

`go build ./...` and `go test ./internal/diagnostics/... ./internal/api/... -count=1` pass.

Quick104:
The four review comments posted here just before this PR merged are addressed in follow-up PR #450 (https://github.com/Silo-Server/silo-server/pull/450): the list-manifest dereference regression (server-side `app_build` projection + TS type split), lossless `UseNumber` manifest comparison, the multipart no-drain rejection, and a documented SKIP for the quota reservation (archiveMatches guarantees claimed==actual, so no stored report exceeds its reservation).

PR #450, opened 2026-07-21
fix(diagnostics): post-merge review follow-ups for #445

Quick104:
These changes address the four review comments posted on #445 at 2026-07-21T17:55Z — roughly 20 minutes before #445 was merged, so they never landed in that PR. This is the post-merge follow-up.

## Findings addressed

- **Stop dereferencing omitted list manifests (P2).** The admin list projection omits the manifest, but list rows dereferenced `report.manifest.report.app_build`, throwing as soon as any report rendered. Fixed server-side by projecting `COALESCE(manifest->'report'->>'app_build','')` into both the list and detail queries (a cheap JSONB text extraction — no manifest shipped in list responses) and scanning it into a new `Report.AppBuild`. On the web side the type is split into `DiagnosticReportSummary` (list; no manifest) and `DiagnosticReport extends …Summary { manifest }` (detail via `GetByID`), so list rows can no longer dereference the manifest at compile time. Rows and the detail pane read `report.app_build`; the detail pane still renders the full manifest.

- **Preserve JSON numbers when comparing manifests (P2).** `embeddedManifestMatches` decoded via `json.Unmarshal` into `map[string]any`, collapsing every number to `float64`; two manifests differing only in a large integer above 2^53 (e.g. `log_summary.lines`) rounded to the same float and compared equal, letting a tampered bundle through. Now decoded with `json.Decoder` + `UseNumber()` on both sides so numbers keep their exact decimal text, plus a re-asserted no-trailing-data check to match `json.Unmarshal` strictness.

- **Quota reservation from claimed bytes (P2) — SKIP with rationale.** Reserving the client-claimed `manifest.archive.bytes` is sound: a report only reaches `MarkReady` (which rewrites `blob_bytes` to the measured size) after `archiveMatches` confirms the claimed bytes/sha/entries exactly equal the streamed bundle, so no stored report can ever occupy more bytes than it reserved — under-claiming to exceed the storage quota is impossible, and a lying upload is rejected and compensated (blob deleted, row marked failed) before it counts. The residual cost of such rejected churn is bandwidth, bounded by the per-account single in-flight upload cap; `max_bytes_per_user` is a storage cap enforced oldest-first by retention (per the design doc), not a bandwidth limit. Documented at the reservation site; no code change beyond the comment.

- **Reject bad multipart parts without draining (P2).** `nextDiagnosticsPart` called `part.Close()` on a wrong-name/wrong-content-type part, which drains the unread body first — letting a max-size wrongly-named first part stream up to the bundle limit while holding the in-flight slot before returning 400. It now abandons the mismatched part and returns immediately, so malformed uploads fail promptly under load. Error semantics are unchanged (still 400 `invalid_bundle`).

## Tests / checks

- Go: `go build ./...` clean; `go test ./internal/diagnostics/... ./internal/api/... -count=1` all pass. New tests: `TestEmbeddedManifestMatchesPreservesLargeIntegers` and `TestDiagnosticsUploadWrongFirstPartNotDrained` (verified the latter fails when draining is restored — reads the full 1 MiB body — and passes with the fix).
- Web: `pnpm run lint` 0 errors, `npx tsc --noEmit` clean, `vitest diagnostics.test.ts` 2 passed, prettier + gofmt clean.

Part of the client diagnostics capability — spec: #430.

AI-use disclosure: authored by Claude Code (Claude Opus 4.8) under human direction; all build/test/lint checks above were run and pass.

https://claude.ai/code/session_012e3QjbPo96ed9Mn2qRiUkh

## Summary by CodeRabbit

- **New Features**
  - Diagnostic reports now include and display the application build (`app_build`) directly in both report lists and report details.
  - Report data is presented using distinct “summary” (list) vs “full detail” (view) shapes for more consistent UI behavior.

- **Bug Fixes**
  - Invalid diagnostic uploads are rejected without unnecessarily processing large request bodies.
  - Manifest JSON validation now preserves large integers correctly and rejects trailing non-whitespace/extra JSON to improve report integrity.

### DIFF
--- a/internal/api/handlers/diagnostics.go
+++ b/internal/api/handlers/diagnostics.go
@@ -220,12 +220,16 @@ func nextDiagnosticsPart(mr *multipart.Reader, expectedName, expectedContentType
 	if err != nil {
 		return nil, err
 	}
+	// Reject a mismatched part without calling part.Close(): Close drains the
+	// unread part body first, so a max-size wrongly-named/typed first part would
+	// stream up to the bundle limit — holding the per-user/global in-flight slot —
+	// before we return 400. Abandoning the part returns immediately; net/http then
+	// discards only a small bounded prefix of the request before closing the
+	// connection, so malformed uploads fail promptly under load.
 	if part.FormName() != expectedName {
-		_ = part.Close()
 		return nil, errDiagnosticsUnexpectedPart
 	}
 	if !diagnosticsContentTypeMatches(part.Header.Get("Content-Type"), expectedContentType) {
-		_ = part.Close()
 		return nil, errDiagnosticsUnexpectedPart
 	}
 	return part, nil
--- a/internal/api/handlers/diagnostics_test.go
+++ b/internal/api/handlers/diagnostics_test.go
@@ -85,6 +85,60 @@ func TestDiagnosticsUploadWrongPartOrder(t *testing.T) {
 	assertDiagnosticsRejectionLog(t, logs, "invalid_bundle", 42)
 }
 
+func TestDiagnosticsUploadWrongFirstPartNotDrained(t *testing.T) {
+	service := newFakeDiagnosticsService()
+	handler := NewDiagnosticsHandler(service)
+
+	// A large, wrongly-named first part must be rejected without draining its
+	// body: an invalid upload must not be allowed to stream its whole payload
+	// (holding the in-flight slot) before it gets a 400.
+	var body bytes.Buffer
+	writer := multipart.NewWriter(&body)
+	header := make(textproto.MIMEHeader)
+	header.Set("Content-Disposition", `form-data; name="bundle"`)
+	header.Set("Content-Type", "application/gzip")
+	w, err := writer.CreatePart(header)
+	if err != nil {
+		t.Fatalf("create multipart part: %v", err)
+	}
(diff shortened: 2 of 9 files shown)


Quick104:
Addressed review findings in 31e6f1ef:

- **service.go (trailing data after decoded manifest)** — Fixed. `Decoder.More()` only reports array/object iteration, so a stray closing delimiter like `{"log_summary":{}}}` slipped through where `json.Unmarshal` used to reject it. `decodeJSONObject` now requires the stream to reach `io.EOF` after decoding (a second `Decode` must return `io.EOF`), which rejects any trailing token — junk or an extra `}`/`]` — while still allowing trailing whitespace, and applies to both the received and embedded sides. Added `TestEmbeddedManifestMatchesRejectsTrailingData` covering `}`, `]`, trailing garbage, a second object, and trailing whitespace.
- **repo.go (cleanup paying for app_build JSONB extraction)** — Fixed. Split the projection: `reportListSelectSQL` keeps `COALESCE(manifest->'report'->>'app_build','')` for the admin list, and a new `reportCleanupSelectSQL` omits it so retention/stale batches never read the manifest JSONB just to delete a row. Added `scanReportCleanup` (via a shared `summaryScanTargets` helper) and switched `RetentionCandidates`/`StaleReceiving` to the cleanup projection.

`go build ./...` and `go test ./internal/diagnostics/... ./internal/api/... -count=1` both pass.

Issue #469, opened 2026-07-25
Requests stay queued forever after Radarr/Sonarr imports the download


Quick104:
## Summary

A fulfilled request never leaves `queued`. Once Radarr/Sonarr imports the download the item
leaves the arr queue, and nothing in the fulfillment path can tell "imported" apart from
"never grabbed" — so the target stays live forever and the request never completes, even
though the media is sitting in the library.

Found while investigating `/requests/movie/1228710` ("Star Wars: The Mandalorian and Grogu"),
which showed `queued` despite `movie-tmdb-1228710` being present and playable.

## Root cause

Two independent gaps that combine into a deadlock.

**1. The arr plugin can never report completion.**
`CheckMovieStatus` ([silo-plugins-requests-arr `internal/arr/radarr.go`]) only reads
`/api/v3/queue/details?movieId=`. An imported movie is gone from that queue, so
`EvaluateQueue(nil)` returns `{State: queued, ExternalStatus: "not_in_queue"}` and
`StatusFromQueueEvaluation` only ever emits `queued` or `downloading`. `StatusCompleted` is
declared in the plugin's `types.go` and produced by no code path. `movieResource` does not
even parse `hasFile`. Sonarr has the same shape.

**2. The host's presence shortcut is gated off precisely when it is needed.**
`reconcileRequest` (`internal/requests/service.go`) will complete a presence-confirmed request,
but only when `!hasLiveTargets` — deliberately, so a quality-agnostic TMDB presence hit cannot
orphan an in-flight download. Because gap 1 keeps the target live forever, the shortcut stays
disabled forever.

Result: reconcile runs on schedule and achieves nothing. Real `task_executions` rows before
the fix, repeating every 30 minutes:

```
{"errors": 0, "failed": 0, "checked": 49, "skipped": 0, "completed": 0, "submitted": 0, "downloading": 0}
```

## Scale (production deployment)

49 requests stuck `active`/`queued`; **24 of them already in the library** by TMDB id:

```
 status | total | in_library
--------+-------+------------
 queued |    49 |         24
```

Target rows show the staleness — `external_status` still holds the value written at submission
time, because reconcile only writes on a status *change*:

```
 target_status | external_status | count
---------------+-----------------+-------
 queued        | queued          |    45
 queued        | not_in_queue    |     4
 failed        |                 |     1
```

## Fix

Two PRs, one per gap:

- Plugin: consult the movie/series record when the queue is empty (`hasFile` for Radarr, all
[...]

PR #470, opened 2026-07-25
fix(requests): retire stalled targets when presence confirms the media

Quick104:
## Problem

`reconcileRequest` will complete a request whose media presence-checks as available, but only
when `!hasLiveTargets`. That gate is deliberate and correct: the presence check is
quality-agnostic (TMDB id only), so completing a request with a live target could orphan a
2160p download that is still in flight behind an already-present 1080p copy.

The gate assumes the router eventually drives every target to a terminal state. When it does
not, the request is pinned open permanently. That is not hypothetical — the arr plugin could
never report completion at all (Silo-Server/silo-server#469, fixed in
Silo-Community/silo-plugins-requests-arr#5), so on this deployment 49 requests sat `active`/`queued`,
24 of them already in the library, with reconcile logging `completed: 0` every 30 minutes.

The plugin fix addresses the cause. This PR is the backstop, so no future router bug can
silently pin requests open again.

Part of #469

## Approach

When presence confirms the media and live targets exist, retire targets that have sat in
`queued` past a 24h horizon without a single status transition, then let the existing target
aggregate (`aggregateStatus`) derive the request status rather than force-setting it.

Two deliberate constraints keep the original orphan protection intact:

- **`downloading` targets are never retired.** Those are exactly the in-flight downloads the
  gate exists to protect. Only `queued` is eligible.
- **The horizon reads real transitions.** `UpdateTargetStatus` is only called when the status
  actually changes, so `updated_at` means "last real transition", not "last poll". A target
  older than 24h has had ~48 reconcile passes to move and has not.

Retirements log at WARN with the target's identity and age — reaching this path means a router
is misbehaving, and that should be visible rather than silently healed. `hasLiveTargets` becomes
`liveTargets` so the caller can inspect them.

## Verification

```
$ go test ./internal/requests/...
ok  	github.com/Silo-Server/silo-server/internal/requests	0.016s
ok  	github.com/Silo-Server/silo-server/internal/requests/arrclient	0.016s

$ make verify-local-paths
scripts/check-local-path-leaks.sh          # exit 0

$ cd web && pnpm run lint && pnpm run format:check
✖ 158 problems (0 errors, 158 warnings)    # pre-existing warnings, 0 errors
```

`make lint` currently fails on `main` in this repo with a large body of pre-existing findings
(`misspell: 50`, `unused: 50`, `staticcheck: 49`, …), so a bare pass/fail says nothing. Instead I
diffed the findings for the touched files with and without this diff:

```
$ diff <lint before> <lint after>
< string `radarr` has 16 occurrences ...      > string `radarr` has 17 occurrences ...
< string `req-1` has 18 occurrences ...       > string `req-1` has 20 occurrences ...
< string `queued` has 4 occurrences ...       > string `queued` has 5 occurrences ...
```

No new findings — only occurrence counts on pre-existing `goconst` warnings in the test file.
No `sloglint` finding on the new log call.

New tests: a stalled queued target is retired and the request completes with
`external_status = presence_confirmed`; a target inside the horizon is left alone; a
`downloading` target stalled 30 days is never retired.

**Live validation.** Deployed after the plugin fix, so the two paths are distinguishable by
`external_status`:

```
 integration_kind |  external_status   |  status   | count
------------------+--------------------+-----------+-------
 radarr           | imported           | completed |    20     <- plugin fix
 radarr           | presence_confirmed | completed |    12     <- this PR
 sonarr           | presence_confirmed | completed |     2     <- this PR
```

14 WARN lines were emitted, one per retirement, e.g.:

```
level=WARN msg="requests: retired stalled target on presence" component=requests
request_id=129112453467340804 target_id=23 media_type=movie tmdb_id=1273221 quality=1080p
integration_kind=radarr external_id=33757 external_status=not_in_queue
target_updated_at=2026-07-23T15:28:17.065Z
```

Requests in the library but still `active` went 24 → 0, and the originally reported request
(`tmdb 1228710`) now reads `completed`. Subsequent passes are stable — no churn, no
re-completion.

## Risks

- The 24h horizon is a package constant, not a setting. Deliberate: adding a tunable means a
  migration plus admin surface for a backstop that should rarely fire. Easy to promote later.
- A request whose media is present at one quality while a *queued* (not downloading) target for
  another quality is genuinely still waiting on an indexer for >24h will be completed early. The
  alternative — leaving requests pinned open forever — is worse, and `downloading` targets are
  exempt, but this is the trade-off worth a maintainer's eye.
- One observed request now reads `queued`/`outcome=failed` with a retired 1080p target and a
  pre-existing failed 2160p target. That is `aggregateStatus`'s documented completed+failed
  behaviour (surface as failed so Retry can re-submit), not a regression from this change.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Involvement: AI-assisted (investigation, patch, tests), human-directed
- Adversarial review: the main risk I chased was whether retiring a target could make
  `submitApprovedRequest` re-submit a duplicate for the same quality. It cannot — that path
  builds a `healthy` map and skips any quality with a non-failed target, and a retired target is
  `completed`. I also rejected an earlier design that keyed the horizon off request `updated_at`:
  it would have broken the moment reconcile started refreshing rows on unchanged polls, whereas
  target `updated_at` only advances on a real transition.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request completion handling when media presence confirms availability.
  * Queued/stalled targets older than the 24-hour horizon are now retired and marked completed (with presence-confirmed external status).
  * Targets currently downloading are preserved and not prematurely completed.
  * Requests with multiple targets now complete only the stalled queued portion while any active downloading target remains in progress.

### DIFF
--- a/internal/requests/service.go
+++ b/internal/requests/service.go
@@ -1557,17 +1557,24 @@ func (s *Service) reconcileRequest(ctx context.Context, req Request, fc *fulfill
 		// orphan in-progress downloads. Only take the shortcut for legacy/no-live
 		// -target requests; otherwise let per-target reconcile + aggregate drive
 		// completion.
-		hasLiveTargets, err := s.hasLiveTargets(ctx, req.ID)
+		live, err := s.liveTargets(ctx, req.ID)
 		if err != nil {
 			return reconcileUnchanged, err
 		}
-		if !hasLiveTargets {
+		if len(live) == 0 {
 			if req.Status == StatusCompleted {
 				return reconcileUnchanged, nil
 			}
 			if _, err := s.store.SetStatus(ctx, req.ID, StatusCompleted, Viewer{}); err != nil {
 				return reconcileUnchanged, err
 			}
+			return reconcileCompleted, nil
+		}
+		updated, retired, err := s.retireStalledTargets(ctx, req, live)
+		if err != nil {
+			return reconcileUnchanged, err
+		}
+		if retired && updated != nil && updated.Status == StatusCompleted {
 			return reconcileCompleted, nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


Quick104:
Addressed both CodeRabbit nitpicks in 30aafecf:

- **`presence_confirmed` constant** — extracted as an exported `ExternalStatusPresenceConfirmed`, and used in the tests. Exported rather than package-private because clients have a real reason to tell "the arr reported completion" apart from "the backstop found the media itself".
- **Mixed-target test** — added `TestReconcileRequestsRetiresStalledQueuedWhileDownloadingStaysOpen`, covering the `retired == true && updated.Status != StatusCompleted` fallthrough. A stalled 1080p target is retired via presence while a live 2160p download keeps the request open at `downloading`. Good catch: that was the one branch the original three tests only covered implicitly.

Codex had no findings.

```
$ go test ./internal/requests/...
ok  	github.com/Silo-Server/silo-server/internal/requests	0.017s
ok  	github.com/Silo-Server/silo-server/internal/requests/arrclient	0.016s
```

Lint delta on the touched files is unchanged — still no new findings, only occurrence counts on pre-existing `goconst` warnings.

PR #471, opened 2026-07-25
feat(scanner): detect deleted media files without a full scan

Quick104:
## Summary

- Add a `verify_file_presence` background task (15 min interval by default) that stats every file the catalog believes is on disk per enabled library and marks vanished ones `missing_since`, closing the gap between a deletion and the next full scan.
- Add `Scanner.VerifyPresence`: bounded-concurrency stat sweep (16 workers), skips files under unreachable/suspect-empty roots, and reconciles library membership + orphaned image dirs when anything is marked.
- Only `os.ErrNotExist` counts as removal — permission errors, timeouts, and other faults leave rows untouched.
- Bulk-removal backstop: if ≥50% of checked files vanish (and ≥20 files), the sweep aborts for that folder and defers to a full scan instead of gutting the catalog on a dropped mount.
- Add repo helpers `ListPresentByFolder` and `MarkMissingByIDs` (single statement, does not extend the removal grace for already-missing rows).
- Detail service now drops vanished files from playback versions inline, budgeted at 250ms / 32 files per request, and persists the finding via a new optional `MissingFileMarker`, so users stop getting dead play buttons.
- Wire the task and the missing-file marker in `cmd/silo/main.go`; the sweep does not empty trash — that stays the full scan's job.

## Testing

- `internal/scanner/presence_sweep_test.go`: stat sweep reports only absent files, ignores EACCES (skipped as root), handles empty input, errors on cancelled context, and asserts the exact abort-threshold boundaries.
- `internal/catalog/detail_vanished_test.go`: vanished versions are hidden and marked, everything kept when nothing vanished, still hidden when `MarkMissing` fails, works with no marker wired, respects the per-request check limit, and `preparePlaybackFiles` filters end to end.
- `make lint` — Not run.
- Manual verification against a live library (delete a file, wait for the sweep, confirm it disappears from the detail view) — Not run.

## Summary by CodeRabbit

- **New Features**
  - Added automatic verification of cataloged media files to detect files removed from disk.
  - Missing files are hidden from playback and compatibility API results.
  - Added periodic library checks with progress reporting and summary statistics.
  - Added safeguards to avoid marking files missing during uncertain or unusually large removals.
  - Library memberships and orphaned image data are reconciled when applicable.

- **Bug Fixes**
  - Prevents unavailable media versions from appearing as playable content.
  - Preserves files when filesystem checks encounter permission or transient errors.

### DIFF
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -1967,6 +1967,9 @@ func main() {
 		if deps.FolderRepo != nil && deps.LibraryScanQueue != nil {
 			taskMgr.Register(tasks.NewScanLibrariesTask(deps.FolderRepo, deps.LibraryScanQueue, deps.EventBus))
 		}
+		if deps.FolderRepo != nil && deps.Scanner != nil {
+			taskMgr.Register(tasks.NewVerifyFilePresenceTask(deps.FolderRepo, deps.Scanner, 0))
+		}
 		taskMgr.Register(tasks.NewCleanupOrphanedMediaItemsTask(catalog.NewOrphanedProvisionalCleaner(deps.DB)))
 		taskMgr.Register(tasks.NewBackfillMediaItemAliasesTask(catalog.NewItemAliasRepository(deps.DB)))
 		if deps.S3Public != nil {
@@ -2206,6 +2209,9 @@ func main() {
 		absDetailSvc := catalog.NewDetailService(absItemRepo, absEpisodeRepo, absSeasonRepo, absPersonRepo, absFileFetcher)
 		if deps.ImageResolver != nil {
 			absDetailSvc.SetImageResolver(deps.ImageResolver)
+		}
+		if deps.FileRepo != nil {
+			absDetailSvc.SetMissingFileMarker(deps.FileRepo)
 		}
 		var absScopeResolver scopeResolver
 		if policySystem != nil {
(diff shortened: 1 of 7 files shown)


Quick104:
Closing in favour of #472.

Investigating on the dev server showed this was solving the wrong problem, in a way that would have made the real one worse.

**The real bug:** files were being marked missing *while sitting on disk*. 1,247 rows were flagged missing and 191 of them still existed — a 15% false-positive rate. Comparing `ctime` against `missing_since`, 190 of those 191 had been on disk for 12–22 hours before being flagged. The cause is that dead-root protection guards deletion but not marking, so an unreachable CephFS mount hides its entire catalog. Since every catalog read filters `missing_since IS NULL`, that presents to users as "this file isn't available anymore" for perfectly intact media — the exact symptom that prompted this work.

**Why the sweep was wrong:** this deployment already runs a CephFS autoscan source that polls every 60s, completes in 11–16ms, and has resolved 1.37M changes into 924k scans since June. A 15-minute stat sweep duplicated that at far higher cost — measured at ~511k directory stats per pass against 1.75M files — and every one of those stats is another chance to catch a transient ENOENT and hide a good file.

#472 fixes the marking guard instead: no new tasks, no new tables, nothing on a timer. Detection stays with autoscan.

PR #473, opened 2026-07-25
feat(settings): add the canonical contract manifest and validator

Quick104:
First implementation step for the settings contract (#477). Adds the artifact everything else depends on. **Nothing reads it yet** — no routes, no storage, no behavior change.

## What changed

**`contracts/settings/v1/`** — the artifact, at a stable path because clients vendor it and generate bindings from it
- `manifest.json` — **38 definitions** (35 remote, 3 contract-known `client_local`)
- `manifest.schema.json` — the contract format
- `schemas/` — object value schemas for subtitle appearance, theme variable overrides, library page state
- `embed.go` — a package containing nothing but the embed directive, because `go:embed` cannot reach outside its own directory

**`internal/settingscontract/`** — loading, structural validation, value validation, RFC 8785 canonicalization, ETag, and a public projection that strips maintainer notes

## What the inventory turned up

The manifest covers every key the legacy registry accepts, every key the extension bag was silently accepting, every unregistered device key Android writes, and the profile columns that become settings. Each non-obvious case is recorded in a `notes` field:

| Finding | Disposition |
|---|---|
| `ui_theme`, `ui_text_scale`, `ui_text_weight`, `ui_high_contrast`, `ui_custom_theme_vars`, `ui_custom_css` reached the server **only** via the unknown-key extension bag — untyped and unvalidated, with `ui_custom_theme_vars` storing arbitrary JSON | Registered, typed, renamed to the dotted convention, moved to profile scope |
| `player.match_frame_rate` and `player.sleep_timer_default_minutes` are written by Android against a server that registers neither, so **every write and every reset returns 400 today** | Registered |
| `player.next_up_prompt_seconds` is Android's alias for `playback.next_up_prompt_seconds` | Not a definition; pinned as a migration alias in the test matrix |
| Android clamps playback speed to `4.0`, the server to `3.0` | Contract maximum is `3.0` |
| `subtitle_appearance` is the only key with no domain prefix | Renamed to `playback.subtitle_appearance` |

## Validation is stricter than the schema

The JSON Schema checks shape. The Go layer checks what actually causes bugs:

- A resolution order must end in `default` and may only resolve scopes the definition allows
- **Every writable scope must actually be read** — a setting cannot accept writes at a scope it will never honor, which is the "I changed it and nothing happened" failure mode
- Defaults validate against their own value schema, so a default outside its own range or enum fails at load
- Revision tags can never exceed the manifest revision, which is what makes revision-aware client filtering trustworthy
- `ceiling`/`floor` policy constraints are rejected on unordered types, where capping would silently do nothing. This forced `playback.preferred_quality`'s enum to be ordered ascending so its `max_playback_quality` ceiling has a defined direction

`ValidateValue` is the single validation path, so the mutation endpoint, the migration, and the manifest's own default checks cannot diverge later. Numbers decode through `json.Number`, so an integer setting rejects `30.5` instead of truncating, and object values validate against their referenced schema rather than accepting any JSON the way `validateJSONSetting` does today.

## Verification

- **124 tests across 16 cases pass** — the inventory matrix, per-definition defaults and resolution orders, revision filtering in both directions, canonical-byte stability and idempotence, notes stripping, and a table of eleven malformed manifests that must be rejected
- `golangci-lint run ./internal/settingscontract/... ./contracts/...` → **0 issues**
- `make verify-local-paths` passes; `go build ./...` clean
- Two failures in `internal/api/handlers` reproduce unchanged on `main` and are unrelated (one needs ffmpeg, one is a temp-dir cleanup race)

## Notes for review

- Promotes `santhosh-tekuri/jsonschema/v6` from indirect to direct
- The three `client_local` definitions (`downloads.wifi_only`, `downloads.keep_watched`, `subtitle.matches_device`) are contract-owned because Apple and Android both implement them with shared semantics, but the value never leaves the device. Validation enforces exactly one `client_local` scope and no policy constraint
- The renames are sanctioned by the design's non-goal of preserving accidental key names and are free during a breaking cutover, but they are judgment calls worth confirming
- `ui.theme` carries a device override, added in the second commit — light theme on a phone in daylight, dark on a TV. `ui.custom_theme_vars` and `ui.custom_css` stay profile-wide, so custom tokens tuned against a dark theme will sit on top of a light one if a device overrides. Widening those later is an additive revision bump

Part of #376. Design: #477.

## AI-use disclosure

Implemented with AI assistance (Claude Opus 5) under maintainer direction. The inventory was derived by reading the current server registry, web manifest, Android `PlaybackSettingsKeys`, and the profile/library/series preference stores; every drift claim is reproducible from those sources.

### DIFF
--- a/contracts/settings/v1/embed.go
+++ b/contracts/settings/v1/embed.go
@@ -1,0 +1,18 @@
+// Package settingsv1 embeds the canonical cross-platform user settings
+// contract so the server binary carries the exact bytes it was built from.
+//
+// This package deliberately contains nothing but the embed directive. The
+// contract files are the artifact clients vendor and generate bindings from, so
+// they live at this stable path rather than inside an internal package; the
+// embed has to sit beside them because go:embed cannot reach outside its own
+// directory.
+//
+// Loading, validation, and lookup live in internal/settingscontract.
+package settingsv1
+
+import "embed"
+
+// FS holds manifest.json, manifest.schema.json, and schemas/.
+//
+//go:embed manifest.json manifest.schema.json schemas
+var FS embed.FS
--- a/contracts/settings/v1/manifest.json
+++ b/contracts/settings/v1/manifest.json
@@ -1,0 +1,675 @@
+{
+  "api_version": 1,
+  "revision": 1,
+  "definitions": [
+    {
+      "key": "playback.audio_language",
+      "introduced_in": 1,
+      "persistence": "remote",
+      "allowed_scopes": ["profile", "profile_device", "profile_library", "profile_series"],
+      "resolution_order": [
+        "profile_series",
+        "profile_library",
+        "profile_device",
+        "profile",
+        "default"
+      ],
+      "value_schema": { "type": "language_tag", "nullable": true },
+      "default_value": null,
+      "category": "playback",
+      "label": "Preferred audio language",
+      "description": "Choose which spoken language Silo should prefer first.",
+      "recommended_control": "select",
+      "notes": "Migrates user_profiles.language as the roaming fallback. Existing user_device_settings values become real overrides, and the per-series value comes from AudioPreference.audio_language. AudioPreference.audio_track_index and track_signature stay specialized: they identify a concrete track, not a default."
+    },
+    {
+      "key": "playback.subtitle_language",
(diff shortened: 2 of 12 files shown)


Quick104:

PR #477, opened 2026-07-25
docs(settings): define the cross-platform settings contract

Quick104:
Design document only. No code, no runtime behavior change.

## What this adds

`docs/superpowers/specs/2026-07-10-cross-platform-user-settings-contract-design.md` — the decision-complete contract for how user settings work across the server, bundled web client, Apple clients, and Android clients. It turns the audit in #376 into something implementable.

## The decisions it makes

**Ownership.** Every production user-facing setting needs a server-owned manifest entry, even when the value is only ever stored on one client. The single exception is private `local.<client>.*` diagnostics and experimental knobs, bounded by five conditions.

**Types and scopes.** Native JSON values instead of strings. Five remote scopes (`account`, `profile`, `profile_device`, `profile_library`, `profile_series`) plus `client_local`, and each definition declares its own resolution order rather than inheriting a global precedence.

**Preferences versus restrictions.** `internal/policy` already resolves `max_playback_quality` and metadata-language limits over the same controls this contract resolves preferences for. Definitions declare `constrained_by`; the effective response reports the permitted value alongside the user's stored one; and a mutation that exceeds a restriction is **stored, not rejected** — a capped 4K preference should take effect the day the cap lifts, not be destroyed by it.

**Compatibility.** Widening a scope, adding an enum member, or widening a range is additive and revision-tagged. Narrowing anything requires a new key. `introduced_in` is a manifest revision and attaches to individual enum members and scopes, not just whole definitions, so a newer client never offers a choice an older server will reject.

**Rollout.** One coordinated breaking release — no compatibility shim, no projection, no client fallback. **After the cutover, no future setting requires coordination**; manifest revisions move independently per repository. No settings version check is added to the authenticated middleware and nothing returns `426`: deleting the old routes already produces the break, and a middleware gate would be more code in four repos for the same outcome, permanently coupling every endpoint to one subsystem's versioning.

**Scope placement.** Appearance and date/time move from account to profile scope. Account scope was an artifact of pre-profile `user_settings` storage; leaving it there means one household shares a theme and text size, and any non-child profile can restyle everyone else.

**Read path.** Batched context resolution, index requirements, a session-snapshot rule, and a no-regression benchmark gating the storage consolidation — `profile_series` resolution is per-item, so a season view would otherwise issue one request per episode.

## Verified against

The current server, Apple, and Android implementations at the baselines in #376, re-checked against `internal/api/handlers/settings.go`, `web/src/lib/settingsManifest.ts`, `internal/policy`, `internal/events`, `internal/jellycompat/handlers_displayprefs.go`, `internal/userdb`, `internal/userstore/pgstore`, and `docs/architecture/v1-scope.md`.

Two things that check turned up and that shape the design: the unknown-key extension bag is real (`keyUsesUserScope` accepts any unregistered key), and **v1 scope reads NOT LOCKED** — so removing the legacy surface needs no amendment if it lands before lock.

`make verify-local-paths` passes.

## Implementation status

| Step | PR |
|---|---|
| Canonical manifest and validator | #473 |
| Web appearance cache ownership | (this repo, see linked PR) |
| Android contract drift | Silo-Server/silo-android#101 |
| Apple audio language scope | Silo-Server/silo-apple#103 |

Supersedes #377, which was closed when its branch was renamed.

Related to #376. Does not close the implementation issue.

## AI-use disclosure

Drafted and revised with AI assistance (Claude Opus 5) under maintainer direction. Every claim about current behavior was verified against the cited source files.

### DIFF
--- a/docs/superpowers/specs/2026-07-10-cross-platform-user-settings-contract-design.md
+++ b/docs/superpowers/specs/2026-07-10-cross-platform-user-settings-contract-design.md
@@ -1,0 +1,782 @@
+# Cross-platform user settings contract
+
+**Date:** 2026-07-10
+
+**Status:** Draft — coordinated breaking-release design for issue #376
+
+**Scope:** `silo-server`, `silo-apple`, `silo-android`, and the Silo web client
+
+**Tracking:** https://github.com/Silo-Server/silo-server/issues/376
+
+> Commands and paths in this document are repository-relative; assume the relevant repository root
+> is the cwd. Cross-repository references are prefixed with the repository name.
+
+## Decision
+
+The server repository owns the canonical contract for every **production, user-facing setting**.
+That is true even when the value is intentionally stored only on one client. A client PR must not
+invent a production setting key, type, default, range, or scope independently.
+
+There is one narrow exception: a client may add a private implementation, diagnostics, or
+experimental knob without a server PR when all of the following are true:
+
+1. Its key is in `local.<client>.<domain>.<name>` (for example,
+   `local.apple.player.decoder_logging`).
+2. It is not shown as a normal production setting.
+3. It is never sent to any Silo API.
+4. It is not expected to roam, survive reinstall, appear in admin UI, or have shared semantics with
+   another client.
+5. Promoting it to a production feature requires adding it to the shared contract first.
+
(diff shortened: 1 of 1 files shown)


Quick104:

PR #478, opened 2026-07-25
fix(settings): stop web appearance caches leaking between accounts

Quick104:
Signing into a second account on the same browser showed you the *first* account's theme, text size, contrast, and custom CSS until you saved your own. Verified finding from #376.

Incognito is not the leaking case — it starts empty. The bug is two accounts in one regular browser profile.

## What changed

**`web/src/utils/storage.ts`** — new `createOwnedCache(ownerKey, memberKeys)` returning `isTrusted` / `get` / `set` / `clear`, plus three instances and two new owner sidecar keys.

| Behavior | Why |
|---|---|
| Null owner (bootstrapping / signed out) is trusted | The warm start still paints instead of flashing defaults |
| Unstamped legacy cache is **not** trusted once an account is known | This is the leak |
| `get` returns null when untrusted | Callers fall through to the server value |
| `clear` drops every member key and re-stamps for the new owner | A later single save can't re-trust the rest of the old account's state |

**`web/src/hooks/useTheme.tsx`** and **`useCustomTheme.ts`** — derive an owner, read trust once per render before any effect re-stamps, gate every `apiValue ?? localValue` fallback on it, and clear + reset to defaults when the cache belongs to someone else. That also restores the admin default theme for the new account, which a leftover `silo-theme` key had been suppressing.

**`web/src/hooks/useDateTimeFormat.tsx`** — refactored onto the shared mechanism. Behavior unchanged; it already did this correctly and was the pattern the other two were missing. Reuses its existing owner key so nobody loses their warm start.

**Three cache groups, three separate stamps** — deliberately not one shared stamp. These are written by hooks nested inside one another and React runs effects inner-first, so a shared stamp would let whichever hook resolved first vouch for the other's still-stale values.

## Verification

- `vitest run src/hooks src/utils` → 36 files, **148 tests pass**
- `tsc -b`, `format:check` → clean; `lint` → **0 errors** (160 warnings, same count as `main`)
- **The new tests were confirmed to catch the bug**: with the two hook fixes stashed, 7/7 fail with the leak signatures (`expected 'large' to be 'default'`, `expected 'cobalt-studio' to be null`, admin default not applied)
- `pnpm test` has 6 pre-existing failures (`SeasonContent`, `ServerStorageStep`, all `No QueryClient set`), confirmed pre-existing by stashing the whole change

## Known limits

- **One-time appearance reset for existing users** on first load after deploy: legacy untagged caches are untrusted until the settings query resolves and re-stamps. Same tradeoff `useDateTimeFormat` already shipped; stamping the existing cache with whoever signs in first would reintroduce the leak.
- **Owner is the user ID**, correct for today's user-scoped `/settings`. Household profiles share a `user_id`, so profile switching still shares the cache. #477 moves appearance to profile scope — that is a one-line change in `appearanceCacheOwner()`.
- A hard reload can still flash the previous account's appearance for under a second while auth bootstraps. Inherent to the warm start.
- Not exercised in a real browser with two live accounts; covered by rendering tests with mocked auth/settings.

Part of #376. Design: #477. Supersedes #476, closed when its branch was renamed.

## AI-use disclosure

Implemented with AI assistance (Claude Opus 5) under maintainer direction; tests, tsc, lint, and format were re-run independently, and the ownership logic was read against the already-shipped `useDateTimeFormat` pattern.

### DIFF
--- a/web/src/hooks/appearanceCacheOwnership.test.tsx
+++ b/web/src/hooks/appearanceCacheOwnership.test.tsx
@@ -1,0 +1,180 @@
+import { render } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import { appearanceCache, customThemeCache, storage } from "@/utils/storage";
+import { DEFAULT_THEME } from "@/lib/themes";
+
+const mocks = vi.hoisted(() => ({
+  useOptionalAuth: vi.fn(),
+  useSettings: vi.fn(),
+  useBranding: vi.fn(),
+  mutate: vi.fn(),
+}));
+
+vi.mock("@/hooks/useAuth", () => ({
+  useOptionalAuth: () => mocks.useOptionalAuth(),
+}));
+
+vi.mock("@/hooks/queries/settings", () => ({
+  useSettings: (options?: { enabled?: boolean }) => mocks.useSettings(options),
+  useSetSetting: () => ({ mutate: mocks.mutate }),
+}));
+
+vi.mock("@/hooks/useBranding", () => ({
+  useBranding: () => mocks.useBranding(),
+}));
+
+import { ThemeProvider, useTheme } from "./useTheme";
+import { useCustomTheme } from "./useCustomTheme";
+
+const KEYS = storage.KEYS;
+
+interface Captured {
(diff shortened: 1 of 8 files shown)


Quick104:

PR #481, opened 2026-07-25
feat(abs): add ebook compatibility layer

Pukabyte:
## Summary

- expose Silo ebook libraries through the Audiobookshelf compatibility API
- add ebook reader files, primary selection, delivery, and progress lifecycle
- populate ebook home shelves, author and series details, and book filter metadata
- keep audiobook-only playback, RSS, collection, playlist, and bookmark surfaces unambiguous

## Maintainer remediation

The original contribution was reviewed defect-first and then hardened before deployment. The remediation:

- preserves profile-scoped access and requires effective metadata-curation authority for server-global primary-file changes
- implements official-client ebook progress fields, atomic explicit unfinish, monotonic stale autosaves, and correct reset errors
- uses real ebook media-file IDs, safe authenticated GET/HEAD delivery, MIME/disposition hardening, and strict audio/RSS boundaries
- hydrates ebook author/series details and pushes publisher filtering into bounded, PostgreSQL-safe queries
- handles nil stores and operational access failures without false success or response-detail leakage
- preserves EPUB-first reader files when the batched primary-selection lookup is temporarily unavailable
- preserves upstream Audiobookshelf's intentional behavior where selecting the current primary toggles it back to supplementary

The branch is merged with `main` at `3d01fda434d3850a3e1ac523b240afbcce33c4c8`. Published head: `263cd71608508ba49187f211390c45f6f9c3b44f`.

## Second-pass review remediation (`263cd716`)

A further multi-angle review of `b109e9fa` produced 29 candidates; 10 were refuted by adversarial verification (among them the global `nosniff` header, the personalized-shelf skeleton, and the empty `PATCH …/ebook/{fileid}/status` body, which all match upstream Audiobookshelf). The 19 that survived are fixed in one commit:

- **Un-finish semantics.** `isFinished:false` compared the incoming progress against the finished threshold, so `{isFinished:false}` on an unfinished book wiped its position. The reset is now evaluated in SQL against the stored row, matching upstream's `!payload.isFinished && this.isFinished`.
- **Atomic progress writes.** The read-modify-write around ebook progress is gone; the store merges a partial patch inside one `UPDATE … RETURNING` (insert only on first write), so concurrent devices cannot clobber each other and the hot path is a single round trip.
- **Un-hide** deletes the history watermark rather than rewriting `updated_at`, so Continue Reading order and `lastUpdate` stay truthful.
- **Item-type guard** on the ebook file and status endpoints (500 / 404 / 400 "item is not an ebook"), so a non-ebook item with an ebook-extension file can no longer be served or pinned.
- **Default library** (sentinel, `/me` `defaultLibraryId`) prefers an audiobook folder now that ebook folders are listed; smart collections exclude ebook libraries and ebook rows.
- **ABS metadata curation goes through the policy PDP**, so custom policy overrides and decision logging cover ABS clients; a missing decider fails closed.
- **Native reader parity.** `defaultEbookFileID` / mark-as-read honor the ABS-curated primary file when it is accessible, falling back to the EPUB-first default otherwise. Surfacing the pin as a v1 field is a separate additive follow-up.
- **Efficiency.** Range requests use a projection-only item-type lookup (was a fully hydrated item fetch with people + series); the per-call `media_folders` type lookup is removed from 11 store methods; filterdata aggregates run concurrently, are `LIMIT`-bounded, and memoized per (library, access) for 60s; personalized shelves query concurrently and enrich ebook shelves once; the file-batch access gate projects ids only.
- **Structure.** New `internal/ebookformat` owns the format table, MIME map, preferred-file selection, and Content-Disposition for both `api/handlers` and `abs` (the ABS copy had dropped `.fb2.zip` and the container fallback). Capability interfaces are folded into `abs.MediaStore` / `abs.AccessResolver` (no runtime type assertions); `GetPrimaryEbookFileID` returns `EbookPrimarySelection`. The two primary-file migrations are collapsed into one with nullable `file_id` and `ON DELETE CASCADE` to `media_items` and `media_files` — deleting the pinned file falls back to the EPUB-first default instead of reading as "all supplementary".

Second-pass validation on `263cd716`: `go build ./...`, `go vet ./...`, `golangci-lint run --new-from-merge-base=origin/main ./...` (0 issues), `make migrate-validate`, `make verify-local-paths`, `gofmt`, `git diff --check`, and `go test -count=1` over `internal/audiobooks/... internal/ebookformat/... internal/catalog/... internal/api/... internal/policy/... internal/sections/... internal/playback/... migrations/... cmd/...` (20 packages, all pass); `go test -race ./internal/audiobooks/...` clean. The merge SQL was additionally exercised against a local Postgres 18 with session-temp tables. No new behavioral sandbox run was made for this pass.

## Validation

- `go build ./...` — pass
- `go vet ./...` — pass
- `go test ./internal/audiobooks/... ./internal/playback ./migrations` — pass
- `go test ./internal/access ./internal/auth` — pass against the current inherit/override policy model
- `go test -race ./internal/audiobooks/...` — pass
- `golangci-lint v2.12.2 run --new-from-merge-base=origin/main ./...` — `0 issues`
- `make migrate-validate` — pass
- `make verify-settings-bindings` and `make verify-playback-fixtures` — current
- `make verify-local-paths` — pass
- web lint — 0 errors; existing warnings only
- web format check — pass
- web tests — 283 files / 1,992 tests pass; the subsequently merged featured-hero suite adds 25/25 passing tests
- `git diff --check origin/main...HEAD` — pass

The repository-wide Go suite passes every PR-owned package locally but, on macOS, stops on two existing `internal/jellycompat` process-identity tests. Both failures reproduce unchanged on the canonical `main` checkout; the PR does not modify those tests or their implementation. The Linux GitHub checks remain the merge gate.

## Behavioral evidence

An isolated DevBuilder sandbox was built from trusted remediation SHA `4b7dbc32775ec6fb08f363ce39641ef262cf18d6`. It passed real ABS client-shape and HTTP probes for restricted-profile scoping, bounded ebook batches, browse/search/filter/author/series metadata, progress finish/unfinish/reset/concurrency, primary-file authorization and persistence, authenticated GET/HEAD/Range/conditional delivery, MIME and disposition safety, ebook rejection from audio-only surfaces, audiobook direct play, restart persistence, readiness, and cleanup. The later current-`main` reconciliation passed the local current-base gates above.

Redacted visual evidence was captured locally from the isolated live sandbox and was not uploaded.

### AI Disclosure

- Tool(s): T3 Code / Codex
- Model(s): gpt-5.6-sol
- Involvement: AI-assisted maintainer review, remediation, conflict resolution, and validation
- Second pass: Claude Code (Claude Fable 5) — multi-angle review with adversarial verification, remediation, and validation
- Adversarial review: The defect-first review found profile-scope and authorization gaps, incomplete official-client progress semantics, identifier ambiguity, unsafe delivery/error boundaries, unbounded or invalid PostgreSQL query behavior, missing ebook detail hydration, and nil-store failure paths. The validated findings were fixed and re-reviewed. One bot suggestion was rejected after comparison with upstream Audiobookshelf because the proposed no-op would break its intentional primary-to-supplementary toggle.

The contributor commits predate this maintainer pass; no contributor AI-use declaration was provided, so this disclosure covers the maintainer-assisted commits only.

## Summary by CodeRabbit

* **New Features**
  * Added Audiobookshelf ebook support, including reading, downloads, file selection, progress tracking, and mark-as-read flows.
  * Added ebook-aware libraries, search filters, metadata, shelves, collections, playlists, and item details.
  * Added bookmark listing and improved audiobook progress and continue-listening behavior.
  * Added publisher, genre, and language filtering with improved filter-data performance.
* **Bug Fixes**
  * Improved access enforcement, error responses, media-type validation, and safe download headers.
  * Preserved ebook metadata, file counts, sizes, languages, and preferred file selections in responses.


Quick104 (commented):
Reviewed pinned head `52c36109b576d33ba99d82d0ef7ead447b67ef6a` against live `main` (merge-base `73488d1bfaf12c2ac2bc8a24ef7e04dbddfe06a7`).

## Standards

- **High — authorization:** `internal/audiobooks/abs/extras_handlers.go:279` lets any authenticated profile with read access set or clear the server-global primary ebook row. The ABS token carries no account role/update permission, while `abs_ebook_primary_files` is keyed globally by `content_id`. Upstream ABS treats this as a catalog update and requires update permission; Silo needs an equivalent authorization boundary or a profile-scoped ownership model.
- **Medium — swallowed failure:** `internal/audiobooks/abs/extras_handlers.go:146` converts catalog lookup failures into `204 No Content`, so Reset Progress can report success without deleting anything.
- **Medium — duplicated persistence ownership:** `internal/audiobooks/abs_ebook_progress_store.go` reimplements the native `PGEbookReaderProgressStore` SQL and semantics. They already differ in how a finished row handles file/location updates.
- **Medium — success-shaped fallbacks:** `resolveLibrariesForItems` fabricates/type-guesses library identity after store failures, and ebook batch-enrichment failures return successful but unreadable entries.
- The PR links no issue/spec/capability despite adding a feature and schema. Its AI disclosure covers only remediation commit `f01ef817`; the contributor commits need the required disclosure or explicit “none.”

## Spec

The PR body is the only available specification.

- **P1 — completed ebooks cannot be marked unfinished:** `handleSetEbookProgress` handles only `isFinished:true`. A PATCH with `{"isFinished":false}` returns success but the handler and DB monotonic guards retain finished progress. This is the payload ABS clients use for “Mark as not finished.”
- **P1 — advertised publisher filters return zero books:** filter metadata exposes publishers, but `FilterKind`, `Filter.Matches`, and SQL pushdown have no publisher case. Selecting a published value performs an unbounded Go-filter fetch and rejects every item.
- **P2 — ebook author/series drill-down remains audio-only:** `GetAuthorByID` hardcodes `mi.type = 'audiobook'`; `GetSeriesByName` reads only `audiobook_series`. Ebook authors can appear in lists but open without their books, while ebook-only series return 404.
- **P2 — personalized Books is only partially aligned:** the implementation omits ABS’s Continue Series and Newest Authors shelves.

Validation passed:

- `go test -race ./internal/audiobooks/... ./migrations`
- `make migrate-validate`
- `make verify-local-paths`
- `git diff --check`

These are coverage gaps rather than existing test failures. GitHub CI is green, but the PR is currently conflicting with `main` and remains `REVIEW_REQUIRED`.

PR #483, opened 2026-07-26
feat(playback): measure baked-in letterbox so clients can keep subtitles on the picture

RXWatcher:
## Problem

A 2.39:1 image mastered inside a 1920x1080 frame — routine on WEB-DL sources — is a 16:9 video as far as the container, the probe, `SourceDescriptorV3.Width/Height`, and every client's video layer are all concerned. The black bars are *pixels*, not geometry anyone declares.

The consequence lands on subtitles: a client anchors cues to the bottom of the video rect, the video rect covers the full frame, and the cues render inside the baked-in bar instead of over the picture. There is no client-side fix, because nothing the client can see distinguishes this source from a genuine 16:9 one.

## Approach

Measure it server-side and publish it. `internal/playback/letterbox_detect.go` runs ffmpeg's `cropdetect` at four points spread across the runtime and **intersects** the results, keeping the smaller bar per edge — one frame cannot tell a bar from a fade to black, a night exterior, or titles over black, but a real baked-in bar is in every frame, so the minimum converges on the truth while transient dark frames drop out. Each edge is measured independently; symmetry is not assumed.

Results are published as fractions of frame height, so a client insets its overlay without knowing the frame's dimensions:

- `SourceDescriptorV3.letterbox_top_fraction` / `letterbox_bottom_fraction` (v3 plan)
- the same two fields on the v2 `playbackSessionResponse` — the Apple clients do not speak v3, and this is the response they decode

Both are additive and `omitempty`. Absent means "not measured", which is exactly what a failed measurement reports, so every existing client keeps doing what it already did.

## Guardrails

- **Never blocks a playback start.** Detection decodes frames. `LetterboxCache` returns what is already known and warms the measurement in the background, detached from the request context so a client disconnect cannot cancel it. The first play of a file plans without geometry — today's behavior — and later plays carry it.
- **The planner stays a pure decision function.** `PlannerInputV3.Letterbox` is supplied by the caller; nothing under `internal/playback` planning shells out.
- **Absurd measurements are refused rather than guessed at.** Under 2% per edge is rounding noise; over 25% would leave half a frame and would put cues mid-picture if acted on. Both collapse to "not measured". Clients clamp independently as well.
- **A zero result is cached.** "Measured, no bars" must not re-probe on every play of a normal 16:9 file.
- **Replaced files re-measure.** The cache key is path + file size, so an in-place upgrade or re-encode cannot keep serving the old file's bars.
- **The cache is bounded** at 4096 entries, evicted oldest-first.

## Known limits / follow-up

- **In-memory, so cold after a restart.** The natural home is a column on the media file row, written by the scanner/probe rather than lazily at playback. Deliberately out of scope here — that is a migration and a scanner change, and this PR is meant to be one concern. Worth an issue if the approach is accepted.
- Only sources with a known frame height and duration are measurable; anything else returns zero.

## Testing

- 11 unit tests in `internal/playback/letterbox_detect_test.go` covering the parse (including malformed and self-contradictory output), the intersection, the sanitising thresholds, sample placement, cache-key invalidation, eviction, and the nil/unmeasurable-input paths.
- `go test ./internal/playback/ ./internal/api/handlers/` green apart from three pre-existing NVENC-probe failures that also fail on a clean `main` (they probe for NVIDIA hardware and this is a Mac).
- `make lint` and `make verify-local-paths` clean for the touched files.
- **Verified end to end on real content.** A build of this branch ran on my server against the Android TV client: a 2.39:1 title measured 12.78% per edge — matching `(1080 − 1920/2.39) / 2 / 1080` — and the subtitles moved off the bar and onto the picture.

## Client side

- Android consumes the v3 fields (already on `main` in my fork).
- Apple: Silo-Server/silo-apple#107, which needs this to land first.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who directed the approach, reviewed the change, and ran the end-to-end verification described above.

## Summary by CodeRabbit

- **New Features**
  - Playback now detects baked-in black bars in video content.
  - Playback responses include optional top and bottom letterbox measurements.
  - Detected measurements are used to improve playback planning across alternate sources.
  - Measurements are cached and warmed in the background to reduce delays during playback.

- **Bug Fixes**
  - Improved handling of inconsistent or transient dark frames by validating measurements across multiple samples.

---

### Maintainer remediation AI disclosure
- Tool(s): Codex in T3 Code
- Model(s): `gpt-5.6-sol`
- Involvement: AI-assisted
- Adversarial review: Independent Standards and Spec reviews found unbounded background FFmpeg measurement work and acceptance of incomplete four-point sampling. The maintainer-assisted commit bounds work to two non-blocking measurements, preserves retry eligibility, and requires all four samples. Final independent Standards and Spec reviews found no new remediation defect; architecture-, API-design-, author-process-, and Fowler-heuristic observations were left report-only.

Verification evidence for the maintainer-assisted commit:

```text
go test -race ./internal/playback -run 'TestIntersectLetterboxSamplesRequiresEveryPoint|TestCacheBoundsConcurrentMeasurementsAndRetriesDroppedWarm' -count=20
PASS

golangci-lint run --new-from-rev e338f9ebd54176a769ae8b9e5743b88dcb8f045c
0 issues.

go vet ./internal/playback ./internal/api/handlers
PASS

go build ./cmd/silo
PASS

make verify-local-paths
PASS

git diff --check
PASS

cd web && pnpm run build && pnpm run format:check
PASS

cd web && pnpm run lint
PASS (158 existing warnings, 0 errors)

make lint
BASELINE FAILURE (306 findings unrelated to this PR)

Configured dev-backend deployment under the shared backend lock:
/api/v1/ready -> {"status":"ok"}
three distinct playback starts -> HTTP 201, HTTP 201, HTTP 201
first wave -> exactly two concurrent cropdetect measurements
overflow source -> retried after a slot freed; total measurements advanced to three
measurement durations -> 2126 ms, 1983 ms, 1525 ms
post-deploy panic/migration/readiness scan -> clean
```

---

### Maintainer merge-conflict resolution AI disclosure
- Tool(s): Codex in T3 Code
- Model(s): `gpt-5.6-sol`
- Involvement: AI-assisted
- Adversarial review: Independent Standards and Spec refreshes confirmed that the resolutions retain both sides of each overlap: `DVRPUStrippable` plus `Letterbox` in all planner calls, and `VideoCopyUnsafe` plus both letterbox fields in `SourceDescriptorV3`. No new Standards or Spec finding was introduced; the six prior report-only Standards observations remain unchanged.

Verification evidence for merge commits `4c3b636d` and `a95766e2`:

```text
git merge-tree --write-tree HEAD <live-main-sha>
exit 0

patch-scoped golangci-lint against the live base
0 issues.

focused letterbox, Dolby RPU, copy-safety, and planner race tests (-count=20)
PASS

focused protocol-v3 start/replan handler tests (-count=20)
PASS

go vet ./internal/playback ./internal/api/handlers
PASS

go build ./cmd/silo
PASS

make verify-local-paths
PASS

git diff --check
PASS

Configured dev-backend deployment under the shared backend lock:
/api/v1/ready -> {"status":"ok"}
two protocol-v3 playback starts -> HTTP 201/playable, HTTP 201/playable
letterbox measurements across both starts -> 1 (cached on the second start)
panic/migration/readiness errors -> 0
probe sessions cleaned -> true
```


RXWatcher:
Both nitpicks were fair. Addressed in e338f9eb.

**1. Background goroutine — fixed, and there was a worse bug underneath it than the one reported.**

The inflight key was deleted only on the success path. Any early return *or* panic left it behind, and `Warm` treats a present key as "a measurement is already running" — so that file would never be measured again for the life of the process, silently. It's released in a `defer` now, which also covers the panic case.

Added the `recover()` as suggested. The goroutine is detached, so nothing upstream would catch a panic there and a subtitle overlay is not worth taking the server down for.

**2. Shutdown context — skipped, with a reason.**

The suggested `context.WithoutCancel(ctx)` only helps if the caller supplies an application-lifecycle context, and the only context the handler has is the request's — which is precisely the one that must *not* cancel this work, since the measurement is for the *next* play. Plumbing a real lifecycle context down to the cache is worth doing, but it's a change to how the handler is constructed rather than to this file, and it belongs with the persistence follow-up already noted in the PR description (the measurement wants to live on the media file row, at which point the cache's lifetime question changes shape entirely).

Meanwhile the exposure is bounded: each `ffmpeg` sample is a `CommandContext` with a 20s timeout, and container teardown kills the process group regardless. Happy to do it here instead if you'd rather not carry it.

**3. Tautological test — replaced.**

Correct, that test asserted Go's assignment operator works. It now calls `PlanPlaybackV3` with `PlannerInputV3.Letterbox` set and asserts `result.Plan.Source` carries both fractions. I checked it isn't vacuous by deleting the two wiring lines in `plan_v3.go` and confirming it fails:

```
--- FAIL: TestPlannerPublishesLetterboxOnTheSource
    plan did not carry the measurement: {... LetterboxTopFraction:0 LetterboxBottomFraction:0}
```

Added a companion test that an unmeasured source omits the fields from the wire entirely, so a client can distinguish "not measured" from a real zero without a sentinel.

`go test ./internal/playback/` is green apart from four NVENC/hwaccel tests that fail identically on a clean `upstream/main` worktree here — they probe for NVIDIA hardware and this is a Mac.

RXWatcher:
Force-pushed to trim this branch back to the letterbox concern only. Two unrelated fixes had accumulated on it while I was testing against my own deployment; they are now separate PRs:

- Silo-Server/silo-server#484 — TMDB id-slug matching
- Silo-Server/silo-server#485 — Dolby Vision RPU probe

No letterbox commits changed; the previously reviewed content is intact.

zZebrahz:
Wondering how this will work for sources where the letterbox changes mid-video? There are multiple occasions where letterbox is used as a visual tool in the media itself. There are others, such as ones filmed in IMAX that have certain scenes that expand to fill the 16x9 frame while other scenes remain 2.39.

zZebrahz:
Also want to ensure that this only affects **Lower Third** placement and not **Bottom** placement

Quick104:
Closing after review because baked-in matte detection should be client-owned. Clients can inspect the actual decoded frames, handle changing aspect ratios, and apply platform-specific video and subtitle layout without adding a lazy process-local server probe or API contract. Android already follows that model; Apple work should do the same. Thanks for exploring and documenting the case.

PR #485, opened 2026-07-26
fix(playback): stop a broken Dolby Vision RPU from hanging playback

RXWatcher:
## Problem

A Dolby Vision Profile 7 source whose RPU ffmpeg cannot parse takes the whole playback session down. The `dovi_rpu` bitstream filter does not fail cleanly — it rejects every packet while ffmpeg keeps running:

```
[dovi_rpu] Failed to read unit 1 (type 39).
[dovi_rpu] Failed to read access unit from packet.
[vost#0:0/copy] Error applying bitstream filters to a packet:
    Invalid data found ... Invalid SEI message: payload_size too large (64 bytes)
```

One observed session emitted **376,316** stderr lines before the process was killed. No manifest was ever built, `master.m3u8` returned 503 after 9.7 seconds, and the client sat on a spinner. From the user's side the title simply never starts, with no error.

## Approach

Whether the strip works is a property of *the file*, not of the ffmpeg build, so `supportsDoviRPUFilter` cannot answer it. Asking ffmpeg to strip two seconds to the null muxer can, in about a second.

The probe reads **stderr**, not the exit code. ffmpeg treats a per-packet bitstream-filter error as non-fatal and exits `0` — which is precisely how a stream that could never start reached a live session in the first place. There is a test pinning that behaviour.

**The verdict is a planning input, not a transport patch.** It sits in `PlannerInputV3` alongside the transformation registries: the registries answer whether the executor carries the transformation, this answers whether the file does. A source that fails it is never planned onto a strip route at all.

That placement is what makes the fix hold. The plan is where `DynamicRange = "hdr10"`, `Claims.Video.HDR10` and the `dolby_vision_metadata_removed` / `hdr10_base_layer_preserved` validated claims are set, and it is what `session.RemuxDVMode` is derived from and persisted as. Suppressing the filter further downstream would leave all of those describing a stream the server is no longer producing, and would leave every path that re-derives the filter from the durable session free to put it straight back:

- `HandleStartTranscode` — the client-driven restart behind quality changes, seeks and burn-in — which also re-derives the filter from `file.PrimaryDVProfile() == 7` with no plan involved at all
- the remote audio-switch restart, from `updatedSession.RemuxDVMode`
- the progressive-remux transport, whose branch `plan_v3` evaluates *before* the HLS branches

### What the client gets

With no tone-map recipe in this tree, an HDR10-only client on an unstrippable P7 source has no route left. Rather than the generic `hdr_transcode_unsupported` (which sends the user looking for a missing encoder), the planner returns `dv_conversion_unsupported` — the terminal reason the `server_dv7_to_hdr10` spec already declares — with a message naming the real cause. A client that advertises its own DV→HDR10 or DV→DV8.1 transformation still gets that route, carrying a `dolby_vision_strip_unsupported_by_source` degradation warning so the server-side drop is not silent.

### The two paths that bypass the plan

Gated at the executor instead:

- **legacy/auto remux** (web player, jellycompat, pre-v3 stream tokens) neutralizes the profile exactly as it already does for a missing `dovi_rpu` filter, so it still starts and still plays.
- **the explicit v3 strip recipe** fails loudly. Reaching it with an unstrippable source means a session or stream token minted before the verdict was known; copying the base layer without the strip would emit dangling RPUs under an HDR10 claim, and attempting the strip is the hang. The request gets a definite 500 and the next start re-plans against the now-cached verdict.

## The probe

- **Tri-state verdict.** Only a stderr-confirmed rejection is cached. A timeout, a cancelled request, or an ffmpeg that will not start is inconclusive: the strip is kept and nothing is written. One client disconnecting mid-probe can no longer disable the strip for that file permanently.
- **Fails open.** "We did not find out" means strip, consistently — the same rule the nil probe follows, because most Profile 7 sources genuinely need it.
- **Singleflight**, so concurrent or retried starts of the same title share one ffmpeg.
- **6s timeout**, inside the budget a client waits on the manifest, and off the session lifecycle lock now that it runs at planning time. A source that cannot hand over two seconds of copied video in 6s cannot feed a live transcode either.
- **Keyed on size and mtime** as well as path and binary, so a file replaced in place with the same length is re-probed rather than inheriting the old verdict.
- **stderr capture bounded at 64 KiB.** The markers are in the first lines; a rejecting filter emits a pair per frame.
- **Bounded at 4096 entries**, evicted oldest-first.
- **Head-only coverage is stated, not asserted.** This catches a source that rejects from the first access unit — the observed failure. A source that parses cleanly at the head and breaks an hour in still fails mid-stream; that is the separate "bail out on repeated per-packet filter errors" problem below.

## Testing

Unit tests for the stderr detection (using the real log lines above), the tri-state caching rules, cache-key invalidation, eviction, singleflight, follower cancellation, and the nil/unprobeable paths. Planner tests that an unstrippable source is never planned onto the strip, that the terminal names the Dolby Vision cause, that a client-side transformation still wins, and that a non-DV source never pays for a probe. Executor tests that legacy/auto drops the filter and the explicit recipe refuses.

`go build ./...`, `go vet ./internal/...` and `-race` clean. `internal/playback` and `internal/api/handlers` green apart from two failures that reproduce identically on the parent commit: `TestServeDirectPlayChangedEntityRejectsOldIfRange` and `TestHandleReplanPlaybackV3SeekFailureRecoveryNeverChangesMediaVersion`.

**Caveat stated plainly:** the logic and the fallback paths are verified; that the specific file which triggered this now plays is not. Its RPU is malformed enough that the base layer may have problems of its own. The guaranteed improvement is a fast, definite, honestly-labelled outcome instead of a ten-second hang and a spinner.

## Follow-up not in scope

Even with this, a bitstream filter that fails at runtime for some other reason still spews unbounded stderr until the process is killed. Failing fast on repeated per-packet filter errors would be worth doing on its own.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who hit this on his own library and directed the investigation. Root cause came from the server logs of the failing session. The revision moving the decision into the planner was made by Claude Opus 5 in response to maintainer review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Dolby Vision Profile 7 playback by only applying DV RPU-to-HDR10 strip processing when the source supports it.
  * Prevented playback hangs and packet rejections caused by attempting strip conversions on incompatible sources.
  * Added clearer degradation/failure handling when server-based conversion isn’t usable, with safer fallbacks.

* **Tests**
  * Added coverage for Dolby Vision strip eligibility probing, caching behavior, concurrency/coalescing, cancellation handling, and remux start/stop behavior for unstrippable sources.

### DIFF
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -181,12 +181,8 @@ type PlaybackHandler struct {
 	// hwaccel, transcode dir). Wired to the live config in integrated mode
 	// so admin changes apply to newly started transcodes. Read it through
 	// playbackConfig(), which falls back to defaults when unset.
-	PlaybackConfig func() config.PlaybackConfig
-	FFmpegLogSink  playback.FFmpegLogSink
-	// dvRPUProbe answers whether a given Profile 7 source survives the RPU
-	// strip. Built lazily; nil keeps the previous strip-always behaviour.
-	dvRPUProbe        *playback.DVRPUProbe
-	dvRPUProbeOnce    sync.Once
+	PlaybackConfig    func() config.PlaybackConfig
+	FFmpegLogSink     playback.FFmpegLogSink
 	copySeekAnchor    copySeekAnchorResolver
 	realtimeCommandMu sync.Mutex
 	realtimeCommands  map[string]playbackCommandRecord
@@ -2613,8 +2609,12 @@ func (h *PlaybackHandler) HandleChangeAudioTrack(w http.ResponseWriter, r *http.
 				}
 				// A v3 DV strip remux carries its bitstream filter in the
 				// durable session route; dropping it here would hand the node
-				// a DV7 copy recipe that leaves dangling RPUs.
-				if updatedSession.RemuxDVMode == playback.RemuxDVStripToHDR10V3 && strings.EqualFold(nodeReq.TargetCodecVideo, "copy") {
+				// a DV7 copy recipe that leaves dangling RPUs. Sources that
+				// fail the RPU probe are the exception — for them the filter
+				// rejects every packet, so re-adding it on an audio switch
+				// would hang a session that was playing a moment ago.
+				if updatedSession.RemuxDVMode == playback.RemuxDVStripToHDR10V3 && strings.EqualFold(nodeReq.TargetCodecVideo, "copy") &&
+					playback.DVRPUStrippable(r.Context(), h.playbackConfig().FFmpegPath, file.FilePath) {
 					nodeReq.VideoBitstreamFilter = playback.DV7ToHDR10BitstreamFilter
 				}
 
@@ -3104,10 +3104,21 @@ func (h *PlaybackHandler) HandleStartTranscode(w http.ResponseWriter, r *http.Re
 	// for "copy" has no way to know the source needs the strip. Derived after
 	// the burn-in guard above so a copy request it rewrites to h264 never carries
 	// a copy-only bitstream filter.
+	//
+	// Gated on the per-source probe for the same reason the planner is: a
+	// source whose RPU ffmpeg cannot parse turns the filter into a per-packet
+	// rejection that never produces a segment, and this endpoint would
+	// otherwise put it back on every quality change, seek and burn-in restart
+	// of a session the planner had already routed away from it.
 	videoBitstreamFilter := ""
 	if strings.EqualFold(req.TargetCodecVideo, "copy") &&
 		(session.RemuxDVMode == playback.RemuxDVStripToHDR10V3 || file.PrimaryDVProfile() == 7) {
-		videoBitstreamFilter = playback.DV7ToHDR10BitstreamFilter
+		if playback.DVRPUStrippable(r.Context(), h.playbackConfig().FFmpegPath, file.FilePath) {
+			videoBitstreamFilter = playback.DV7ToHDR10BitstreamFilter
+		} else {
+			slog.WarnContext(r.Context(), "restart dropped the dolby vision rpu strip: source cannot be stripped",
+				"component", "api", "playback_session_id", req.SessionID, "file_id", file.ID)
+		}
 	}
 
 	// The request-level permission check above intentionally runs before the
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -181,8 +181,12 @@ type PlaybackHandler struct {
 	// hwaccel, transcode dir). Wired to the live config in integrated mode
 	// so admin changes apply to newly started transcodes. Read it through
 	// playbackConfig(), which falls back to defaults when unset.
-	PlaybackConfig    func() config.PlaybackConfig
-	FFmpegLogSink     playback.FFmpegLogSink
+	PlaybackConfig func() config.PlaybackConfig
+	FFmpegLogSink  playback.FFmpegLogSink
+	// dvRPUProbe answers whether a given Profile 7 source survives the RPU
+	// strip. Built lazily; nil keeps the previous strip-always behaviour.
+	dvRPUProbe        *playback.DVRPUProbe
+	dvRPUProbeOnce    sync.Once
 	copySeekAnchor    copySeekAnchorResolver
 	realtimeCommandMu sync.Mutex
 	realtimeCommands  map[string]playbackCommandRecord
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -425,7 +425,7 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
 	result := playback.PlanPlaybackV3(playback.PlannerInputV3{
 		Request: req, RequestedFile: requestedFile, EffectiveFile: effectiveFile,
 		AudioTrackIndex: audioIndex, Settings: settings,
-		Registry: h.transformationRegistryV3(r.Context()), HLSRegistry: h.lazyHLSPlanningRegistryV3(r.Context()), Now: time.Now(),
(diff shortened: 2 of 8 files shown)


Quick104:
Pushed `5538fb2d` to this branch addressing the review. Flagging up front that it **changes the approach**, not just the details — @RXWatcher you'll want to reset your local copy rather than rebase onto it.

### The core change: the decision moved into the planner

The original patch answered the right question in the wrong place. Suppressing the bitstream filter while building the transport left everything upstream of it still believing the strip was happening:

- the plan still carried `DynamicRange = "hdr10"`, `Claims.Video.HDR10 = true`, and the `dolby_vision_metadata_removed` / `hdr10_base_layer_preserved` validated claims — so the client was told it was getting clean HDR10 and handed P7 with dangling RPUs, which is the input this repo's own comments say stalls hardware decoders;
- `session.RemuxDVMode` was still persisted as `strip_to_hdr10`, and three separate paths re-derive the filter from it — `HandleStartTranscode` (quality change / seek / burn-in restart), the remote audio-switch restart, and the progressive-remux transport. Any of them puts the hanging filter back mid-session.

`HandleStartTranscode` was the worst of them: it derives the filter from `file.PrimaryDVProfile() == 7` with no plan involved at all, so it would have reproduced the hang on the first seek. And the progressive-remux branch in `plan_v3.go` is evaluated *before* the HLS branches, so for a good number of clients the broken file never reached the probe in the first place — the fix was wired into 2 of at least 4 places that apply the filter.

The verdict is now a `PlannerInputV3` field next to the transformation registries, which is the right analogy: the registries answer whether the *executor* carries the transformation, this answers whether the *file* does. Lazy and memoized like `HLSRegistry`, consulted only after every cheap gate has passed, so a non-DV start never shells out. A source that fails it is never planned onto a strip route, and the claims, `RemuxDVMode` and all three restart paths are consistent by construction.

### One consequence worth calling out

I'd assumed the fallback would be a transcode. It isn't — `hdrTranscodeUnavailableV3` rejects any non-SDR source and there's no tone-map recipe in the tree, so an HDR10-only client on an unstrippable P7 source has **no route at all**. That makes it a terminal.

I think that's the honest answer, and better than either alternative: the original behaviour was a 10s spinner, and copying without the strip is not reliably playable for dual-layer P7 (dangling RPUs on the base layer) while also being a silent lie about the format. So the planner returns `dv_conversion_unsupported` — the terminal reason the `server_dv7_to_hdr10` spec already declares, so no new client contract — with a message naming the actual cause instead of the generic `hdr_transcode_unsupported`, which would send someone looking for a missing encoder. Clients that advertise their own DV→HDR10 or DV→DV8.1 transformation are unaffected and still get that route, with a `dolby_vision_strip_unsupported_by_source` degradation warning explaining the server-side drop.

Happy to revisit if you'd rather it degrade to something playable, but I don't think there's a playable option for that combination today.

### Probe fixes

- **Cache poisoning.** `CanStrip` cached unconditionally, so a probe killed by a client disconnect, a missing ffmpeg, a transient I/O error or a timeout would permanently disable the strip for that file for everyone — directly contradicting the deliberate nil-probe "keep stripping" fallback 25 lines above. Now tri-state: only a stderr-confirmed rejection is cached; inconclusive keeps the strip and writes nothing.
- **Timeout.** 15s → 6s, i.e. inside the ~10s client budget rather than beyond it. Also no longer inside `LockSessionLifecycle`, since it runs at planning time.
- **Singleflight** added; concurrent and retried starts of a title share one run.
- **Cache key** now includes mtime and the ffmpeg binary, not just path and size — same-length in-place replacement was inheriting the old verdict.
- **Bounded stderr capture** (64 KiB). You were right that `-t 2` bounds the read in the healthy case, but a source killed at the timeout mid-flood isn't bounded by it, and it costs nothing to cap.
- **Comment corrections.** "a file that survives this survives the run" is now stated as what it actually is — this catches a source that rejects from the first access unit, not one that breaks an hour in. Also dropped the reference to "the letterbox cache", which doesn't exist in this repo; the real precedent is `supportsDoviRPUFilter` in `internal/playback/remux.go`, now named correctly.

### Executor-level gating

The two paths that bypass the plan entirely are handled where they are, rather than left to a plan they never consult: legacy/auto remux neutralizes the profile exactly as it already does for a missing `dovi_rpu` filter, and the explicit v3 strip recipe returns an error (surfacing as a 500 from `ServeRemuxWithDVMode`, not the empty 200 the old path would have produced) so the next start re-plans against the cached verdict.

### Verification

`go vet ./internal/...` and `-race` clean. Two test failures remain in `internal/playback` and `internal/api/handlers`; both reproduce identically on `73d4de7f` — `TestServeDirectPlayChangedEntityRejectsOldIfRange` (fails 3/3 there) and `TestHandleReplanPlaybackV3SeekFailureRecoveryNeverChangesMediaVersion`. The NVENC tests pass here; that box is Linux with an NVIDIA card, which is why they behaved differently on your Mac.

New tests cover the planner gating and terminal, the client-transformation route surviving, a non-DV source never paying for a probe, both executor paths, and the probe's tri-state caching, singleflight, follower cancellation and output bounding.

---

Quick104:
Pushed `3d1935ee`: the probe now runs under `context.WithoutCancel`.

The leader was probing on the caller's context, so if its client disconnected the run returned `dvRPUUnknown` — and `CanStrip` handed that fail-open straight to every follower already blocked on the shared `call`, even though their own requests were fine. One client walking away was enough to give a live session the strip the source cannot survive, i.e. the exact hang this PR exists to prevent, now reachable through an unrelated client. The work was discarded too, so the next start re-paid for the probe.

`dvRPUProbeTimeout` still bounds the run, so nothing is left dangling, and a verdict reached after the leader has gone is still correct and still worth caching. Follower behaviour is unchanged — a follower whose own request is cancelled still leaves immediately instead of waiting on someone else's probe.

`TestProbeSurvivesTheLeaderLeaving` covers it, and fails on the previous commit.

---

PR #486, opened 2026-07-26
feat(watch-together): make vote rooms actually vote

RXWatcher:
## Problem

`selection_mode` has been stored, normalized and published on every room snapshot since Watch Together landed, and **nothing has ever read it**. A `vote` room behaved exactly like a `host_pick` one: members could suggest and vote, the tally was recorded and broadcast to everyone, and then the host promoted whatever they liked regardless of it.

So the mode was a label on an unimplemented feature, and the vote counts on everyone's screen were decoration.

## What a vote room does now

The host **starts** the winner rather than choosing it. Promoting anything other than the leading suggestion is refused (`409 not_vote_winner`), because a host who can overrule the tally makes the mode host_pick with extra steps.

The winner is the head of the repository's existing ordering — `vote_count DESC, created_at ASC`. Most votes; ties to whoever suggested first. Deterministic, and re-suggesting a title cannot jump the queue.

A room where nobody has voted has **no** winner and says so (`409 no_votes_cast`), rather than quietly promoting the oldest suggestion as though a vote had happened.

`host_pick` rooms are untouched: the host still promotes freely.

## Both doors, not one

`PromoteSuggestion` is not the only way into a room's selection. `SelectItem` (`PUT /rooms/{room_id}/selection`) is host-only but was not gated by mode at all, so a host could set any title directly and bypass the vote entirely. Enforcing the tally on one path and not the other is the same as not enforcing it. A vote room now refuses a direct selection (`409 vote_room_selection`).

I got this wrong in the first commit here and caught it while working through the host's options — worth a look in review, since "is that every path into this state?" is exactly the question this change lives or dies on.

## Testing

The winner rule is split out from the database (`winnerFrom`) and unit tested: most-voted wins, ties go to the earlier suggestion, and an unvoted room yields `ErrNoVotesCast` rather than a default. `internal/watchtogether` and `internal/api/handlers` green; `make lint` clean for the touched files.

Running on my server. **Not yet exercised with real participants** — the rule has tests on both sides, but nobody has had two devices in a room voting against each other.

## Client side

Silo-Server/silo-android#106 carries the client half: hosting offers a vote room, the lobby marks the leader WINNING, and the host gets a single "Start winner" action. The client mirrors this winner rule exactly so its idea of who is leading cannot drift from the server's.

Note the clients also had to stop pre-selecting a title when creating a vote room — doing so set the room playing immediately and the vote never happened.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who specified the semantics (anyone suggests and votes, host starts the winner, host cannot override) and whose question about the host's options surfaced the `SelectItem` gap.

## Summary by CodeRabbit

* **New Features**
  * Added vote-winner selection for watch-together rooms, including promotion based on the current vote winner.

* **Bug Fixes**
  * When promoting before any votes are cast, the app now returns a clear **409 Conflict** error (“Nobody has voted yet”).
  * Deterministic tie-breaking ensures consistent winner selection when vote counts are equal.

* **Tests**
  * Added unit tests covering winner selection rules and the no-votes error case.

### DIFF
--- a/internal/api/handlers/watch_together.go
+++ b/internal/api/handlers/watch_together.go
@@ -568,6 +568,12 @@ func (h *WatchTogetherHandler) HandlePromoteSuggestion(w http.ResponseWriter, r 
 		switch {
 		case errors.Is(err, watchtogether.ErrRoomForbidden):
 			writeError(w, http.StatusForbidden, "forbidden", "Only the host can promote a suggestion")
+		case errors.Is(err, watchtogether.ErrNotVoteWinner):
+			writeError(w, http.StatusConflict, "not_vote_winner",
+				"This room votes for what plays; start the title that is winning")
+		case errors.Is(err, watchtogether.ErrNoVotesCast):
+			writeError(w, http.StatusConflict, "no_votes_cast",
+				"Nobody has voted yet")
 		case errors.Is(err, watchtogether.ErrSuggestionNotFound):
 			writeError(w, http.StatusNotFound, "not_found", "Suggestion not found")
 		case errors.Is(err, watchtogether.ErrRoomNotFound):
--- a/internal/api/handlers/watch_together.go
+++ b/internal/api/handlers/watch_together.go
@@ -367,9 +367,6 @@ func (h *WatchTogetherHandler) HandleSelectRoomItem(w http.ResponseWriter, r *ht
 	})
 	if err != nil {
 		switch {
-		case errors.Is(err, watchtogether.ErrVoteRoomSelection):
-			writeError(w, http.StatusConflict, "vote_room_selection",
-				"This room votes for what plays; start the winning suggestion instead")
 		case errors.Is(err, watchtogether.ErrRoomForbidden):
 			writeError(w, http.StatusForbidden, "forbidden", "Only the host can start or switch room playback")
 		case errors.Is(err, watchtogether.ErrRoomNotFound):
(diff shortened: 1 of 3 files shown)


RXWatcher:
**Semantics changed on review of the actual UX: the host can now override the vote.**

The original rule here was that the host *starts* the winner but cannot choose something else — the reasoning being that a host who can ignore the tally makes the vote decoration. On reflection that is too rigid for a living room: the person hosting is usually the one holding the remote, and a room can perfectly well vote and then have someone say "actually, let's watch the other one".

So the tally is advisory now. Concretely, on top of the two commits already here:

- `PromoteSuggestion` no longer refuses a non-winner
- `SelectItem` no longer refuses a direct selection in a vote room
- Both associated error paths are gone with them

I removed **both** gates rather than one, because leaving either would produce a room where the host can override through one door and not the other — harder to explain than either rule on its own.

`VoteWinner` stays, with its tests. It is still the room's answer and both clients use it for the WINNING badge and the one-press "Start winner" action; it simply no longer binds the host.

Client side follows in Silo-Server/silo-android#106 — the per-row "Pick" returns in vote rooms, where it had been hidden precisely because the server used to refuse it.

Note this also makes the PR title slightly overstated: vote rooms *do* now vote, and the tally drives the default action, but it is a recommendation rather than a constraint.

Quick104:
Closing because this branch has been superseded by the binding vote-room implementation now on main. The final commit changes the semantics to advisory voting with a host override, while the PR title and body still describe winner-only enforcement; resolving the conflicts would therefore silently reverse the current behavior. If advisory voting is still desired, please open a fresh, focused PR against current main with the updated rationale and tests. Thanks for the work here.

PR #487, opened 2026-07-26
fix(watch-together): stop a dropped socket reading as the host leaving

RXWatcher:
## Problem

`hostDisconnectTTL` was 15 seconds. That is not "how long before we assume the host left" — an explicit leave and an explicit close both tear the room down **immediately**, so this timer only ever covers a host who has *not* said they are going.

At 15s any transient drop read as a departure. A host who backgrounded the app, hit a brief network blip, or simply moved to a screen where the client did not hold the socket open lost the room for everybody, mid-conversation, with a `host_left` reason nobody could explain.

## Change

Two minutes. Long enough to survive a reconnect or an app switch; short enough that a genuinely departed host does not leave a room sitting open all evening. The janitor still reaps idle rooms independently, and both explicit paths still close immediately.

## Why now

This is one half of a pair. Silo-Server/silo-android#106 moves the room's websocket out of screen scope so a member can stay in a room while browsing for something to suggest — which is what a vote room needs its members to do.

The two are complementary rather than redundant, and neither is sufficient alone:

- The **client** change stops us reporting "host left" when the host merely navigated. That is the bug; papering over it with a longer timeout alone would leave the room silently disconnected while it looked fine.
- This **server** change covers the drops the client cannot prevent — backgrounding, connectivity, process death — where 15 seconds is simply too tight a window to reconnect in.

Reviewing either alone invites the wrong question ("why hoist if the TTL is generous?" / "why 2 minutes if navigation no longer drops the socket?"). They answer different failure modes.

## Testing

`internal/watchtogether` green. The existing suite already overrides this value where it matters (`service.hostDisconnectTTL = time.Hour`), so no test asserted the old constant.

**Not exercised against a real disconnect.** The value change is trivially correct in isolation; what is unverified is the end-to-end behaviour it exists to support, which needs two devices in a room with one of them backgrounding.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who described the flow — create a room, leave it, find something to suggest, come back — that this 15s timeout made impossible.

## Summary by CodeRabbit

* **New Features**
  * Added vote-based winner validation when promoting suggestions.
  * Vote-mode rooms now select playback based on the winning suggestion.
* **Bug Fixes**
  * Prevented direct content selection in vote-mode rooms.
  * Added deterministic tie-breaking for equal vote counts.
  * Added clear `409 Conflict` responses for invalid vote-room actions (including no votes yet / not the winner).
* **Tests**
  * Added coverage for winner selection and vote-mode gating behavior.
* **Chores**
  * Increased the host-disconnect grace period to two minutes.

### DIFF
--- a/internal/api/handlers/watch_together.go
+++ b/internal/api/handlers/watch_together.go
@@ -577,6 +577,12 @@ func (h *WatchTogetherHandler) HandlePromoteSuggestion(w http.ResponseWriter, r 
 		case errors.Is(err, watchtogether.ErrNoVotesCast):
 			writeError(w, http.StatusConflict, "no_votes_cast",
 				"Nobody has voted yet")
+		// Promoting the winner is the sanctioned way into a vote room's
+		// selection, so this should not escape the service. Mapped anyway so a
+		// regression in that path reads as a conflict rather than a 500.
+		case errors.Is(err, watchtogether.ErrVoteRoomSelection):
+			writeError(w, http.StatusConflict, "vote_room_selection",
+				"This room votes for what plays; start the winning suggestion instead")
 		case errors.Is(err, watchtogether.ErrSuggestionNotFound):
 			writeError(w, http.StatusNotFound, "not_found", "Suggestion not found")
 		case errors.Is(err, watchtogether.ErrRoomNotFound):
--- a/internal/api/handlers/watch_together.go
+++ b/internal/api/handlers/watch_together.go
@@ -568,6 +568,12 @@ func (h *WatchTogetherHandler) HandlePromoteSuggestion(w http.ResponseWriter, r 
 		switch {
 		case errors.Is(err, watchtogether.ErrRoomForbidden):
 			writeError(w, http.StatusForbidden, "forbidden", "Only the host can promote a suggestion")
+		case errors.Is(err, watchtogether.ErrNotVoteWinner):
+			writeError(w, http.StatusConflict, "not_vote_winner",
+				"This room votes for what plays; start the title that is winning")
+		case errors.Is(err, watchtogether.ErrNoVotesCast):
+			writeError(w, http.StatusConflict, "no_votes_cast",
+				"Nobody has voted yet")
 		case errors.Is(err, watchtogether.ErrSuggestionNotFound):
 			writeError(w, http.StatusNotFound, "not_found", "Suggestion not found")
 		case errors.Is(err, watchtogether.ErrRoomNotFound):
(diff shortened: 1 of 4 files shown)


Quick104:
Pushed 769c28bf to this branch (maintainer edit) — it fixes a blocker in the vote-gating half of the diff and adds the tests for it.

**Vote rooms could not start playback by any route.** The gate landed on both doors into a room's selection, but `PromoteSuggestion` walks through `SelectItem` to commit, so the guard meant to stop the host bypassing the vote also stopped the vote itself. Direct select was gated, and promote — the only intended path — was gated by the same check.

The fix splits the commit path: public `SelectItem` keeps the gate for direct requests; `PromoteSuggestion` goes through an internal `selectItem(..., viaVote)` once it has confirmed the suggestion is the winner. `ErrVoteRoomSelection` is now also mapped in `HandlePromoteSuggestion` — unreachable after the fix, but there so a regression in that path reads as a conflict instead of a 500.

**Tests.** `internal/watchtogether/vote_selection_test.go` covers promote-the-winner starts the room, promote-a-runner-up → `ErrNotVoteWinner`, promote-before-any-votes → `ErrNoVotesCast`, direct select in a vote room → `ErrVoteRoomSelection`, and host_pick untouched by either gate. Confirmed the first one fails against the pre-fix code. The existing `vote_winner_test.go` only exercised the pure `winnerFrom` helper, which is why the suite stayed green while vote rooms were broken.

**TOCTOU on the tally** (`VoteWinner` reads unlocked, `SelectItem` commits after): documented rather than closed. The host presses start on the standings the room could see, and closing the window means holding the room lock across a suggestion-store read, which stalls every other room for a worst case of one vote. Reasoning is in the comment above the check.

**Open question for review — API compatibility.** `PUT /rooms/{id}/selection` used to succeed in a vote room and now returns `409`. That is a previously-succeeding request that now fails on an existing `/api/v1` endpoint, which the additive-only rule in CLAUDE.md points at. I left the gate in because it is the point of the change, and the exposure looks small: the web client never calls it in vote mode (`WatchTogetherRoomPage.tsx:451` routes vote rooms to "suggest"), and `selection_mode` is already on the room response so clients have the feature-detection signal. Someone should confirm `silo-apple` and `silo-android` do not call `/selection` directly in a vote room — if either does, this needs a coordinated client change before it ships.

Also worth noting: the PR description still describes this as a one-line `hostDisconnectTTL` change, but the diff also hard-gates selection in vote rooms. Worth updating the body so reviewers know to look at that half.

Verified: `go test -count=1 -race ./internal/watchtogether/` passes; `gofmt` clean on both touched packages. Not run: `make lint` (golangci-lint not installed locally) and `go build ./...` (fails on the pre-existing `web/embed.go` missing `dist`, unrelated).

PR #494, opened 2026-07-27
feat(diagnostics): chunked report upload fallback for proxy body caps

Quick104:
## Problem

Diagnostics bundles can be up to `max_bundle_bytes` (10 MiB default), but the production edge in front of Silo caps request bodies at nginx's default `client_max_body_size` of **1 MiB** (verified against the live deployment: a body of 1,048,576 bytes reaches Silo, 1,048,577 bytes gets an OpenResty 413 in 90 ms). The proxy's 413 carries no Silo error envelope, the bundle never arrives, and the client can never deliver any report over the cap — the "failed to send" seen on tvOS.

## Change

Adds a chunked upload fallback under `/api/v1/diagnostics/reports/uploads`:

| Step | Route | Notes |
|---|---|---|
| init | `POST /` | `{manifest, bundle_bytes}` — manifest verbatim, size checked against live `max_bundle_bytes` |
| chunk | `PUT /{id}/chunks/{index}` | raw bytes, fixed 768 KiB chunk size (proxy-safe) |
| complete | `POST /{id}/complete` | ingests the assembled bundle, same 201 shape as single-shot |
| abort | `DELETE /{id}` | best-effort, idempotent |

Design points:

- **One validation authority.** The assembled bundle goes through the exact same `Ingest` path as the single-shot endpoint — manifest contract, archive sha/bytes/entries equality, quotas, and profile attribution all apply identically. Init validates only sizes.
- **Reuses `internal/uploads`** (the plugin chunked-upload spool manager) with a small owner map on top for per-user session isolation (foreign sessions read as 404; foreign aborts are no-ops). Added a read-only `Manager.Peek` for expiry sweeps.
- **Bounded buffering.** Sessions spool to disk under `os.TempDir()/silo-diagnostics-uploads`, expire after 15 min, cap at 1/user and 16 global (worst case ≈ 160 MiB at default bundle size). A new init replaces the user's previous session so a crashed client can restart immediately.
- **Capacity shared with single-shot.** `complete` acquires the existing per-user + global in-flight limiter; a `busy` answer leaves the session intact for a retry of complete.
- `/diagnostics/status` now advertises `upload_chunk_bytes`; older servers omit it and clients treat that as unsupported (Android's status model uses `ignoreUnknownKeys`, the Apple model has the field optional).
- Demo guard: the diagnostics prefix now also blocks `PUT` to cover the chunk route.

## Verification

- New handler tests: happy path (multi-chunk assembly byte-for-byte into Ingest), init over bundle limit / disabled / manifest-too-large, foreign-session isolation, incomplete complete, re-init replacement, busy-complete retryability, oversized chunk, ingest error mapping, status advertisement.
- `go test ./internal/api/handlers ./internal/uploads ./internal/diagnostics/...` green. (`TestHandleReplanPlaybackV3SeekFailureRecoveryNeverChangesMediaVersion` fails on a clean origin/main checkout too — pre-existing, unrelated.)
- **End-to-end** against an OpenResty proxy with the 1m default cap: single-shot 413s, the same 1.6 MiB bundle uploads in 3 chunks and lands as an accepted, fully validated report. Also exercised from the tvOS simulator via the client fallback (client PR: Silo-Server/silo-apple — chunked upload + bare-413 classification).

## Summary by CodeRabbit

- **New Features**
  - Added chunked diagnostics report uploads with init, chunk, complete, and abort workflows.
  - Diagnostics status now advertises the configured diagnostics upload chunk size.
- **Bug Fixes**
  - Improved demo-mode protections to block destructive diagnostics upload routes for non-admin users.
  - Enhanced diagnostics upload deadline handling and chunk concurrency behavior (returns a clear conflict when a chunk is already uploading).
- **Tests**
  - Expanded end-to-end HTTP tests and upload-session concurrency/lifecycle coverage (including busy, retryable, limits, and authorization scenarios).

### DIFF
--- a/internal/api/handlers/diagnostics.go
+++ b/internal/api/handlers/diagnostics.go
@@ -57,6 +57,10 @@ func NewDiagnosticsHandler(service DiagnosticsService) *DiagnosticsHandler {
 		chunkSessions: newDiagnosticsChunkSessions(filepath.Join(os.TempDir(), "silo-diagnostics-uploads")),
 		logger:        slog.Default(),
 	}
+	// Reclaim abandoned chunk spool bytes on a timer, not only from later API
+	// traffic. The handler lives for the process, so the sweeper needs no
+	// stop signal.
+	handler.chunkSessions.startSweeper(nil)
 	if adminService, ok := service.(AdminDiagnosticsService); ok {
 		handler.adminService = adminService
 	}
@@ -76,9 +80,14 @@ func (h *DiagnosticsHandler) HandleStatus(w http.ResponseWriter, r *http.Request
 	writeJSON(w, http.StatusOK, status)
 }
 
-func (h *DiagnosticsHandler) HandleUpload(w http.ResponseWriter, r *http.Request) {
-	// Extend the read deadline for this request only; the server-wide
-	// ReadTimeout is too short for slow, large bundle uploads (see the constant).
+// extendDiagnosticsUploadDeadlines lifts this request's read (and, when
+// includeWrite, write) deadline to diagnosticsUploadReadTimeout. The
+// server-wide ReadTimeout (30s) is too short for slow bundle/chunk uploads,
+// and the WriteTimeout (120s) can expire after Ingest has already stored the
+// report — the lost 201 would make the client retry an upload that
+// succeeded. Shared by the single-shot upload, chunk PUT (read only), and
+// chunked complete (both).
+func (h *DiagnosticsHandler) extendDiagnosticsUploadDeadlines(w http.ResponseWriter, r *http.Request, includeWrite bool) {
 	rc := http.NewResponseController(w)
 	if err := rc.SetReadDeadline(time.Now().Add(diagnosticsUploadReadTimeout)); err != nil {
 		h.diagnosticsLogger().WarnContext(r.Context(), "diagnostics upload read deadline not extended",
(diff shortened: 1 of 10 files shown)


Quick104:
Addressed the review feedback in d50b29ff:

- **Atomic session admission (Codex P1, CodeRabbit critical):** init now reserves the per-user slot and a global cap slot in one critical section (`reserve` → `commit`/`unreserve`), with in-flight reservations counted against the cap. Concurrent inits by one account get `busy` instead of fanning out; creation failures roll the reservation back. New `-race` test fires 8 parallel inits and asserts ≤1 owned session, 0 leaked reservations, and a working upload afterwards.
- **Manager-wide lock held across body I/O (Codex P1):** chunk body copies now run outside the `uploads.Manager` mutex. A per-chunk in-flight flag (`ErrChunkBusy` → 409) rejects duplicate concurrent writes to the same offset, and cancel/expiry defer spool-dir removal to the last finishing writer. Covered by a test that stalls one writer mid-body and proves another session's chunk write proceeds.
- **Read deadline on chunk PUTs / write deadline on complete (Codex P2 ×2):** both now go through the same `extendDiagnosticsUploadDeadlines` helper as the single-shot handler.
- **TTL as idle timeout (Codex P2):** chunk arrivals refresh `ExpiresAt`; a slow-but-progressing upload no longer dies at an absolute 15-minute deadline. Tested with a manipulated clock.
- **Orphaned spool dirs after restart (Codex P2):** `Manager.ReclaimOrphanedDirs()` runs at handler construction and removes root-dir entries no live session owns.
- **No sweep without traffic (Codex P2):** `Manager.StartExpirySweeper` runs cleanup on a 1-minute ticker for the process lifetime, so abandoned sessions are reclaimed during quiet periods.
- **Transient status failure destroying the spool (Codex P2):** complete now distinguishes a status *load error* (500, session kept, complete retryable without re-upload) from a definitive disabled/storage-unavailable answer (session discarded). Tested.
- **Process-local sessions (CodeRabbit nitpick):** documented in the file header — this targets the integrated single-process server; a multi-replica API deployment needs sticky routing for one upload's few requests, or the client transparently restarts from init on the 404.

Re-verified: `go test -race` on handlers + uploads green, and the full curl E2E (init → 3 chunks → complete → 201, accepted report) against the updated binary.

Quick104:
Addressed the second review round in 856d000c:

- **Detached writers escaping the cap (Codex P1):** `uploads.Manager` now parks a canceled/expired session whose chunk writer is still draining into a `detached` set instead of forgetting it; the last finishing writer removes both the entry and the spool directory. `DetachedWriterSessions()` exposes the count and the diagnostics init admission gate adds it to `owners + reserved`, so a cancel-and-reinit loop now stalls at the 16-session cap instead of stacking unbounded live writers behind it. New test stalls a writer mid-body, cancels the session, and asserts the count is 1 during the write and 0 once the writer drains.
- **Chunk PUT write deadline (Codex P2):** chunk PUTs now extend the write deadline too (previously read-only) — on an uplink slow enough to outlive the 120s WriteTimeout, the stored chunk's JSON acknowledgement would have been lost and the client would retry an already-accepted chunk.

Re-verified: `go test -race` green on uploads + handlers; redeployed the dev server and re-ran the E2E through the 1 MiB-capped OpenResty proxy — curl chunked flow (init → 3×768 KiB chunks → complete → 201 accepted) and the tvOS simulator client fallback (single-shot 413 → init → chunk → complete → 201 per report) both pass on this build.

PR #515, opened 2026-07-29
feat(autoscan): make source setup descriptor-driven and self-contained

Quick104:
## Problem

Autoscan setup asked operators to hold five concepts — *plugin*, *capability*, *source*, *connection*, *delivery mode* — before anything scanned, spread across four tabs. Reading the code turned up five concrete failure modes:

1. **The same Sonarr configured twice.** `request_integrations` and `autoscan_connections` are separate encrypted stores for the same base URL + API key. Reuse existed but sat behind a "Connection type" dropdown defaulting to *"Enter own credentials"*, so most operators never saw it.
2. **Connections was a peer tab that most sources don't need.** A webhook source and a filesystem watcher both need none, but its placement implied "set this up first."
3. **"When do we scan?" lived in four places** — autoscan settings, per-source override, the `scan_libraries` task, and `scanner.schedule`.
4. **A source never said which library it fed.** Path rewrites are the real binding and were collapsed behind a disclosure, so "I set it up and nothing happened" usually meant missing rewrites, with no visible signal.
5. **Third-party plugins could not configure themselves.** `SourcesPanel.tsx` held **33 references** to `silo.autoscan.cephfs` / `silo.autoscan.arr-webhook`, plus a bespoke CephFS editor with hardcoded field keys. A new scan-source plugin could render no config UI without a PR to silo-server.

## Approach

**Host: a `ScanSourceDescriptor` per capability**, read from manifest metadata — delivery modes, connection requirement, connection kinds, and the per-source config form. This follows the precedent already set by `WatchSyncProviderDescriptor`: typed metadata that lets the host build setup UI *without launching the plugin*.

Capabilities declaring nothing resolve to `poll` + `optional` connection — exactly the pre-descriptor behavior, so **existing installs are unaffected**. Malformed descriptors degrade to defaults rather than making a plugin undiscoverable.

No SDK release is needed: generic capability `metadata` already round-trips intact through `convert.go`. (Only the *typed* `admin_form` under `config_schema` is dropped, which a later SDK change can clean up.)

**Compatibility descriptors** for the two first-party plugins live in one file — `internal/autoscan/compat.go` — replacing the 33 scattered checks. The manifest always wins over the compat value, so a plugin takes ownership simply by publishing its own, with no coordinated host change.

**UI: steps are built from the descriptor, not from plugin ids.** A single-mode source is never asked how changes arrive; a credential-free one never sees the connection step. Per-source fields render through the existing `SchemaForm`, so `CephFSConfigEditor` is deleted rather than rewritten.

## What changed for operators

- **4 tabs → 2.** Connections and Settings fold into a collapsed Advanced section. Old `?tab=connections` / `?tab=settings` links land on Sources with it expanded.
- **Connections are created inline.** Previously a dead end: with no saved connection the picker offered only "— No connection —", forcing a cancel-and-restart. Now "+ Add a server" expands in place, pre-selecting a Sonarr already configured under Requests — one click, no API key retyped.
- **Webhook setup finishes in one dialog.** Path mappings, then the URL with the exact triggers to tick. The trigger list is derived from `arrwebhook/parse.go`, not from memory: On Import / On Upgrade / On Rename, plus the service-specific delete trigger. Anything the host ignores is deliberately absent.
- **Rows say what they feed**, and warn when a source can never resolve a library.

### Mapping rows: 96 → 2

Seeding one row per library path looked fine on fixtures and was wrong on real data. Against this install's **96 library paths**, an early "drop nested paths" collapse returned **96 rows** — none of those paths is an ancestor of another; they're all siblings (`movies/00s`, `movies/10s`, `television-int/zh`…). The test I'd written passed only because the fixture happened to contain a parent path.

Collapsing to the **common ancestor per mount point** gives:

```
input paths : 96
mapping rows: 2
  /mnt/sharedrives
  /tmp/silo-transcode/test-extras
```

Safe because `applyRewrites` matches by longest prefix at a segment boundary, so one row per mount covers everything beneath it. Guards: different mounts never merge to `/`, and a mount with a single path is left as-is rather than widened.

## API compatibility

Additive only, within the `/api/v1` rules. `/autoscan/scan-source-plugins` gains optional `description` and `descriptor` fields; the three existing identity fields are untouched. No renames, no type changes, no repurposed status codes. No migration.

## Risks and follow-up

- **Not verified against a live backend.** Every flow was driven in a browser against seeded fixtures via a temporary dev-only harness (deleted before commit). The create → `POST /webhook` → instructions chain and a genuine Sonarr test-payload delivery are **unexercised**. Sonarr's own *Test* button is the quickest check — Silo accepts test payloads and surfaces them on the row.
- **Path mapping can't be auto-detected for webhook sources.** `SuggestRewrites` needs a bound connection to read the arr's root folders (`suggest.go:27-29`); a webhook source has none. The Silo side is seeded from real library paths and the operator supplies the arr side. At least one complete mapping is now required before creation, since a webhook source without one accepts deliveries and resolves nothing.
- **Scanner & Matcher settings deliberately stay** on the Settings page — that's performance tuning, not setup.
- Client repos need no follow-up: no client-visible contract changed.

## Verification

```
$ make verify-local-paths
scripts/check-local-path-leaks.sh          # exit 0

$ go build ./...                            # ok

$ go test ./internal/autoscan/...
ok  	github.com/Silo-Server/silo-server/internal/autoscan	0.058s
ok  	github.com/Silo-Server/silo-server/internal/autoscan/arrwebhook	0.048s

$ go test ./internal/api/handlers/ -run 'Autoscan'
ok  	github.com/Silo-Server/silo-server/internal/api/handlers	0.091s

$ cd web && pnpm run lint
✖ 156 problems (0 errors, 156 warnings)     # all pre-existing, vendor + ui primitives

$ pnpm run format:check

$ pnpm run build
✓ built in 17.48s

$ ./node_modules/.bin/vitest run src/pages/admin src/components/admin
Test Files  38 passed (38)
     Tests  239 passed (239)                # 39 new across 3 suites
```

**Known pre-existing failures, not introduced here** (both reproduce on a clean tree at `a7601028`):
- `internal/api/handlers`: `TestHandleReplanPlaybackV3SeekFailureRecoveryNeverChangesMediaVersion`, plus a jellycompat `TempDir` cleanup flake
- `web`: 6 failures in `ServerStorageStep.test.tsx` and `SeasonContent.test.tsx`
- `golangci-lint` reports 51 pre-existing `goconst` findings in these packages before this branch

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Adversarial review: Reviewing my own diff found a real bug — a webhook source is created *enabled* and its endpoint generated in a second request; if that request failed the dialog closed, leaving an enabled source with no endpoint and no explanation. Fixed in 66d7d032 by holding the created source and rendering a retry. The review also confirmed `POST /webhook` returns `webhook_url` in its response (so the instructions panel has a real URL), and that collapsing paths is safe only because rewrites match at segment boundaries — `/mnt/media` cannot swallow `/mnt/media2`, which has a regression test. Separately, checking against real server data — rather than my fixtures — is what exposed the 96→96 collapse bug described above.

## Summary by CodeRabbit

* **New Features**
  * Added descriptor-driven autoscan setup in the Add-source flow, including per-source delivery/connection requirements, config forms, and richer descriptions/icons/summaries.
  * Upgraded the Sources admin experience to be descriptor-driven, including library-path “fill from” helpers and improved webhook setup (URL copy, provider selection, path mapping editor with splitting).
  * Added an inline connection picker with “reuse” when compatible integrations exist.
* **Bug Fixes**
  * Improved resilience when scan-source metadata is missing or malformed, including safer defaults and legacy CephFS compatibility behavior.
  * Fixed a React admin autoscan rendering issue that could cause infinite re-renders.
* **Tests**
  * Added/expanded descriptor and UI logic test coverage for parsing, compatibility, and webhook mapping helpers.

### DIFF
--- a/internal/api/autoscan_wiring.go
+++ b/internal/api/autoscan_wiring.go
@@ -99,20 +99,51 @@ func (l PluginScanSourceLister) ListScanSources(ctx context.Context) ([]autoscan
 				PluginID:     inst.PluginID,
 				CapabilityID: c.ID,
 				DisplayName:  scanSourceDisplayName(inst.PluginID, c),
+				Description:  scanSourceMetadataString(c, "description"),
+				Descriptor:   scanSourceDescriptor(inst.PluginID, c),
 			})
 		}
 	}
 	return out, nil
 }
 
+// scanSourceMetadataString reads a trimmed string out of a capability's
+// manifest metadata, returning "" when absent or of the wrong type.
+func scanSourceMetadataString(c *plugins.Capability, key string) string {
+	if c == nil || c.Metadata == nil {
+		return ""
+	}
+	value, ok := c.Metadata[key].(string)
+	if !ok {
+		return ""
+	}
+	return strings.TrimSpace(value)
+}
+
+// scanSourceDescriptor resolves the setup contract for a discovered capability.
+// It reads whatever the manifest declares (falling back to host defaults that
+// reproduce pre-descriptor behavior), then lets a host-side compatibility
+// descriptor fill gaps for first-party plugins that have not yet published one
+// of their own.
+func scanSourceDescriptor(pluginID string, c *plugins.Capability) autoscan.ScanSourceDescriptor {
+	var metadata map[string]any
+	if c != nil {
+		metadata = c.Metadata
+	}
+	declared := autoscan.DescriptorFromMetadata(metadata)
+	capabilityID := ""
+	if c != nil {
+		capabilityID = c.ID
+	}
+	return autoscan.ApplyCompatibilityDescriptor(pluginID, capabilityID, declared)
+}
+
 // scanSourceDisplayName derives a human-friendly label for a scan_source
 // capability: the capability manifest's display_name when present, else the
 // plugin id (with the capability id appended when it adds information).
 func scanSourceDisplayName(pluginID string, c *plugins.Capability) string {
-	if c != nil && c.Metadata != nil {
-		if name, ok := c.Metadata["display_name"].(string); ok && strings.TrimSpace(name) != "" {
-			return strings.TrimSpace(name)
-		}
+	if name := scanSourceMetadataString(c, "display_name"); name != "" {
+		return name
 	}
 	switch {
 	case pluginID != "" && c != nil && c.ID != "":
--- a/internal/api/handlers/autoscan.go
+++ b/internal/api/handlers/autoscan.go
@@ -454,10 +454,15 @@ func (h *AutoscanHandler) HandleListSources(w http.ResponseWriter, r *http.Reque
 
 // --- Available scan-source plugins (Add-source picker) ---
 
+// autoscanScanSourcePluginResponse describes one creatable scan source. The
+// descriptor is additive: clients that predate it keep reading the same three
+// identity fields.
 type autoscanScanSourcePluginResponse struct {
-	PluginID     string `json:"plugin_id"`
-	CapabilityID string `json:"capability_id"`
-	DisplayName  string `json:"display_name"`
+	PluginID     string                        `json:"plugin_id"`
+	CapabilityID string                        `json:"capability_id"`
+	DisplayName  string                        `json:"display_name"`
+	Description  string                        `json:"description,omitempty"`
+	Descriptor   autoscan.ScanSourceDescriptor `json:"descriptor"`
 }
 
 // HandleListAvailableScanSources returns every installed scan_source capability
@@ -474,6 +479,8 @@ func (h *AutoscanHandler) HandleListAvailableScanSources(w http.ResponseWriter, 
 			PluginID:     a.PluginID,
 			CapabilityID: a.CapabilityID,
 			DisplayName:  a.DisplayName,
+			Description:  a.Description,
+			Descriptor:   a.Descriptor,
 		})
 	}
 	writeJSON(w, http.StatusOK, struct {
--- a/internal/autoscan/adminform.go
+++ b/internal/autoscan/adminform.go
@@ -1,6 +1,9 @@
 package autoscan
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"strings"
+)
 
 // AdminForm describes per-source configuration fields for a scan source, in the
 // same shape the admin UI's generic schema renderer already consumes for plugin
@@ -50,6 +53,7 @@ const (
 	ControlText     = "TEXT"
 	ControlTextarea = "TEXTAREA"
 	ControlSelect   = "SELECT"
+	ControlPassword = "PASSWORD"
 )
 
 // Fill sources the admin UI understands for AdminFormField.FillFrom.
@@ -111,8 +115,28 @@ func adminFormFromMetadata(value any) *AdminForm {
 	if err := json.Unmarshal(raw, &form); err != nil {
 		return nil
 	}
+	form.Fields = withoutSecretFields(form.Fields)
 	if len(form.Fields) == 0 {
 		return nil
 	}
 	return &form
+}
+
+// withoutSecretFields drops fields a plugin marked as holding a credential.
+//
+// A source's values land in autoscan_sources.source_config, which is plain
+// JSONB and is returned verbatim by the source API — unlike connection API
+// keys, which go through the repository's encrypted path. Rendering a masked
+// input over a value stored in the clear would misrepresent how it is held, so
+// the host declines to collect it at all. Plugins needing a credential should
+// take a connection instead.
+func withoutSecretFields(fields []AdminFormField) []AdminFormField {
+	kept := make([]AdminFormField, 0, len(fields))
+	for _, field := range fields {
+		if field.Secret || strings.EqualFold(field.Control, ControlPassword) {
+			continue
+		}
+		kept = append(kept, field)
+	}
+	return kept
 }
--- a/internal/autoscan/adminform.go
+++ b/internal/autoscan/adminform.go
@@ -115,25 +115,34 @@ func adminFormFromMetadata(value any) *AdminForm {
 	if err := json.Unmarshal(raw, &form); err != nil {
 		return nil
 	}
-	form.Fields = withoutSecretFields(form.Fields)
+	form.Fields = withoutUnsupportedFields(form.Fields)
 	if len(form.Fields) == 0 {
 		return nil
 	}
 	return &form
 }
 
-// withoutSecretFields drops fields a plugin marked as holding a credential.
+// withoutUnsupportedFields drops fields the source-config surface cannot honour.
 //
-// A source's values land in autoscan_sources.source_config, which is plain
-// JSONB and is returned verbatim by the source API — unlike connection API
-// keys, which go through the repository's encrypted path. Rendering a masked
-// input over a value stored in the clear would misrepresent how it is held, so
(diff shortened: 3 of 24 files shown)


Quick104:
Worked through all 18 findings. Each was verified against the code before acting; two are declined with reasons.

## Fixed

**Secrets in `source_config`** (Codex P1) — confirmed: `source_config` is plain JSONB and is returned verbatim by the source API, unlike connection keys which take the repository's encrypted path. Rather than render a masked input over a plaintext value, the host now drops `secret`/`PASSWORD` fields from scan-source forms entirely. Plugins needing a credential should take a connection.

**Plugin-declared webhook mode** (Codex P2) — confirmed against `resolveDeliveryMode`: webhook is accepted only for the built-in ARR identity, so a plugin claiming it produced a setup option whose every submission returned HTTP 400. Now dropped at parse time. The built-in supplies its descriptor directly, so it keeps webhook.

**Typed schema values** (Codex P1) — confirmed. `String()`-ing every change turned `false` into the truthy `"false"` (switch re-rendered enabled) and an array into a join `SchemaForm` read back as no selection. Values now stay typed until submit; `serializeConfigValues`/`parseConfigValues` handle the boundary, with round-trip tests for switch, multi-select, and number.

**Submit gated on form validity** (Codex P1) — the host stores `source_config` uninterpreted, so required/validated fields are only enforceable client-side. `onValidityChange` is now wired into both the add flow and the row editor's save button.

**Row editing ignored the connection contract** (Codex P1) — a `required` source could be unbound and a `none` source could be given credentials right after the add flow correctly prevented it. The row now applies the same descriptor rules.

**Compat match was OR** (CodeRabbit) — capability ids are author-chosen and not unique, so any plugin naming a capability `cephfs` inherited CephFS's form. Now requires both identifiers, with a regression test.

Also fixed: disabled libraries excluded from `sourceTargets` (the scanner rejects them, so a source looked wired while its deliveries were dropped); inline reuse now compares `request_integration_id` rather than connection id (different identifier spaces — the dedupe matched nothing); disabled Requests integrations excluded from reuse; webhook instructions follow the operator's chosen provider instead of always resolving to `auto`; webhook rows render their plugin-declared config form; stable mapping-row ids; memoized fields; `descriptor` marked optional in the TS type.

## Declined, with reasoning

**Sibling-ancestor collapse** (Codex P1) — keeping it, with a mitigation.

The concern is correct in principle: one prefix rule cannot serve branches the arr exposes under different roots (`/downloads/films` vs `/downloads/series`). But I tried the suggested "only eliminate paths already covered by an actual ancestor" against this deployment's real data and it produces **96 rows from 96 paths** — zero reduction, because none of those paths is an ancestor of another. They are all siblings. That is the sprawl the collapse exists to remove.

So rather than drop it, the editor now offers to split a collapsed row into one row per branch (`expandedRootsFor`), pre-filled from the same library paths. The seeded rows are a starting point, not a constraint — and an operator with divergent arr roots gets one click instead of retyping.

**Legacy-key comment** (CodeRabbit) — the comment was wrong, not the code. Migrating the row forward on first save is the intended behavior; carrying both keys indefinitely would mean the alias never goes away. Comment corrected to describe what actually happens.

## Verification

```
$ go build ./...                                    # ok
$ go test ./internal/autoscan/...
ok  	github.com/Silo-Server/silo-server/internal/autoscan	0.057s
ok  	github.com/Silo-Server/silo-server/internal/autoscan/arrwebhook

$ go test ./internal/api/handlers/ -run 'Autoscan'  # ok
$ make verify-local-paths                            # ok

$ cd web && pnpm run lint
✖ 156 problems (0 errors, 156 warnings)              # all pre-existing

$ pnpm run format:check

$ pnpm run build                                     # ✓ built in 15.52s

$ vitest run src/pages/admin src/components/admin
Test Files  38 passed (38)
     Tests  247 passed (247)                         # was 239; +8 for these fixes
```

Unchanged caveat from the PR body: none of this is exercised against a live backend — the create → `POST /webhook` → instructions chain still needs a click-through on a real server.

Quick104:
Second review pass addressed in `5567a95e`. Two of these were serious — the descriptor mechanism was effectively dead for real plugins, and I'd have shipped it that way.

## The two P1s that mattered

**Nested capability metadata.** Verified against the pinned SDK: `CapabilityRecordsFromManifest` stores a capability's arbitrary metadata struct under `Capability.Metadata["metadata"]` rather than flattening it. My top-level `metadata["scan_source"]` lookup therefore matched nothing for any installed plugin — every one silently resolved to poll/optional defaults. The whole "plugins configure themselves" premise of this PR was inert on the real install path. Now reads the nested location first, keeping the flat path for host-built descriptors and tests.

**`admin_form` dropped on persist.** Same converter writes `config_schema` entries with only key/title/description/json_schema/required. A plugin's form could never survive installation through that route. The canonical location is now `config_form` inside the typed `scan_source` block, which does survive; `config_schema[].admin_form` is retained only as a fallback for host-built descriptors.

Both now have regression tests that construct the exact persisted shape, which is what I should have done originally instead of testing against hand-built flat maps.

**Misleading path label.** `arrwebhook.Parse` consumes `episodeFile.path` / `movieFile.path` — the arr's *imported library* paths, not the download client's working directory. My "Path your download manager reports" label with a `/downloads/tv` placeholder would have produced rewrites that never match, so deliveries are accepted and silently enqueue nothing. Relabelled to "Sonarr/Radarr root folder" with a `/tv` example.

**Invalid config leaking through unrelated saves.** Correct — only the config button was gated, while enable/label/interval/rewrite mutations all called `fullBody` with the in-progress draft. Unrelated saves now send the persisted config.

## Also fixed

Compat merging now tracks *declared* fields rather than comparing to defaults (a plugin explicitly declaring `connection: "optional"` was indistinguishable from silence and got overwritten — my "manifest wins" claim was false for exactly the values a plugin is most likely to state); legacy key migration scoped to the CephFS identity; rows re-parse when the descriptor arrives after `/sources`; `show_when`-hidden values dropped and displayed defaults included, both via the existing `buildSchemaValues`; manual connection form asks which service when a descriptor accepts several; picker remounts per source; `dynamic_options` fields rejected at the contract since no option-loading path exists for source forms; `emits_native_paths` sources report unknown targets rather than "can't match anything".

## On the rehydration fix

My first attempt used `useEffect` + `setState`, which the repo's `react-hooks/set-state-in-effect` rule flagged — the same class of bug as the render loop in `35ea359f`. Rewrote it as render-phase derivation. Lint is back to the 156-warning baseline.

Verified in a browser with the descriptor arriving *after* `/sources`: a stored `"false"` now renders as an off switch (previously enabled), zero React console errors.

## Verification

```
$ go build ./...                                    # ok
$ go test ./internal/autoscan/...
ok  	github.com/Silo-Server/silo-server/internal/autoscan
ok  	github.com/Silo-Server/silo-server/internal/autoscan/arrwebhook

$ make verify-local-paths                            # ok

$ cd web && pnpm run lint
✖ 156 problems (0 errors, 156 warnings)              # baseline

$ pnpm run format:check

$ vitest run src/pages/admin src/components/admin
Test Files  39 passed (39)
     Tests  248 passed (248)

$ pnpm run build                                     # ✓ built in 17.55s
```

Unchanged caveat: still no live-backend run of create → `POST /webhook` → instructions. Given the nested-metadata bug, a real plugin install is now the most valuable thing to exercise before merge.

Issue #523, opened 2026-07-31
events: add a read-only SSE transport alongside the WebSocket hub


RXWatcher:
The events hub is reachable only over `GET /api/v1/events/ws`. That is the right transport for the first-party clients, which need bidirectional control — but it is an expensive way to *observe*.

A read-only consumer today has to:

- send a `subscribe` frame within 5 seconds or be closed with a policy violation. A client that connects and waits silently receives nothing and is given no reason;
- implement the handshake, ping/pong, reconnect and backoff itself;
- carry a WebSocket client library, in ecosystems where SSE is the standard-library answer for one-way server push.

That cost is paid by every observer: companion apps, home-automation glue, contract tests that assert on live server behaviour, and any future first-party surface that only needs to watch. It is a barrier to writing a short integration against Silo.

`events.Hub.Subscribe()` already returns a plain channel of envelopes — only the delivery layer is WebSocket-bound, so the cost is incidental rather than inherent.

## Proposal

`GET /api/v1/events/sse?channels=sessions,jobs` — a read-only sibling of the WebSocket handler over the same hub, reusing the existing role-based channel allow-list and per-event authorization. The WebSocket path is unchanged and remains the transport for clients that need control.

Design points worth agreeing before implementation:

- **Channel selection is a query parameter, not a handshake.** SSE is one-way, so there is nothing to negotiate, and no subscribe deadline is needed to distinguish a live client from a stalled one. Omitting `channels` subscribes to everything the caller's role allows; requesting a forbidden channel omits it silently rather than failing the whole stream.
- **Bearer auth, no ws-ticket.** SSE requests can set an `Authorization` header, so the ticket mechanism is unnecessary. Consequence: profile-scoped notification channels are *not* offered over SSE, since they depend on that binding.
- **Frames** are named SSE events whose name is the channel, so a consumer can `addEventListener('sessions', …)` and ignore the rest. A `hello` frame on connect and periodic `ping` keepalives, mirroring the WebSocket hello.
- **Discoverable via a capability endpoint** rather than version sniffing, per the v1 API rules — `GET /api/v1/events/capability` reporting both transports.

## Notes

Two implementation details that are easy to miss and worth recording here:

[...]

PR #524, opened 2026-07-31
feat(events): add a read-only SSE transport for the events hub

RXWatcher:
Part of #523

## Problem

The events hub is reachable only over `GET /api/v1/events/ws`. That is the right transport for the first-party clients, which need bidirectional control, but it is an expensive way to *observe*: a read-only consumer must send a `subscribe` frame within 5 seconds or be closed with a policy violation (and is told nothing), implement handshake, ping/pong, reconnect and backoff itself, and carry a WebSocket library in ecosystems where SSE is the standard-library answer for one-way push.

`events.Hub.Subscribe()` already returns a plain channel of envelopes — only the delivery layer is WebSocket-bound, so that cost is incidental rather than inherent.

## Approach

`GET /api/v1/events/sse?channels=sessions` — a read-only sibling of `HandleWebSocket` over the same hub, reusing `allowedChannelsForRole` and `allowsEventForClaims` unchanged. The WebSocket handler, the hub, and event publishing are untouched; this adds a transport and changes no existing behaviour. The diff is entirely additive (7 files, +1084, −0).

Three deliberate differences from the WebSocket path:

- **Channel selection is a query parameter, not a handshake.** SSE is one-way, so there is nothing to negotiate and no subscribe deadline is needed. Requesting a channel the caller's role forbids omits it silently rather than failing the stream, so one forbidden channel does not deny the rest.
- **Bearer auth, no ws-ticket.** SSE requests can set headers, so the ticket mechanism is unnecessary. Consequently profile-scoped notification channels are *not* offered here — `allowsEventForClaims` is called with an empty `boundProfileID`, which can only narrow access, never widen it.
- **`X-Accel-Buffering: no`**, since openresty fronts some deployments and would otherwise buffer the stream indefinitely.

Discovery is via `GET /api/v1/events/capability`, reporting both transports and the schema version, per the v1 rule that features be feature-detected rather than version-sniffed. It follows the existing per-subsystem capability convention (`/notifications/capability`, `/playback/capability`, `/downloads/capability`, …) rather than adding a flag to an unrelated surface.

## Risks and follow-ups

- **Backpressure** matches the WebSocket path exactly: the hub gives each subscriber a buffered channel and drops for slow consumers, so a slow SSE client loses events rather than stalling the hub.
- **Not replicated from the WebSocket path:** its snapshot-refetch-on-empty-data enrichment. Deliberate — that behaviour is coupled to the control-plane use case. Worth revisiting if an SSE consumer needs it.
- **Deployments behind other proxies** may need their own buffering opt-out; only the openresty header is set here.

## Verification

Actual output, not summarized:

```
$ go test -count=1 ./internal/api/handlers/ ./internal/events/
ok  github.com/Silo-Server/silo-server/internal/api/handlers  26.254s
ok  github.com/Silo-Server/silo-server/internal/events         1.022s

0 issues.

$ make verify-local-paths
(pass)

Checking formatting...

$ cd web && pnpm run lint
✖ 157 problems (0 errors, 157 warnings)
```

The web lint warnings are pre-existing; this branch touches no web files.

`make test-go` fails in `internal/jellycompat` and `internal/playback` (`TestBeginWebOperation*`, `TestResolveHWAccelWithFFmpeg*`, `TestFFmpegSupportsNVENC*`). I verified those fail identically on a clean `upstream/main` worktree — they are hardware/NVENC-dependent and unreachable from a diff that only adds files under `internal/api/handlers`.

## AI Disclosure

- **Tool(s):** Claude Code
- **Model(s):** claude-opus-5
- **Involvement:** fully AI-generated
- **Adversarial review:** an independent AI review of the diff found one critical and two important defects, all fixed before submission.

  **Critical —** the API server sets `WriteTimeout: 120 * time.Second` (`cmd/silo/main.go`), which Go applies as a hard connection deadline set once when the request is read and *not* renewed by handler writes. Every SSE stream would have been force-closed at ~120s. Unit tests could not catch it (`httptest.ResponseRecorder` has no connection) and a short `curl` would have looked healthy. Fixed by clearing the deadline with `http.NewResponseController` inside the handler, leaving the shared server config untouched — mirroring the existing `WriteTimeout: 0 // no timeout for long streams` precedent in `startStandaloneServer`.

  **Important —** the handler wrote the raw `evt.Envelope` to the wire, exposing `source_id`, `admin_only`, `target_plugin_id`, `user_id` and `profile_id`, which `HandleWebSocket` deliberately strips by wrapping in `evt.EventsEventMessage`. Fixed by projecting through the same wire-safe shape so both transports agree on the event schema.

  **Important —** the endpoint's core authorization guarantee (a non-admin cannot obtain an admin-only channel by naming it in the query string) had no test. Added, along with malformed-`channels` coverage and an explicit deny-by-default assertion.

  Also verified during review and left as-is: no hub subscription or ticker leak on any return path; the empty `boundProfileID` can only narrow access; SSE framing is newline-safe because Go's JSON encoder escapes control characters; and chi's `Compress(5)` middleware does not buffer `text/event-stream`.

## Summary by CodeRabbit

* **New Features**
  * Added a read-only Server-Sent Events stream at `GET /api/v1/events/sse`.
  * Added authenticated event capability discovery at `GET /api/v1/events/capability`.
  * SSE streams now provide an initial connection message, filtered event delivery, and periodic keepalive pings.
  * Added role-based channel filtering and authorization for streamed events.

* **Bug Fixes**
  * Unauthenticated requests to event endpoints are rejected.
  * Invalid or unauthorized channel selections are denied by default.

* **Tests**
  * Added coverage for authentication, filtering, streaming behavior, headers, and capability responses.

### DIFF
--- a/docs/superpowers/plans/2026-07-31-events-sse-stream.md
+++ b/docs/superpowers/plans/2026-07-31-events-sse-stream.md
@@ -1,0 +1,428 @@
+# Events SSE Stream Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a read-only Server-Sent Events transport over the existing events hub at `GET /api/v1/events/sse`, so external integrations can observe realtime events without implementing the WebSocket subscribe handshake.
+
+**Architecture:** A new handler method on the existing `EventsHandler`, reusing `events.Hub.Subscribe()`, `allowedChannelsForRole` and `allowsEventForClaims`. Channel selection is a query parameter rather than a handshake. The WebSocket handler, the hub, and event publishing are not modified.
+
+**Tech Stack:** Go, chi router, `net/http` with `http.Flusher`, existing `internal/events` package, `httptest` for tests.
+
+## Global Constraints
+
+- Base branch is `feat/events-sse-stream`, already created from `upstream/main` (d70b291b). Do not create branches or worktrees.
+- **Do not modify** `internal/events/hub.go`, `internal/events/types.go`, `HandleWebSocket`, or anything under `internal/api/handlers/events_publish.go`. This adds a transport; it changes no existing behaviour.
+- Existing tests in `events_ws_test.go` and `events_ws_user_settings_test.go` must continue to pass untouched.
+- New DB changes: none. This feature adds no migration.
+- Go stays `gofmt`/`goimports` clean.
+- Verify with `make lint` and `make test-go`. CI runs `golangci-lint --new-from-merge-base`, so only lines this branch touches must be lint-clean; pre-existing findings elsewhere are not yours.
+- Conventional Commit subjects. Commit after every task.
+- Profile-scoped notification channels are OUT OF SCOPE — they depend on the ws-ticket binding which SSE does not use.
+
+---
+
+### Task 1: SSE handler with hello frame and channel filtering
+
+**Files:**
+
+- Create: `internal/api/handlers/events_sse.go`
+- Create: `internal/api/handlers/events_sse_test.go`
+- Modify: `internal/api/router.go` (beside the existing `r.Get("/events/ws", …)` registration, around line 1998)
+
+**Interfaces:**
+
+- Consumes: `EventsHandler` (existing struct), `h.hub.Subscribe()`, `allowedChannelsForRole(role string) []evt.EventChannel`, `allowsEventForClaims(claims *auth.Claims, boundProfileID string, env evt.Envelope) bool`, `apimw.GetClaims(ctx)`.
+- Produces: `func (h *EventsHandler) HandleSSE(w http.ResponseWriter, r *http.Request)`, registered at `GET /api/v1/events/sse`.
+
+Read `internal/api/handlers/events_ws.go` first, particularly `HandleWebSocket` (nil/claims guards, hub subscribe, hello frame, the event loop with its channel-subscription and `allowsEventForClaims` checks). The SSE handler mirrors that structure with an SSE writer in place of the WebSocket connection, and no inbound message handling.
+
+- [ ] **Step 1: Write the failing test**
+
+Create `internal/api/handlers/events_sse_test.go`. Follow the conventions already in `events_ws_test.go` (same package, `httptest`, fake collaborators). Model the claims/context setup on how the existing WebSocket tests build an authenticated request; read that file and reuse its helpers rather than inventing new ones.
+
+```go
+package handlers
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	evt "github.com/Silo-Server/silo-server/internal/events"
+)
+
+func TestHandleSSESetsStreamingHeaders(t *testing.T) {
+	hub := evt.NewHub()
+	h := &EventsHandler{hub: hub}
+
+	req := newAuthedSSERequest(t, "/api/v1/events/sse?channels=sessions")
+	rec := httptest.NewRecorder()
+
+	runSSEHandlerBriefly(t, h, rec, req)
+
+	if got := rec.Header().Get("Content-Type"); got != "text/event-stream" {
+		t.Fatalf("Content-Type = %q, want text/event-stream", got)
+	}
+	if got := rec.Header().Get("Cache-Control"); got != "no-cache" {
+		t.Fatalf("Cache-Control = %q, want no-cache", got)
+	}
+	// openresty fronts several deployments and buffers the stream without this.
+	if got := rec.Header().Get("X-Accel-Buffering"); got != "no" {
+		t.Fatalf("X-Accel-Buffering = %q, want no", got)
+	}
+}
+
+func TestHandleSSEWritesHelloFrameFirst(t *testing.T) {
+	hub := evt.NewHub()
+	h := &EventsHandler{hub: hub}
+
+	req := newAuthedSSERequest(t, "/api/v1/events/sse?channels=sessions")
+	rec := httptest.NewRecorder()
+
+	runSSEHandlerBriefly(t, h, rec, req)
+
+	body := rec.Body.String()
+	if !strings.Contains(body, "event: hello") {
+		t.Fatalf("missing hello event, body = %q", body)
+	}
+	if !strings.Contains(body, `"required_action":"none"`) {
+		t.Fatalf("hello should not demand a subscribe action, body = %q", body)
+	}
+}
+
+func TestHandleSSEWritesSubscribedChannelEvent(t *testing.T) {
+	hub := evt.NewHub()
+	h := &EventsHandler{hub: hub}
+
+	req := newAuthedSSERequest(t, "/api/v1/events/sse?channels=sessions")
+	rec := httptest.NewRecorder()
+
+	body := runSSEHandlerWithPublish(t, h, hub, rec, req, evt.Envelope{
+		Channel: evt.ChannelSessions,
+		Event:   "sessions.replaced",
+	})
+
+	if !strings.Contains(body, "event: sessions") {
+		t.Fatalf("subscribed channel event was not written, body = %q", body)
+	}
+}
+
+func TestHandleSSESuppressesUnsubscribedChannel(t *testing.T) {
+	hub := evt.NewHub()
+	h := &EventsHandler{hub: hub}
+
+	req := newAuthedSSERequest(t, "/api/v1/events/sse?channels=sessions")
+	rec := httptest.NewRecorder()
+
+	body := runSSEHandlerWithPublish(t, h, hub, rec, req, evt.Envelope{
+		Channel: evt.ChannelJobs,
+		Event:   "jobs.updated",
+	})
+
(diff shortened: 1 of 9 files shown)


RXWatcher:
Thanks — all four findings were valid and are fixed in `6f0ccd60`.

**Major, per-frame write deadline.** Correct, and it caught a real trade I made badly. Clearing the connection deadline was necessary because the shared server sets `WriteTimeout: 120s`, which Go applies once at request read and never renews — without clearing it every stream dies at ~120s. But clearing it outright meant a stalled-but-not-closed client could block `w.Write` indefinitely and hold its hub subscription open. Now the initial clear stays (it is what stops the server-level deadline governing the whole connection) and `writeSSEFrame` renews a bounded deadline before each hello, ping and event write. Bound is `ssePingInterval` (25s): a healthy write completes in microseconds, and a peer that cannot absorb one frame within a ping cycle is not keeping up, so it is dropped rather than held forever.

**Test race.** Real, and it would have flaked in CI rather than locally. The publish helper now runs the handler in a goroutine and waits on a flush signal — the hello frame is written after `hub.Subscribe()` returns, so first flush proves subscription — then publishes. No `time.Sleep`. Verified with `go test -count=10 -run TestHandleSSE`: 90/90 sub-runs pass, no flakes.

**Schema assertion.** The sharpest of the three. That test only checked for `event: sessions`, so it would have passed even if the handler regressed to serializing the raw `evt.Envelope` — which is precisely the defect an earlier review round found and fixed, meaning the guard could not detect its own regression. It now decodes the `data:` payload, asserts the public `EventsEventMessage` shape, and asserts `source_id`, `admin_only`, `target_plugin_id`, `user_id` and `profile_id` are absent — publishing an envelope with those fields populated so removing the projection genuinely fails the test.

**Doc drift.** Fixed. The spec's ping example showed a `ts` field the implementation never sent, and the capability section still described extending an existing endpoint rather than the dedicated `/api/v1/events/capability` that shipped. Both corrected to match the code, along with the corresponding plan section.

Verification after the fixes:

```
go test -count=1 ./internal/api/handlers/ ./internal/events/   ok (both)
go test -count=10 -run TestHandleSSE ./internal/api/handlers/  ok, no flakes
golangci-lint run --new-from-merge-base=upstream/main ./...    0 issues
make verify-local-paths                                        pass
```

One note for transparency: the manual smoke test in the plan — confirming a real stream survives past the 120s mark against a live server — has not been run yet. Everything above is unit-level. I will report the result rather than quietly leave it unstated.

RXWatcher:
Fixed the race in `47a0f787`; skipping the MD046 finding with reasoning below.

**Residual race — valid, and the second half was the important part.**

My earlier fix closed the publish-before-subscribe race but left a cancel-before-process one: the helper still cancelled on a fixed 30ms timer, so under contention `HandleSSE` could observe `ctx.Done()` before writing the published event.

The delivery test failing intermittently would at least have been visible. The suppression tests were the real problem — they assert *absence*, so a handler cancelled before processing anything produced an empty body, the assertion passed, and the test went green having proven nothing. That includes the non-admin test, which is the endpoint's core authorization guarantee. A test that passes because the code never ran is worse than no test.

Both shapes now wait on real frame counts rather than wall-clock:
- delivery tests wait for the second flush (first is hello, second is the event), so the wait is causally tied to what is asserted;
- suppression tests publish the suppressed event, then a *subscribed sentinel*, wait for the sentinel's frame, and only then assert the suppressed event is absent — proving the handler ran past it rather than never reaching it.

```
go test -count=20 -run TestHandleSSE ./internal/api/handlers/   ok, 140/140 sub-tests
go test -race -count=5 -run TestHandleSSE ./internal/api/handlers/   no data races
golangci-lint run --new-from-merge-base=upstream/main ./...     0 issues
```

**MD046 on the spec's JSON block — respectfully skipping.**

I checked the premise against this repository before deciding:
- no markdownlint configuration exists (`.markdownlint*`, `package.json`, `Makefile`, `.github/` are all clean);
- markdown is not linted in CI;
- fenced code blocks are the established convention — `docs/ai-contributions.md` uses four, and `docs/architecture/*.md` use them throughout.

Converting this one block to indented would make the document inconsistent with every other doc in the repo in order to satisfy a linter the project does not run. Happy to change it if markdownlint is something you intend to adopt.

Quick104:
Thanks for this — the implementation quality is high, and the adversarial-review section caught three real defects (the `WriteTimeout` deadline in particular is a genuinely non-obvious failure mode that unit tests could not have surfaced). I'm closing it, along with #523, but the reasoning is about direction rather than execution.

## Why not a second transport

**Replacing WebSocket isn't available to us.** Four WS endpoints exist and two are genuinely bidirectional — `session_ws.go:124` handles inbound `hello`/`ack`/`result` for remote playback control, and `watch_together.go:741` reads client frames. Neither can be SSE. So dropping `/events/ws` doesn't converge on one transport; it leaves the events hub as the odd one out among three.

It also isn't free to remove: `silo-android` ships against it today (`HomeRealtimeClient.kt:86`, `NotificationsRealtimeClient.kt`), and `silo-apple` is mid-implementation against it (`docs/realtime-updates-spec.md:68,190`). Under the `/api/v1` additive-only rule a removal would need an entry in the pre-lock table in `docs/architecture/v1-scope.md`.

**Keeping both has a maintenance cost this diff already demonstrates.** The SSE path deliberately omits the WS path's snapshot-refetch-on-empty-`data` enrichment (`events_ws.go:509-521`). Every publisher on the `sessions` channel sends `nil` data — `worker/cleanup.go:180`, `worker/reconciler.go:223`, `handlers/playback_admin_store.go:99` — so `?channels=sessions`, the headline example, delivers `data: null` over SSE while the same channel over WS delivers the full session list. Two more divergences (raw envelope on the wire, missing authorization test) were caught in review before submission. Two transports over one hub is only sustainable if a single code path decides what a subscriber sees; this is a parallel one.

**`Last-Event-ID` is the thing SSE would genuinely add, and it isn't here.** There's no `Last-Event-ID` handling in the tree. You get `EventSource`'s automatic reconnect, but events published during the gap are dropped silently — arguably worse than a WS client that at least knows it reconnected. Resume would need hub-side buffering that doesn't exist.

**HTTP/1.1 connection budget.** `cmd/silo/main.go:2709` is plain `ListenAndServe` with no h2c, so server-side is HTTP/1.1. Each `EventSource` holds one of the browser's ~6 connections per origin permanently; WebSockets don't count against that cap. The web app can already have events, playback control, and watch-together open simultaneously. A TLS-terminating proxy speaking HTTP/2 hides this; a direct LAN or Docker connection doesn't.

## On the battery concern

The battery argument for SSE doesn't hold up here, and I want to be specific rather than hand-wave it.

Keepalive traffic is server-driven and near-identical. `ws_shared.go:15` sets `wsPingInterval = 20s`; this PR sets `ssePingInterval = 25s`. A WS ping frame is a few bytes of binary framing; the SSE ping is `event: ping\ndata: {"type":"ping"}\n\n`, roughly 35 bytes. SSE is marginally *more* wire traffic per keepalive, less often. The one real asymmetry is that WS costs the client a pong write every 20s, but the radio is already up for the server's ping — that isn't a measurable delta.

More to the point, both clients already solved this without changing transport. `silo-apple/docs/realtime-updates-spec.md:230`: `.active` → `client.start()`, `.background` → `client.stop()`. `HomeRealtimeCoordinator.kt:25-26`: "connected only while the app is foregrounded." The Apple spec's own risk table (line 322) reaches the same conclusion: *"Socket only while foregrounded; idle traffic is ping/pong + occasional small frames. Acceptable."*

Neither OS permits holding either kind of socket open in the background — iOS suspends the process, Android Doze does the equivalent — so the drain scenario can't occur with WS and wouldn't be fixed by SSE. Background delivery goes through APNs/FCM via the push relay, which is the actual answer. If there's a measurement pointing the other way I'd genuinely like to see it, since it would contradict the clients' own design notes.

## What we're doing instead

Everything #523 actually complains about is contract, not transport. `/events/ws` is barely bidirectional as it is — the only inbound message type it accepts is `subscribe` (`events_ws.go:245`). So the friction is removable in place:

- accept `?channels=catalog,jobs` on connect and skip the handshake entirely when present;
- don't arm the 5-second policy-violation close (`events_ws.go:169-190`) for connections that already declared their channels;
- report a forbidden or unknown channel in the existing `rejected` array instead of closing the connection over it.

That removes the handshake, the deadline, and the silent-close behavior — the substance of the issue — without a second transport to keep in sync. The `subscribe` frame stays as the optional path, since the web client genuinely uses it to ref-count channels and re-subscribe live as components mount (`RealtimeEventsProvider.tsx:482-501`), which SSE can't do without a reconnect.

A follow-up PR implementing that is coming; I'll link it here.

What that leaves genuinely unaddressed is `curl -N` debuggability and not needing a WS library in an external consumer. Both are real, and if an external consumer actually turns up, the right shape is to extract subscribe → role filter → claims filter → payload enrichment into one shared core with WS and SSE as thin encoders over it — a projection, not a sibling. Worth reopening #523 at that point.

Sorry to close work this carefully done. The diagnosis in #523 was correct; it's the remedy I'd rather take a different way.

---

*AI disclosure: this review was written with Claude Code (claude-opus-5), fully AI-generated, verified against the repository and the `silo-android` / `silo-apple` checkouts at the file references cited.*

Quick104:
Follow-up is open: #525 — removes the handshake requirement, the grace period, and the connection-killing channel rejection from `/events/ws` directly.

PR #525, opened 2026-07-31
feat(events): let websocket clients declare channels on connect

Quick104:
Part of #523. Follows from closing #524 — [reasoning there](https://github.com/Silo-Server/silo-server/pull/524#issuecomment-5143655226).

## Problem

Observing the events hub over `GET /api/v1/events/ws` costs more than it should. A read-only consumer must send a `subscribe` frame within five seconds or be closed with a policy violation, so it has to implement the handshake and hold the write half of the socket open purely to satisfy it.

#523 proposed solving this with a second transport. But the cost is contract, not transport: `subscribe` is the only inbound message type this endpoint accepts. The friction is removable in place, without a parallel delivery path to keep in sync and without stranding the Android and Apple clients that ship against this endpoint.

## Approach

Accept the selection on the URL. `GET /api/v1/events/ws?channels=catalog,user_state` subscribes on connect, answers with the same `subscribed` frame and per-channel snapshot frames the handshake produces, and is not put on the grace-period clock.

An observer can now connect and never write to the socket again.

Two supporting changes:

- **One shared resolver.** Channel selection resolves through `resolveChannelSelection` for both the URL and the handshake, so the two cannot drift on who may subscribe to what. Role checks, profile-binding, validity, and dedup are moved verbatim, not rewritten.
- **Unknown channels are rejected, not fatal.** An unrecognized name is reported in the existing `rejected` array as `unknown_channel` instead of closing the connection. Closing took down every other channel the client held over one bad name, and a client cannot always know which channels its role allows before it asks. `forbidden` and `profile_required` were already handled this way; this makes the third case consistent.

The `subscribe` frame stays fully supported. The web client depends on it to ref-count channels and re-subscribe live as components mount (`RealtimeEventsProvider.tsx:482-501`), which a connect-time-only selection cannot express.

## What the grace period is keyed on

The clock is disarmed by **holding a subscription**, not by having spelled `?channels=`.

This is the correction that matters most from review. The first cut disarmed on `declared`, which meant `?channels=` (empty) and a non-admin's `?channels=sessions` both came up subscribed to nothing and were never reaped — each still holding a hub subscriber, two goroutines, and an envelope channel that every published event fans into. Those are exactly the connections the grace period exists to close.

Selection now resolves *before* the hello frame, so `required_action` reports what the connection actually owes: `"none"` only when at least one channel resolved, `"subscribe"` otherwise, including for a declaration where every entry was refused. Resolution is pure — no query, no I/O — so it cannot stall ahead of the reader goroutine; only the snapshots that accepted channels produce can, and those still run after it.

## Bounding the answer

Making an unknown channel non-fatal removed a brake that was doing real work: pre-change, one bad name closed the connection, so a bad selection was self-limiting. Now every refusal is *answered*, and each refusal quotes the name it refuses — so a large `?channels=` of distinct garbage names produced a much larger `subscribed` frame than the request, buffered server-side. A 900 KB query of ~180k unknown names produced a ~13 MB frame.

Four bounds, so the answer cannot scale with the request:

- At most **32** distinct channels are considered per selection. There are ten client channels; the slack exists only so a caller learns it overshot rather than silently losing channels.
- The overrun is reported as **one** `too_many_channels` entry, not one per excess name.
- An echoed channel name is truncated at **64** bytes — one enormous name is as good an amplifier as many small ones.
- A **64 KiB** read limit on the socket, so an oversize frame cannot be buffered whole before it is rejected. This is the one case that stays fatal, and has to be: a frame is buffered before its type can be read, so refusing it politely would mean first doing the thing the limit prevents.

`max_requested_channels` is published on the capability endpoint.

## Repeated query parameters

`?channels=catalog&channels=user_settings` honored only the first occurrence and returned an **empty** `rejected` array — the connection came up subscribed to less than it asked for and reported nothing wrong. Every occurrence is now read; repeating the parameter and comma-separating are equivalent spellings.

## The capability endpoint and the plugins channel

The endpoint advertised `evt.AllChannels`, which includes `plugins` — host-to-plugin runtime dispatch (`RuntimeHostServer.PublishEventTo`), granted to no role, admin included. An admin following the endpoint's own stated purpose would request it and be told `forbidden` / "Admin access required" while already being an admin, with no way forward.

`evt.ClientChannels` is now the subscribable set (AllChannels minus `plugins`), the capability endpoint advertises that, and `plugins` resolves as `unknown_channel` for every role. The `forbidden` message is reworded to "This channel is not available to this connection" — the hardcoded admin remedy is what turned a soft failure into a dead end, since a client acting on it loops forever.

`TestEventsCapabilityAdvertisesOnlySubscribableChannels` pins the invariant directly: it feeds the endpoint's own channel list to the resolver as the most permissive caller possible (admin, profile-bound) and fails if anything is rejected.

## Compatibility

Additive under the v1 rules — no field changes type, no field disappears, no status code is repurposed. `max_requested_channels` is a new field on a capability endpoint added in this same PR.

Observable changes on the existing handshake path, all strictly more permissive or previously unreachable:

- An unknown channel name previously closed the connection and now returns a `rejected` entry. No client can have depended on the close.
- The `forbidden` rejection's `message` text changed. `code` is unchanged, and it is the code clients branch on.
- `plugins` in a subscribe frame now answers `unknown_channel` instead of `forbidden`. It was never grantable to anyone, so no client could have been subscribed to it.
- A selection naming more than 32 channels is now partially refused. The maximum real selection is ten.

`required_action` in the hello frame is `"none"` for a connection holding a declared subscription, `"subscribe"` otherwise. `web/src/api/types.ts` typed it as the literal `"subscribe"` and is widened to the union — the web client never reads the field, so nothing changes at runtime. `silo-android` decodes it as a plain `String` (`NotificationModels.kt:210`), so it is unaffected.

No client change is required by this PR. Both client repos can adopt `?channels=` when convenient; the Apple realtime spec's handshake description (`silo-apple/docs/realtime-updates-spec.md:72-73`) stays accurate for the path it documents.

## Verification

Actual output.

```
$ go test ./internal/api/handlers/ -run "TestEventsWebSocket|TestParseDeclaredChannels|TestResolveChannelSelection|TestEventsCapability" -v
--- PASS: TestEventsCapabilityReportsDeclaredChannelSupport (0.00s)
--- PASS: TestEventsCapabilityAdvertisesOnlySubscribableChannels (0.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsSkipHandshake (0.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsSurviveGracePeriod (6.00s)
--- PASS: TestEventsWebSocketEmptyDeclarationStillClosed (10.01s)
    --- PASS: TestEventsWebSocketEmptyDeclarationStillClosed/declares_no_channels (5.01s)
    --- PASS: TestEventsWebSocketEmptyDeclarationStillClosed/every_declared_channel_refused (5.00s)
--- PASS: TestEventsWebSocketPartialDeclarationSurvives (6.00s)
--- PASS: TestEventsWebSocketRepeatedChannelsParameter (0.00s)
--- PASS: TestEventsWebSocketSilentConnectionStillClosed (5.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsCannotEscalate (0.00s)
--- PASS: TestEventsWebSocketUnknownChannelDoesNotCloseConnection (0.00s)
--- PASS: TestEventsWebSocketRejectsOversizeFrame (0.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsSurviveSlowSnapshot (32.00s)
--- PASS: TestParseDeclaredChannels (0.00s)
    --- PASS: TestParseDeclaredChannels/absent_parameter_keeps_the_handshake (0.00s)
    --- PASS: TestParseDeclaredChannels/empty_value_declares_nothing,_but_still_declares (0.00s)
    --- PASS: TestParseDeclaredChannels/whitespace_and_empty_entries_are_dropped (0.00s)
    --- PASS: TestParseDeclaredChannels/every_occurrence_of_the_parameter_is_read (0.00s)
    --- PASS: TestParseDeclaredChannels/a_repeated_parameter_with_only_blank_values_still_declares (0.00s)
--- PASS: TestResolveChannelSelectionDeniesByDefault (0.00s)
--- PASS: TestResolveChannelSelectionDeduplicates (0.00s)
--- PASS: TestResolveChannelSelectionDeduplicatesRejections (0.00s)
--- PASS: TestResolveChannelSelectionBoundsTheAnswer (0.00s)
--- PASS: TestResolveChannelSelectionTruncatesLongNames (0.00s)
--- PASS: TestResolveChannelSelectionRefusesPluginsChannel (0.00s)
    --- PASS: TestResolveChannelSelectionRefusesPluginsChannel/user (0.00s)
    --- PASS: TestResolveChannelSelectionRefusesPluginsChannel/admin (0.00s)
--- PASS: TestEventsWebSocketDeliversUserSettingsToNonAdmins (0.00s)
ok  	github.com/Silo-Server/silo-server/internal/api/handlers	59.128s

$ go test ./...          # with a stub web/dist, see note below
(all packages ok)

$ golangci-lint run --new-from-merge-base=origin/main ./internal/api/handlers/... ./internal/events/...
0 issues.

$ make verify-local-paths
(pass)

$ cd web && pnpm exec tsc -b
(exit 0)

$ cd web && pnpm run lint
✖ 157 problems (0 errors, 157 warnings)
```

The 157 web warnings are pre-existing and all in `web/vendor/foliate-js`.

`go test ./...` fails setup in `cmd/silo` and `web` with `pattern all:dist: no matching files found` unless the frontend has been built. I confirmed that failure is identical on a clean checkout with no changes applied, then re-ran the suite with a stub `web/dist/index.html` present — everything passes. The stub was removed afterwards; the tree is clean.

The websocket tests go through a real socket rather than calling the handler directly, because what a connection is subscribed to before it has said anything only exists in the connection loop. The ones that wait out the real five-second grace period are skipped under `-short`.

**Every guard was verified to bite** by reverting it and watching the test fail:

| Reverted to | Failing test | Output |
| --- | --- | --- |
| role check removed from `resolveChannelSelection` | `…CannotEscalate` | `non-admin was granted the sessions channel: ["sessions","user_settings"]` |
| `if !declared` for the grace period | `…EmptyDeclarationStillClosed` | both subtests time out waiting for the close that never comes |
| `query.Get("channels")` (first occurrence only) | `…RepeatedChannelsParameter` | `channel "user_settings" was dropped: ["catalog"]` |
| `maxRequestedChannels`/`maxChannelNameLength` raised to 1<<30 | `…BoundsTheAnswer`, `…TruncatesLongNames` | `rejected 5000 entries`; `echoed name is 4096 bytes` |
| capability advertising `evt.AllChannels` | `…AdvertisesOnlySubscribableChannels` | `capability advertises channels an admin cannot subscribe to: [{plugins unknown_channel Unknown channel}]` |
| read limit raised to 64 MiB | (temporary probe, removed) | a 200 KB frame was accepted |

## Risks

- **Privilege escalation via the query string** is the risk worth scrutinizing — a caller naming an admin-only channel on the URL must not receive it. The URL path runs the identical role check as the handshake because both call one function, and the test above covers it end to end, including that an admin-only event published to a refused channel is not delivered.
- **Notifications remain handshake-only in practice.** They require a profile-bound ws-ticket, which is unchanged; declaring `channels=notifications` on an unbound connection yields `profile_required`, as sending it in a subscribe frame already did.
- **Backpressure, hub fan-out, and per-event authorization are untouched.**

## What this does not address

`curl -N` debuggability, and not needing a WebSocket library in an external consumer. Both are real and neither is solved here. If an external consumer materializes, #523 is worth reopening — with SSE built as a projection over the shared resolver rather than a parallel handler.

## AI Disclosure

- **Tool(s):** Claude Code
- **Model(s):** claude-opus-5
- **Involvement:** fully AI-generated
- **Verification:** all commands above were run and their real output pasted. Each authorization and bounding guard was validated by deliberately reverting it and confirming the corresponding test fails — the table above lists what each one printed.

## Summary by CodeRabbit

* **New Features**
  * WebSocket connections can request channels directly through the connection URL.
  * Declared channels receive acknowledgements and initial snapshots without a separate subscription step.
  * The handshake indicates whether a subscription action is required.
  * Invalid or unauthorized channel requests are reported individually without closing the connection.
  * Added an events capability endpoint describing supported channels and connection features.

* **Bug Fixes**
  * Preserved the timeout for connections without an effective subscription.
  * Limited incoming WebSocket frames to 64 KiB.

### DIFF
--- a/internal/api/handlers/events_capability.go
+++ b/internal/api/handlers/events_capability.go
@@ -1,0 +1,50 @@
+package handlers
+
+import (
+	"encoding/json"
+	"net/http"
+
+	evt "github.com/Silo-Server/silo-server/internal/events"
+)
+
+// eventsCapabilityResponse describes how a client may subscribe to the events
+// websocket, so it can pick a connection flow without version-sniffing.
+//
+// The field that matters is DeclaredChannels. A server predating that support
+// ignores ?channels=, answers required_action:"subscribe", and closes the
+// connection after the grace period — so a client that cannot send a subscribe
+// frame has no safe way to discover the difference from the socket itself.
+type eventsCapabilityResponse struct {
+	SchemaVersion int `json:"schema_version"`
+	// SubscribeFrame reports the handshake: connect, then send a subscribe
+	// frame. Always true; named so a future removal is detectable rather than
+	// silent.
+	SubscribeFrame bool `json:"subscribe_frame"`
+	// DeclaredChannels reports that ?channels= is honored on connect, that such
+	// a connection is exempt from the subscribe grace period, and that its
+	// hello frame carries required_action:"none".
+	DeclaredChannels bool `json:"declared_channels"`
+	// SubscribeGracePeriodSeconds is how long a connection that declared no
+	// channels may stay silent before it is closed. 0 would mean no deadline.
+	SubscribeGracePeriodSeconds int `json:"subscribe_grace_period_seconds"`
+	// Channels is every channel this server knows, independent of role. What
+	// the caller may actually subscribe to arrives as available_channels in the
+	// hello frame, which is role-filtered.
+	Channels []evt.EventChannel `json:"channels"`
+}
+
+// HandleCapability reports the events websocket's subscription capabilities.
+//
+// Per the v1 rules, new functionality is feature-detected rather than inferred
+// from a version. This follows the existing per-subsystem convention
+// (/notifications/capability, /playback/capability, /downloads/capability).
+func (h *EventsHandler) HandleCapability(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(eventsCapabilityResponse{
+		SchemaVersion:               1,
+		SubscribeFrame:              true,
+		DeclaredChannels:            true,
+		SubscribeGracePeriodSeconds: int(subscribeGracePeriod.Seconds()),
+		Channels:                    evt.AllChannels,
+	})
+}
--- a/internal/api/handlers/events_capability.go
+++ b/internal/api/handlers/events_capability.go
@@ -27,10 +27,18 @@ type eventsCapabilityResponse struct {
 	// a connection is exempt from the subscribe grace period, and that its
 	// hello frame carries required_action:"none".
 	DeclaredChannels bool `json:"declared_channels"`
-	// SubscribeGracePeriodSeconds is how long a connection that declared no
-	// channels may stay silent before it is closed. 0 would mean no deadline.
+	// SubscribeGracePeriodSeconds is how long a connection holding no
+	// subscription may stay silent before it is closed. 0 would mean no
+	// deadline.
 	SubscribeGracePeriodSeconds int `json:"subscribe_grace_period_seconds"`
-	// Channels is every channel this server knows, independent of role. What
+	// MaxRequestedChannels is the most channels one selection may name, on the
+	// URL or in a subscribe frame. Names past it are answered with a single
+	// too_many_channels rejection rather than one per name.
+	MaxRequestedChannels int `json:"max_requested_channels"`
+	// Channels is every channel a client may ask for on this server,
+	// independent of role — not every channel the server has: the plugins
+	// channel is host-to-plugin runtime dispatch and is granted to no role, so
+	// naming it here would advertise a request that can only be refused. What
 	// the caller may actually subscribe to arrives as available_channels in the
 	// hello frame, which is role-filtered.
 	Channels []evt.EventChannel `json:"channels"`
@@ -52,6 +60,7 @@ func (h *EventsHandler) HandleCapability(w http.ResponseWriter, r *http.Request)
 		SubscribeFrame:              true,
 		DeclaredChannels:            true,
 		SubscribeGracePeriodSeconds: int(subscribeGracePeriod.Seconds()),
-		Channels:                    evt.AllChannels,
+		MaxRequestedChannels:        maxRequestedChannels,
+		Channels:                    evt.ClientChannels,
 	})
 }
--- a/internal/api/handlers/events_capability.go
+++ b/internal/api/handlers/events_capability.go
@@ -8,12 +8,15 @@ import (
 )
 
 // eventsCapabilityResponse describes how a client may subscribe to the events
-// websocket, so it can pick a connection flow without version-sniffing.
+// websocket.
 //
-// The field that matters is DeclaredChannels. A server predating that support
-// ignores ?channels=, answers required_action:"subscribe", and closes the
-// connection after the grace period — so a client that cannot send a subscribe
-// frame has no safe way to discover the difference from the socket itself.
+// Clients are expected to run a current build rather than negotiate down to an
+// old server, so this is not a branch-on-capability contract. It is how a
+// client tells the difference between "this server does not do that" and "the
+// connection failed", which is what lets it say the deployment is out of date
+// instead of failing opaquely: a server predating declared channels ignores
+// ?channels=, answers required_action:"subscribe", and closes the connection
+// after the grace period, which is indistinguishable from a broken socket.
 type eventsCapabilityResponse struct {
 	SchemaVersion int `json:"schema_version"`
 	// SubscribeFrame reports the handshake: connect, then send a subscribe
@@ -38,6 +41,10 @@ type eventsCapabilityResponse struct {
 // Per the v1 rules, new functionality is feature-detected rather than inferred
 // from a version. This follows the existing per-subsystem convention
 // (/notifications/capability, /playback/capability, /downloads/capability).
+//
+// A client that finds declared_channels false is talking to a server older than
+// its own expectations; the useful response is to tell the user to update the
+// server, not to silently fall back.
 func (h *EventsHandler) HandleCapability(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	_ = json.NewEncoder(w).Encode(eventsCapabilityResponse{
--- a/internal/api/handlers/events_capability_test.go
+++ b/internal/api/handlers/events_capability_test.go
@@ -42,7 +42,39 @@ func TestEventsCapabilityReportsDeclaredChannelSupport(t *testing.T) {
 		t.Errorf("subscribe_grace_period_seconds = %d, want %d", got.SubscribeGracePeriodSeconds, want)
 	}
 
-	if len(got.Channels) != len(evt.AllChannels) {
-		t.Errorf("channels = %v, want all %d channels", got.Channels, len(evt.AllChannels))
+	if got.MaxRequestedChannels != maxRequestedChannels {
+		t.Errorf("max_requested_channels = %d, want %d", got.MaxRequestedChannels, maxRequestedChannels)
+	}
+
+	if len(got.Channels) != len(evt.ClientChannels) {
+		t.Errorf("channels = %v, want all %d client channels", got.Channels, len(evt.ClientChannels))
+	}
+}
+
+// TestEventsCapabilityAdvertisesOnlySubscribableChannels is the point of
+// publishing the list at all: a client that requests everything the endpoint
+// names must not be refused any of it. The plugins channel is the case — it is
+// in evt.AllChannels but granted to no role, admin included, so advertising it
+// would send a client to a request that can only fail.
+func TestEventsCapabilityAdvertisesOnlySubscribableChannels(t *testing.T) {
+	handler := &EventsHandler{}
+	rec := httptest.NewRecorder()
+	handler.HandleCapability(rec, httptest.NewRequest(http.MethodGet, "/events/capability", nil))
+
+	var got eventsCapabilityResponse
+	if err := json.Unmarshal(rec.Body.Bytes(), &got); err != nil {
+		t.Fatalf("decoding capability: %v (%s)", err, rec.Body.String())
+	}
+
+	// An admin, on a profile-bound connection, is the most permissive caller
+	// there is. Every advertised channel must resolve for them.
+	_, accepted, rejected := resolveChannelSelection(
+		got.Channels, allowedChannelsForRole("admin"), "profile-1")
+
+	if len(rejected) != 0 {
+		t.Errorf("capability advertises channels an admin cannot subscribe to: %v", rejected)
+	}
+	if len(accepted) != len(got.Channels) {
+		t.Errorf("accepted %d of %d advertised channels", len(accepted), len(got.Channels))
 	}
 }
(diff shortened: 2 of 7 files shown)


Quick104:
All three findings verified against the code and fixed in 787e1567.

**P2 — reader ordering. Correct, and a regression this PR introduced.** `configureWebSocket` (`ws_shared.go:54`) installs an absolute read deadline that only pongs extend, and gorilla processes pongs solely inside `advanceFrame` (`conn.go:950`), reached only from `NextReader`/`ReadMessage`. The declared path built its snapshots before starting the reader goroutine, so a snapshot slower than the 30s deadline would kill an otherwise healthy connection the instant reading began. The handshake path never had this problem because its snapshots run inside `handleEventsClientMessage`, already downstream of an active reader — which is exactly why moving the work earlier broke it.

Fixed by starting the reader first. `TestEventsWebSocketDeclaredChannelsSurviveSlowSnapshot` stalls a tasks snapshot past `wsPingInterval + wsPongTimeout` and asserts the connection survives; I confirmed it fails with the previous ordering and passes with the fix.

**P1 — capability endpoint. Fair, and the client-facing consequence is the convincing part.** An older server ignores `?channels=`, answers `required_action:"subscribe"`, and closes after five seconds — so a client that cannot send a subscribe frame has no safe way to detect support from the socket itself, and would have to keep the handshake this feature exists to remove. Added `GET /api/v1/events/capability`, following the existing per-subsystem convention (`/notifications/capability`, `/playback/capability`, `/downloads/capability`). It reports both subscription modes, the grace period the handler actually enforces (asserted against the constant, so the two cannot drift), and the known channel list.

**CodeRabbit — duplicate rejections. Real, with one correction to the framing.** It reads as though this PR regressed dedup; it did not. The `seen` check sat after the rejection branches on `main` too, so `channels=sessions,sessions` produced two identical entries before this change as well. It is pre-existing, and cheap to correct in the function this PR extracted, so I fixed it rather than deferring: one `seen` set now guards both paths, and a channel named twice is answered once whether accepted or refused.

Verification after the fixes:

```
$ go test ./internal/api/handlers/ -run "TestEventsWebSocket|TestParseDeclaredChannels|TestResolveChannelSelection|TestEventsCapability" -v
--- PASS: TestEventsCapabilityReportsDeclaredChannelSupport (0.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsSkipHandshake (0.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsSurviveGracePeriod (6.00s)
--- PASS: TestEventsWebSocketSilentConnectionStillClosed (5.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsCannotEscalate (0.00s)
--- PASS: TestEventsWebSocketUnknownChannelDoesNotCloseConnection (0.00s)
--- PASS: TestEventsWebSocketDeclaredChannelsSurviveSlowSnapshot (32.00s)
--- PASS: TestParseDeclaredChannels (0.00s)
--- PASS: TestResolveChannelSelectionDeniesByDefault (0.00s)
--- PASS: TestResolveChannelSelectionDeduplicates (0.00s)
--- PASS: TestResolveChannelSelectionDeduplicatesRejections (0.00s)
--- PASS: TestEventsWebSocketDeliversUserSettingsToNonAdmins (0.00s)
ok      github.com/Silo-Server/silo-server/internal/api/handlers        43.108s

$ go test ./...        # empty web/dist present, see PR description
(all packages ok)

$ golangci-lint run --new-from-merge-base=origin/main ./internal/...
0 issues.

$ make verify-local-paths
(pass)
```

The capability endpoint is additive and mounted in the same authenticated group as `/events/ws`, so it inherits the same auth, rate limiting, and viewer-access middleware.

Quick104:
Updating the rationale for `/events/capability` — the endpoint stays, but not for the reason my earlier comment gave.

We're not planning to support clients negotiating down to an older server; the expectation is that people run a current build. So this is **not** a branch-on-capability contract, and no client should implement a fallback path against it.

What it's actually for: distinguishing *"this server doesn't do that"* from *"the connection failed."* Those are indistinguishable from the socket alone — an older server ignores `?channels=`, answers `required_action:"subscribe"`, and then closes on the grace period, which looks exactly like a broken connection. A client that can tell the difference can say **"your server is out of date"** instead of failing opaquely, which is a materially better experience for a self-hosted deployment where the user controls the update.

Comment-only change in bc04645c; no behavior or wire change, and the response shape is unchanged.

This also drops the version-skew argument from my earlier reply — that framing assumed graceful degradation, which isn't the plan.

Quick104:
Review pass applied in 2f9c3e36 — all five findings, plus a related one they surfaced.

**Rejections amplified the request** (`events_ws.go:205`). Correct, and the diagnosis of *why* was the useful part: making an unknown channel non-fatal removed the brake that used to close the connection on the first bad name, so a bad selection stopped being self-limiting. Four bounds now, since one alone is insufficient — 32 distinct channels per selection, the overrun reported as a single `too_many_channels` entry rather than one per name, an echoed name truncated at 64 bytes (one huge name amplifies as well as many small ones), and a 64 KiB read limit on the socket so an oversize frame is not buffered whole before it is rejected. `max_requested_channels` is published on the capability endpoint. Two tests, one of which asserts the encoded frame size directly so the bound holds whatever the constants become.

The read limit is the sixth item — not in the findings, but the same amplification one layer down: capping what `resolveChannelSelection` *answers* does nothing about what the socket *accepts* before the handler ever sees it. It stays fatal rather than answered, and has to be: a frame is buffered before its type can be read, so refusing it politely would mean first doing the thing the limit prevents.

**Grace period keyed on `declared`, not on being subscribed** (`events_ws.go:229`). Fixed as suggested, and the fix pulled something with it. Disarming on `len(subscriptions) == 0` makes `required_action: "none"` a lie in exactly the cases the finding names — the hello frame would tell a client it owes nothing while the connection is five seconds from a policy-violation close. Selection now resolves before the hello frame so `required_action` reports what is actually owed. That is safe against the reader-ordering constraint the PR already reasons about: resolution is pure, no query or I/O, so nothing moved ahead of the reader goroutine; only the snapshots accepted channels produce can stall, and those still run after it. Table test covers both named cases; a third test pins the boundary on the other side — one accepted channel among refusals is a live subscription and stays off the clock.

**`query.Get` honors only the first occurrence** (`events_ws.go:414`). Fixed. The empty `rejected` array was the sharp end of it — subscribed to less than asked for, with nothing to indicate it.

**Capability advertises `plugins`** (`events_capability.go:55`) **and the hardcoded admin remedy** (`events_ws.go:389`). Taken together, since the second is what makes the first unrecoverable. `evt.ClientChannels` is now the subscribable set — AllChannels minus `plugins`, which is host-to-plugin runtime dispatch and grantable to nobody — the capability endpoint advertises that, and `plugins` resolves as `unknown_channel` for every role. The `forbidden` message no longer promises admin access as the remedy.

Rather than assert the two lists match, `TestEventsCapabilityAdvertisesOnlySubscribableChannels` feeds the endpoint's own output to the resolver as the most permissive caller that exists (admin, profile-bound) and fails if anything comes back rejected. That is the actual invariant, and it holds for whatever channels get added later.

Every guard was verified to bite by reverting it and watching the test fail; the table in the PR body has what each one printed. Full `go test ./...` passes, `golangci-lint --new-from-merge-base` reports 0 issues, tsc and format:check clean, tree clean.

Issue #296, opened 2026-07-03
Subtitle enable mid-playback is slow: on-demand ffmpeg extraction with no caching; compat endpoint buffers whole file


zZebrahz:
## Summary

Enabling a subtitle track mid-playback takes noticeably long to show cues (web player and Apple clients), while disabling is instant on the web. Investigation shows the delay is **on-demand ffmpeg subtitle extraction with no caching**, and the Jellyfin-compat endpoint used by the Apple clients is the slowest possible variant (whole-file pass, fully buffered before the first byte).

Subtitle toggling never restarts the transcode (subtitles are out-of-band and client-rendered on both stacks — this is good architecture), so the fix is purely about extraction latency.

## Current behavior

Every subtitle enable spawns a fresh `ffmpeg` extraction. There is no extracted-subtitle cache anywhere.

**Native API** (`GET /stream/{session}/subtitles/{track}`, used by web):
- Text (SRT→VTT): windowed extraction with input-side `-ss` + `-t` (default 600s window) streamed with per-chunk flush — first cue ~1s. This path is fine.
- **ASS/SSA and PGS: windowing is disabled** (`windowable = !IsASS && !IsPGS`, `internal/playback/subtitle_stream.go:129-142`) — ASS needs the `[Script Info]` header at offset 0, PGS `.sup` is consumed whole by libpgs. Enabling these triggers a whole-file demux from byte 0, which takes many seconds on large files / network storage.

**Jellyfin-compat API** (`GET /Videos/{item}/{mediaSource}/Subtitles/{index}/stream.{format}`, used by Apple clients):
- `playback.ExtractSubtitle` / `ExtractSubtitleWithFormat` (`internal/playback/subtitles.go:58-132`) run `ffmpeg -i file -map 0:s:N` with **no `-ss`, no `-t`, and buffer the entire output into a `bytes.Buffer` before writing a single byte** — for every embedded track, every request. This is strictly worse than the native path.

Web "off is instant" is expected: disable is pure client-side teardown (abort fetch, clear cues); no server round trip.

## Why this matters

Plex toggles subtitles with only a momentary pause. Silo can beat that (no transcode restart needed), but today the enable path stalls on a cold whole-file ffmpeg pass on the two most common real-world cases: ASS subs on the web, and everything on Apple clients.

## Proposed fixes

[...]

PR #542, opened 2026-08-05
perf(playback): cache embedded text-subtitle extracts

### LINKED ISSUE #423: Embedded subtitle fetches stall up to 120s: the 600s extraction window is inert and truncations log as status=200
## Problem

**Turning on subtitles can stall for two full minutes, and on the web player it happens again every ten minutes.**

When a viewer picks a subtitle track that lives *inside* the video file (rather than as a separate `.srt` file next to it), the server has to dig the text out of the video before it can send it. On a big, high-bitrate film sitting on network storage, that digging is slow — and it is far slower than anyone realised.

What this looks like to people:

- **In the browser**, subtitles appear, play for about ten minutes, then freeze on the last line and stay frozen for up to two minutes before catching up. Then it happens again. And again. The player re-fetches subtitles roughly every ten minutes of playback, so a viewer of a two-hour film pays this toll a dozen or more times.
- **On Android/TV and Apple clients**, subtitles simply don't show up for up to ~40 seconds after you turn them on. Once they appear, they're fine for the rest of the film — those clients ask once and keep everything.
- **Subtitles that are separate files alongside the video are always instant** (all such requests finish in well under a second). So the problem is invisible to anyone testing with sidecar subtitles, and only bites on embedded tracks.

Measured over 41.6 hours of production traffic (2026-07-15 → 2026-07-17) on `/api/v1/stream/{session_id}/subtitles/{track}`:

| client | requests | p50 | p95 | max | ≥10s |
|---|---|---|---|---|---|
| browser (web player) | 220 | 72ms | **120,021ms** | 121,470ms | **45** |
| native (Android/TV) | 42 | 145ms | 9,566ms | 41,607ms | 2 |
| sidecar/external files | 118 | — | — | **all <1s** | 0 |

**The two-minute ceiling is not a coincidence, and it's worse than a stall.** 120s is the server's own connection write timeout. It cuts the response off mid-sentence. Because the server has already declared the request successful by the time it starts sending text, **a subtitle file that got chopped in half is recorded in the logs as a perfectly successful request**. Every dashboard we have says this endpoint is healthy. Viewers are getting truncated subtitle tracks and nothing anywhere reports an error.

**What previous reviews missed.**

The server has a feature that is supposed to prevent all of this: it only asks for a ten-minute slice of subtitles around wherever you are in the film, instead of the whole track. That feature does not work. **The instruction that limits the slice is passed to ffmpeg in a position where ffmpeg silently ignores it** — so every single request quietly reads from the seek point to the *end of the file*, no matter what limit is requested.

This survived two rounds of code review, on both sides, because everything *reads* correct:

- The code comment states plainly that it bounds how much of the file is read.
- The flag really is there in the command, exactly as the comment describes.
- The unit test asserts the flag is present — and it passes.
- The log line faithfully reports the window that was requested.

So reviewers checked intent against intent and found agreement. Nobody ran the command. Only actually executing ffmpeg against the production binary shows that asking for 30 seconds, asking for 300 seconds, and asking for nothing at all produce **byte-identical output**, all ending at the final cue of the film.

A related process failure is worth recording: an early investigation grepped the API request logs for the window parameters and concluded no client was sending them. **Those logs have no query-string field at all** — the greps could never have matched anything. The conclusion was drawn from a search that was structurally incapable of returning a result, and it pointed the investigation at the wrong client for a while.

**And the obvious fix is a trap.** The natural reaction is "just move the flag so the limit works." That would break subtitles on every non-browser client. The Apple and Android apps both fetch a subtitle track exactly once and keep the whole thing; they have no concept of re-fetching. They work today *because* the limit is broken and they accidentally receive complete tracks. Enabling the window as written would silently kill subtitles about ten minutes into every film on iOS, tvOS, macOS and Android — and the accidental whole-track behaviour has, in effect, become the contract those clients are written against.

Shrinking the window would be worse still: the web player hardcodes its ten-minute stride and advances blindly without checking what it actually received, so a server-side window smaller than ten minutes produces a subtitle blackout in the gap.

## Technical detail

- `streamExtractArgs` (`internal/playback/subtitle_stream.go:161-168`) passes `-t` as an **input** option. ffmpeg ignores it for these extracts. Only `-ss` takes effect, so cost is *(duration − seek) × bitrate* — which the logs confirm exactly: `seek=8s/130s/519s` → ≥60s; `seek=7797s/8392s/8987s` → <1s. Every 120s outlier is a ≥28 Mbps remux.
- Measured against the production binary (`ffmpeg 7.1.4-Jellyfin`, `/usr/lib/jellyfin-ffmpeg/ffmpeg` inside the `silo` container):

  | command | wall | bytes | last cue |
  |---|---|---|---|
  | `-ss 4000 -t 30 -i F` | 5s | 38,562 | 02:10:04 |
  | `-ss 4000 -t 300 -i F` | 6s | 38,562 | 02:10:04 |
  | `-ss 4000 -i F` (no `-t`) | — | 38,562 | 02:10:04 |
  | `-t 600 -i F` (seek 0) | 86s | 83,092 | 02:10:04 |

  `-to` as an **output** option does bound correctly (`-ss 4000 … -to 4600` → 1s, 9,170 bytes, last cue 01:16:32) — but see the client constraint below.
- The 120s ceiling is `WriteTimeout: 120 * time.Second` (`cmd/silo/main.go:2389`). `WriteHeader(200)` has already been called (`internal/api/handlers/stream.go:576`) and the logging middleware records that status (`internal/api/middleware/request_logger.go:29`), hence `status=200` on a truncated body.
- Client constraint, from scans of the sibling repos:
  - **silo-apple** — `SidecarSubtitleFetcher.swift:48-95` issues a single GET; `SubtitleSession.swift:244-287` caches the whole response and never re-requests. No `position=`/`duration=`.
  - **silo-android** — `SubtitleTrackMerge.kt:56` builds a bare URL handed to Media3 as a `MediaItem.SubtitleConfiguration` (`SiloPlayerFactory.kt:413`); `ProgressiveMediaSource` fetches once and parses every cue. No re-fetch.
  - **web** windows correctly — `web/src/player/hooks/useSubtitleTracks.ts:79` sends `position=`, advancing by 595 (`WINDOW_DURATION 600 − WINDOW_OVERLAP 5`), the exact stride seen in production.
- Distribution is by codec, not by what it first appears: 135 of 143 embedded requests are subrip. ASS is incidentally safe only because `IsASS` blocks `-ss`.
- Ground truth for windowing is `seek_seconds` in the handler log (118 of 143 non-zero), **not** the API request log, which carries no query field.

## Possible directions

Not a decision, just the option space as it currently looks:

- **Cache the extract, change nothing a client sees.** Keep whole-track delivery and make it cheap. `SubtitleCache` (`internal/playback/subtitle_cache.go`) already implements this shape for PGS; text was excluded on the assumption that "VTT is already windowed and fast", which the inert `-t` makes false. Windowing a cached 83 KB VTT costs ~52ms versus ~16s against the original 27 GB remux. Does not help the first viewer of a file+track.
- **Make the window real** (`-t` → `-to`) — blocked until `silo-android` and `silo-apple` send `position=`/`duration=` and re-fetch on coverage exhaustion. Requires coordinated client work.
- **Extract embedded tracks to sidecars at scan time**, putting every subtitle on the already-instant external path and retiring on-demand extraction entirely. Likely the real endgame.

Whatever the direction, the inert `-t` must not be "cleaned up" in isolation — its comment describes a bound that does not exist and needs correcting in place, or the next reader will fix the flag and break both native clients.

## Related

- Non-finite `?position=` (`+Inf` parses and passes the `>= 0` check, reaches ffmpeg and fails after the 200) is a separate genuine bug found during this investigation.
- The `status=200`-on-truncation masking is largely inherent once headers are committed (`LogSubtitleStreamError`, `internal/playback/subtitle_stream.go:273-277`, documents it as deliberate). A rolling deadline like the one full downloads use (`internal/httpstream/rolling_deadline.go`) may be the right follow-up.
- The standalone proxy has separate uncached ASS/text extraction paths (`internal/proxy/server.go:273`, `:286`).

## AI-use disclosure

Investigation and this write-up were prepared with AI (Claude Code) assistance; client impact was assessed by subagent scans of `silo-apple` and `silo-android`.

CoffeeKnyte:
Closes #423. Part of #296.

## Problem

**Turning on a subtitle mid-film could hang for two minutes, and the server never noticed.**
On the web player, picking an embedded subtitle track sometimes did nothing for a very long
time before cues appeared — or before subtitles simply never showed up at all. Over a 41.6
hour window the subtitle endpoint measured p95 **120,021 ms** and max **121,470 ms**, with 45
of 220 fetches taking over ten seconds.

The 120-second number is not a coincidence: it is the server's own hard write deadline. The
response was being cut off mid-flight. Because the server had already told the client
"200 OK" before the slow part began, every one of those cut-off responses was **recorded in
the logs as a success**. Operators looking at error rates saw nothing wrong, while viewers
were watching a film with no subtitles.

The underlying cost is that a subtitle track is not stored in one piece. Its text is
scattered through the whole file, interleaved between video and audio. Pulling out a few
kilobytes of dialogue means reading across that entire stretch of a multi-gigabyte file on
network storage.

**The safety limit that was supposed to bound this does nothing.** The server passes a
600-second window to ffmpeg, and has since the endpoint was written. It is silently ignored
for these extracts: every request reads from its start point all the way to the end of the
film, no matter what the window says. So the cost of a request tracks how much film is left,
and nobody had noticed because the code reads as though it were bounded.

### perf(playback): cache embedded text-subtitle extracts

Measured against the production binary, `-ss 4000 -t 30` and `-ss 4000 -t 300` produce
byte-identical output to passing no `-t` at all (last cue 02:10:04, end of film). `-t` is
passed as an *input* option and ffmpeg discards it for subtitle extraction; only `-ss`
survives. Cost is therefore `(duration - seek) × bitrate`.

**The window cannot simply be repaired.** silo-apple and silo-android both fetch a track
once and depend on receiving the whole thing. Actually bounding the output would silently
kill subtitles roughly ten minutes into every film on both platforms. The accidental
whole-track behaviour is the de-facto client contract, so this keeps whole-track delivery
and makes it cheap instead.

`SubtitleCache` already had the right shape from the PGS work; text was excluded only by the
assumption that "VTT is already windowed and fast", which the inert `-t` makes false.
Windowing a cached 83 KB VTT costs 52 ms against 16 s for the same window over the original
27 GB remux.

Canonicality is derived from the **effective ffmpeg argv**, not from a flag.
`streamExtractPlanFor` (`internal/playback/subtitle_stream.go`) is the single source of
truth for both the argv and the `partial()` predicate, so only a `seek=0`/`duration=0`
extract can fill the cache. Routing on `AllowWindow` instead would have poisoned it:
`AllowWindow` is set only in the PGS branch, so it is always false for text, while
`streamExtractArgs` applies `-ss` to any non-ASS/non-PGS source regardless. A seeked subrip
request would have taken the full-track path, emitted seek→EOF, exited cleanly, and
published that partial as canonical — and every later viewer starting from 0 would lose all
cues before that point. 118 of 143 production requests carry a non-zero seek.

Supporting changes in `internal/playback/subtitle_cache.go`:

- Cache key gains a schema version and the resolved output profile — an ASS source is
  reachable as both `.ass` and `.vtt`, so format has to be part of the key.
- `removeStaleSiblings` groups by profile, so committing a `.vtt` no longer deletes a valid
  `.ass` sibling. Cleanup and eviction no longer hardcode `.sup`, so text entries are
  actually reclaimed and counted against the budget.
- `InputIsExtractedTrack` carries the cached input's format.
- Text hits use a plain copy with `no-store`, matching cold-path HTTP semantics.
  `http.ServeContent` stays on the PGS path only: Media3 uses range-capable data sources,
  and responses must not vary with cache warmth.
- SUP-specific identifiers renamed now that the cache carries text as well.

The inert `-t` is **deliberately retained** and its comment corrected in place
(`internal/playback/subtitle_stream.go:152`) — removing it, or moving it after `-i`, would
bound the output and break the native clients.

This is latency-only: for every (source codec, requested format, seek) combination, the
bytes a client receives are unchanged.

## Risk / follow-ups

- The 600s window stays inert by design. Anyone "fixing" that line in future breaks
  silo-apple and silo-android; the comment now says so, but it is still a trap.
- Cache budget is the existing 2 GiB shared with PGS entries, still a hardcoded
  `defaultSubtitleCacheMaxBytes`. Text entries are small (tens of KB) so they will not
  meaningfully displace bitmap entries, but exposing this as a config knob following the
  `download.artifact_max_bytes` pattern is still outstanding.
- Concurrent requesters for the same track while a fill is in flight run their own un-teed
  extract rather than waiting. That is no worse than current behaviour and avoids making one
  viewer's first-byte latency depend on another client's connection, but it does mean a
  cold-start burst can duplicate work.
- The status=200-on-truncation observability gap is **not** fixed here; this removes the
  latency that was triggering it. Logging truncated bodies distinctly is separate work.
- Does not touch the jellycompat subtitle endpoint's whole-file buffering, which is the
  other half of #296.

## Verification

- `go build ./...` — clean.
- `go vet ./internal/playback/... ./internal/api/handlers/... ./internal/proxy/...` — clean.
- `go test ./internal/playback/...` — pass. `subtitle_cache_test.go` gains coverage for
  profile-keyed entries, sibling retention across formats, partial requests never filling,
  and text eviction accounting.
- `go test ./internal/api/handlers/... ./internal/proxy/...` — pass.
- `golangci-lint run --new-from-merge-base=origin/main` over the touched packages — 0 issues.
- `gofmt -l internal/` — clean. `make verify-local-paths` — clean.
- ffmpeg argv behaviour confirmed empirically against the production binary: `-t` variants
  byte-identical, documented in the plan doc.
- Rebased onto `origin/main` (8bde6f11); the three touched Go files have had no commits on
  `main` since 2026-07-15, so the rebase carried no conflicts.

Rationale, measurements, four superseded revisions and the dead ends are recorded in
`docs/superpowers/plans/2026-07-17-subtitle-extract-cache.md`.

## AI-use disclosure

Investigated and implemented with AI assistance (Claude Code + Codex gpt-5.6-sol). Codex's
review caught the `AllowWindow` cache-poisoning bug described above. All measurements were
taken against the real production binary and production request logs.

## Summary by CodeRabbit

- **New Features**
  - Added subtitle extraction caching for WebVTT, ASS, and SUP formats.
  - Partial subtitle requests can reuse cached full-track results, improving response times.
  - Subtitle cache entries now preserve distinct output formats and serve the appropriate content type.

- **Bug Fixes**
  - Prevented seeked or failed extractions from creating invalid canonical cache entries.
  - Preserved existing subtitle streaming and HTTP behavior, including compatibility with full-track extraction.

### DIFF
--- a/docs/superpowers/plans/2026-07-17-subtitle-extract-cache.md
+++ b/docs/superpowers/plans/2026-07-17-subtitle-extract-cache.md
@@ -1,0 +1,290 @@
+# 2026-07-17 — Embedded subtitle extraction: investigation and fix
+
+Commands assume the repository root is the cwd.
+
+This document records both the **final design** and the **investigation that
+produced it**, including four superseded revisions. The dead ends are kept
+deliberately: every one of them looked correct, three were disproved only by
+measurement, and the last was a release-blocking bug caught in review. Anyone
+tempted to "clean up" the subtitle windowing code should read Evolution first.
+
+## Summary
+
+Extracting an embedded subtitle track walks the interleaved container: subtitle
+packets sit between video and audio across clusters, so harvesting a few KB of
+text means demuxing that stretch of a multi-GB file off network storage (CephFS).
+
+The server believes it bounds this with a 600s window. **It does not** — the flag
+is inert, so every request extracts from the seek point to end-of-file. We cannot
+turn the window on, because two of three clients depend on receiving whole tracks.
+
+**Fix: cache the extract. Change nothing a client can observe.**
+
+## Evidence
+
+41.6h of production logs (2026-07-15 09:19 → 2026-07-17 02:56, 1.29M requests),
+`/api/v1/stream/{session_id}/subtitles/{track}`:
+
+| client | requests | p50 | p95 | max | ≥10s |
+|---|---|---|---|---|---|
+| browser (web player) | 220 | 72ms | **120,021ms** | 121,470ms | **45** |
+| native (Android/TV) | 42 | 145ms | 9,566ms | 41,607ms | 2 |
+
+Sidecar/external subtitles: **118 requests, all <1s** — standalone files, read
+directly, no demux. That fast path already exists and is why the distribution
+first looked bimodal.
+
+Both clients are slow; they differ in *how often they pay*. The web player
+re-fetches a window every ~10 minutes of playback, so each slow fetch **stalls
+subtitles mid-film** — that is the 120s cluster. Native fetches once per track and
+holds all cues, so it rolls the dice once: up to ~41s before subtitles appear, then
+fine.
+
+The 120s ceiling is `WriteTimeout: 120 * time.Second` (`cmd/silo/main.go:2389`)
+cutting the connection mid-body. `WriteHeader(200)` was already called
+(`stream.go:576`) and the logging middleware records that status
+(`internal/api/middleware/request_logger.go:29`) — so **truncated subtitles are
+logged `status=200`** and are invisible in error metrics.
+
+## Root cause: the window flag is inert
+
+`streamExtractArgs` (`subtitle_stream.go:161-168`) passes `-t` as an **input**
+option, reasoning it "caps how much of the file we read". Measured against the
+production binary (`ffmpeg 7.1.4-Jellyfin` at `/usr/lib/jellyfin-ffmpeg/ffmpeg`
+inside the `silo` container — not on `$PATH`; there is no host ffmpeg):
+
+| command | wall | bytes | last cue |
+|---|---|---|---|
+| `-ss 4000 -t 30 -i F` | 5s | 38,562 | **02:10:04** |
+| `-ss 4000 -t 300 -i F` | 6s | 38,562 | **02:10:04** |
+| `-ss 4000 -i F` (no `-t`) | — | 38,562 | 02:10:04 |
+| `-t 600 -i F` (seek 0) | **86s** | 83,092 | **02:10:04** |
+
+`-t 30` and `-t 300` are byte-identical to each other **and to passing no `-t` at
+all**. Only `-ss` works. Cost is therefore *(duration − seek) × bitrate*, which the
+logs confirm exactly: `seek=8s/130s/519s` → ≥60s; `seek=7797s/8392s/8987s` → <1s.
+Bitrate correlates monotonically (39.8 Mbps → 120.6s; 5.0 Mbps → 8.0s; every 120s
+outlier is a ≥28 Mbps remux).
+
+`-to` as an **output** option does bound it correctly — `-ss 4000 … -to 4600` →
+1s, 9,170 bytes, last cue 01:16:32. **We are deliberately not using it.**
+
+## Why the window stays off
+
+Subagent scans of the sibling repos found the accidental whole-track behaviour has
+become the de-facto contract:
+
+- **silo-apple — BREAKS.** `SidecarSubtitleFetcher.swift:48-95` issues a single
+  GET; `SubtitleSession.swift:244-287` caches the whole response in
+  `sidecarCache[urlIndex]` and never re-requests. No `position=`/`duration=`, no
+  window constant, no sliding window. Its comment states the "buffered path" is
+  deliberate.
+- **silo-android — BREAKS.** `SubtitleTrackMerge.kt:56` builds the URL bare and
+  hands it to Media3 as a `MediaItem.SubtitleConfiguration`
+  (`SiloPlayerFactory.kt:413`) → `ProgressiveMediaSource` fetches once, parses every
+  cue. No timer, no position listener, no re-fetch.
+- **web** windows correctly: sends `position=`
+  (`web/src/player/hooks/useSubtitleTracks.ts:79`), advances by 595
+  (`WINDOW_DURATION 600 − WINDOW_OVERLAP 5`, `:9`/`:16`/`:409-410`) — the exact
+  stride observed in production.
+
+Enabling the window would silently kill subtitles ~10 minutes into every film on
+iOS, tvOS, macOS and Android. Caching removes the need: keep whole-track delivery,
+make it cheap.
+
+## Design
+
+`SubtitleCache` (`internal/playback/subtitle_cache.go`) already implements the
+needed shape for PGS: a full-track miss streams to the client while teeing into a
+temp file, atomically renamed on clean exit (`ServeSUPExtract:113`); a hit serves
+from disk; a windowed request re-extracts from the small cached artifact
+(`serveWindowedSUP:161`, via `InputIsExtractedSup`) or starts `WarmInBackground:188`.
+Keyed by path + track ordinal + source mtime+size. Text was excluded only by the
+assumption at `stream.go:557-562` that "VTT is already windowed and fast" — which
+the inert `-t` makes false.
+
+Windowing a **cached** 83 KB VTT costs **52ms** vs 16,000ms against the original
+27 GB remux (`-ss 4000 -i cached.vtt … -to 4600` → 9,216 bytes, cues 01:04:40 →
+01:16:32). Note `-ss` on a webvtt input lands ~2 min *before* the target — a
+coverage superset, not byte-equivalence.
+
+### The load-bearing rule: canonical artifact + effective-partial predicate
+
+A cache entry may only ever hold a **canonical extract**: source track → requested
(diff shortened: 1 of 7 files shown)


Quick104:
Closing as superseded by #938. That PR landed broader subtitle reliability and caching, and current main already has SubtitleCache.ServeExtract. I found no separate requirement in this PR that remains uncovered.

PR #552, opened 2026-08-06
fix(playback): degrade an unavailable track instead of refusing to play

RXWatcher:
## Problem

Starting the next episode could fail outright:

```
The selected subtitle track is unavailable in the effective file version.
selected audio track is unavailable
```

Both come from carrying a selection across files. A viewer picks a subtitle or audio track on one episode; the next episode — or an alternate version the planner falls back to — has a different track layout; the rebase finds no equivalent and the start is refused. On Android TV that surfaces as an error screen the viewer has to replay out of.

Refusing is disproportionate to what actually went wrong. Nothing is broken: the file is playable, only one preference cannot be honoured.

## Approach

- **Subtitles** are an enhancement, so an absent track clears the selection and lets the planner apply the profile default.
- **Audio** cannot simply be dropped, so an out-of-range selection falls back to the file's own default through the existing `normalizeAudioTrackIndex`.
- Both report a degradation warning on the plan — `subtitle_track_unavailable`, `audio_track_unavailable` — so clients can tell the viewer what happened rather than silently swapping tracks under them.

Malformed requests still error. An unparseable track identity, or a negative index, is a client bug rather than a content mismatch, and swallowing it would make that bug harder to find.

## Tests

- `TestRemapSubtitleSelectionV3DropsSelectionWhenAbsentFromTarget` — missing track drops the selection, reports degraded, does not error
- `TestRemapSubtitleSelectionV3RebasesWhenTargetHasTheTrack` — an equivalent track is still preferred over dropping
- `TestResolveV3AudioIndexFallsBackWhenSelectionIsOutOfRange` — falls back to a valid index, reports degraded
- existing `TestRemapSubtitleSelectionV3RejectsNegativeIndex` kept — malformed input must still fail

`go test ./internal/api/handlers/` passes. `./internal/playback/` has 4 pre-existing failures on this machine (`TestResolveHWAccel*`, `TestFFmpegSupportsNVENC*`) which reproduce identically on a pristine `main` — they probe for NVIDIA hardware.

## Note for client authors

Clients that treat a start response as all-or-nothing will now start playing where they previously showed an error. The degradation warnings are the signal to surface "the track you picked isn't on this episode" without blocking playback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Playback now continues when selected audio is unavailable by falling back to the file’s default audio track.
  * Unavailable subtitle selections are safely removed instead of preventing playback.
  * Playback reports when audio or subtitle selections were degraded.
  * Valid subtitle selections are preserved and correctly remapped.
  * Invalid subtitle indices and malformed track selections are rejected.

### DIFF
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -460,7 +460,10 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
 			audioIndex = remapAudioIndexV3(requestedFile, effectiveFile, audioIndex)
 			subtitleDropped, subtitleErr := h.remapSubtitleSelectionV3(r.Context(), requestedFile, effectiveFile, &req)
 			if subtitleErr != nil {
-				writeJSON(w, http.StatusCreated, playback.NewTerminalResponseV3("subtitle_unavailable_in_version", subtitleErr.Error(), false))
+				// Post-degrade, remap errors are malformed-request-only
+				// (bad identity, negative index) — a client bug, not a
+				// content mismatch, matching the audio path above.
+				writeError(w, http.StatusBadRequest, "bad_request", subtitleErr.Error())
 				return
 			}
 			if subtitleDropped {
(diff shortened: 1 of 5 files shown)


RXWatcher:
Both inline comments were raised against the first commit:

- **Replan degradation warnings** — already addressed in `ea64fd0b`, which landed between the review run and now: `executeReplanV3` collects `audioDegraded`/`subtitleDegraded` across all three sites and appends both warning codes to the final plan, deduplicated against the planner's own subtitle warning. That commit also moved the subtitle degrade into `ResolveSubtitlePolicyV3` itself, so the direct-start path (the commonest shape of the reported bug) degrades too.
- **Test assertion** — fixed in `eac54652`: the fallback test now pins the exact index `normalizeAudioTrackIndex` chooses rather than accepting anything in range.

PR #561, opened 2026-08-07
feat(catalog): rework TV recents and poster playback

Rhainland:
## What changed

- Group TV Recently Added entries by exact scan run: a single newly added episode renders as an episode, while multi-episode and legacy additions render as the series.
- Preserve separate events for repeated scan runs and provide event-specific direct-play targets.
- Add immutable scan-run provenance to first-seen files and episode-library memberships, including synchronous admin refresh ingestion.
- Add profile-aware play targets for playable movie, series, season, and episode cards across Home, catalog, search, recommendations, Continue Watching, and season views.
- Add independent parent-series title navigation for episode cards and a centered accessible play overlay.
- Preserve Jellyfin `/Items/Latest` compatibility through an explicit TV-grouping opt-out.
- Invalidate Recently Added caches across catalog scan-complete events with leading/trailing burst coalescing.

## Why

Recently Added previously linked every TV update to the series or season, making a newly arrived episode slower to reach. It also lacked a consistent direct-play interaction across media-card surfaces. This aligns TV additions with Plex-style behavior while preserving existing detail navigation and compatibility clients.

## Reliability and architecture

- Scan provenance is recorded only on first discovery and is never overwritten by rescans or delayed matching.
- Queue workers cannot claim synchronous admin-refresh scan runs.
- Direct admin runs heartbeat while active; abandoned direct runs fail instead of being requeued and executed twice.
- Personalized playback targets are resolved after shared-cache lookup and are never stored in the process-global list cache.
- Play-target resolution is batched and applies library/profile availability before selecting an episode.
- Cache generations apply only to Recently Added, protect against stale in-flight refreshes, and retain a trailing invalidation when scan events are debounced.

## Validation

Passed:

- `make build`
- database-backed Recently Added grouping, provenance, playable-target, scanqueue, and admin-refresh tests
- affected Go package suites: `internal/scanqueue`, `internal/adminjob`, `internal/sections`, `internal/catalog`, `internal/scanner`, and API handlers
- focused Go race-detector tests for cache invalidation and direct-scan heartbeats
- 45 focused frontend card/play-overlay tests
- `make migrate-validate`
- `make verify-local-paths`
- `golangci-lint run --new-from-rev=main` (0 issues)
- `pnpm run lint` (0 errors; existing warnings only)
- `pnpm run format:check`
- `git diff --check`

Repository-wide baseline notes:

- `make test-web`: 1,784 tests passed; 36 existing failures remain in appearance/localStorage and settings-realtime tests.
- `make test-go`: affected packages pass; existing Jellyfin compatibility process-lock/shared-database and playback-environment failures remain.
- `make lint`: the repository-wide baseline currently reports 303 pre-existing findings; changed-line lint reports none.

## Manual QA

Verified against an isolated PostgreSQL/Redis stack using real Home and catalog pages:

- single-episode, multi-episode, repeated-run, and legacy TV events
- independent poster, parent-series title, and direct-play navigation
- movie, episode, series, and season play targets
- unavailable media omitting the play control
- clean browser console and expected API payloads

## Risks

- The migration adds nullable foreign keys and partial FK indexes without backfilling historical rows.
- Historical rows intentionally remain series cards because trustworthy scan provenance cannot be reconstructed.
- Mobile clients are unchanged; all API additions are optional `/api/v1` fields.

## Review follow-up (Quick104)

Addressed all six functional findings from [Quick104's review](https://github.com/Silo-Server/silo-server/pull/561#pullrequestreview-4908762620) in `7c26f9ee`:

- enforce effective playback-quality and source-library constraints during play-target resolution;
- rank series/season targets with one batched read through the configured PostgreSQL or SQLite user store;
- expose keyboard search selection through real focus in a semantic result list while keeping Play as an independent control;
- recover stale direct scans from both `accepted` and `running`;
- use top-level concurrent index cleanup on migration retries.

Follow-up validation: `make build`; affected Go package suites; 13 focused GlobalSearch tests; `make migrate-validate`; `make verify-local-paths`; frontend lint/format checks; and changed-line `golangci-lint` (0 issues).

### CodeRabbit follow-up

Addressed CodeRabbit's follow-up review in `4a5fc686`:

- normalize result focus indices and explicitly wrap ArrowUp from an unselected search box to the final result;
- make the concurrent-index regression assertions whitespace- and statement-boundary-aware, rejecting both blocking ordinary-drop forms;
- exercise stale accepted direct-run recovery through `requested_at` while leaving `started_at` and `heartbeat_at` unset.

Follow-up review found no material issues. Validation passed: affected Go suites, 13 focused GlobalSearch tests, TypeScript build, migration validation, frontend format checks, local-path checks, and changed-line Go lint (0 issues). ESLint reports 0 errors and one pre-existing warning.

### AI Disclosure
- Tool(s): Codex desktop
- Model(s): gpt-5.6-sol
- Involvement: fully AI-generated under human direction and review
- Adversarial review: Reviewed the full branch for concurrency, cache consistency, access boundaries, scan lifecycle, API contracts, and frontend navigation. The review found dropped trailing cache invalidations, missing library-refresh scan-complete publication, and live direct scans being treated as abandoned; all three were corrected and covered by focused and race/database-backed tests.

## Summary by CodeRabbit

* **New Features**
  * Added one-click playback overlays to movie, episode, season, recommendation, search, and continue-watching cards.
  * Playback selects accessible content, supports resume behavior, and preserves return navigation.
  * Episode cards link to their series pages while retaining episode details.
  * Added playback identifiers to item, season, section, and browse responses.
  * Improved Recently Added TV sections with episode/series grouping and pagination.
  * Added scan provenance tracking for newly discovered media.

* **Bug Fixes**
  * Refreshed catalog results now update more reliably after scans.
  * Improved scan tracking, heartbeat monitoring, and failure handling for direct library and item refreshes.


internal/adminjob/direct_scan_run.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("create scan run: %w", err)
+	}
+	if !created {
+		return nil, nil, fmt.Errorf("scan scope already has active run %s", run.ID)
+	}
+	run, err = repo.Start(ctx, run.ID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("start scan run: %w", err)
```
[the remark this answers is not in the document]
Rhainland: Fixed in 0b330d7d. The direct-scan path now retains the ID returned by Create before calling Start. If Start fails, it calls MarkCancelled with a 10-second context derived from context.WithoutCancel(ctx), so an expired request context cannot strand the accepted run. Cleanup failures are joined with the original startup error. TestBeginDirectSubtreeScanCancelsAcceptedRunWhenStartFails verifies both the transition and the detached cleanup context.

internal/adminjob/item_refresh.go
```diff
@@ -494,10 +497,19 @@ func (e *ItemRefreshExecutor) Execute(ctx context.Context, req ItemRefreshReques
...
 	}
+	stopScanHeartbeat := startDirectScanHeartbeat(e.scanRuns, scanRun.ID, directScanHeartbeatEvery)
+	ingestResult, err := e.ingester.IngestSubtree(scanCtx, folder, req.ScanPath)
+	stopScanHeartbeat()
+	if err != nil {
+		return nil, fmt.Errorf("scan scope: %w", failDirectScan(ctx, e.scanRuns, scanRun, err))
+	}
+	if err := completeDirectScan(ctx, e.scanRuns, scanRun, ingestResult); err != nil {
+		return nil, fmt.Errorf("scan scope: %w", err)
+	}
```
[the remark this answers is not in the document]
Rhainland: Fixed in 0b330d7d. EventScanComplete is now published immediately after completeDirectScan succeeds and the later publication was removed, so the event is emitted once even if folder revalidation, target resolution, or metadata refresh subsequently fails. TestItemRefreshPublishesScanCompleteBeforeMetadataFailure covers the post-commit metadata-failure case.

internal/catalog/catalog_resolver.go
```diff
@@ -430,6 +433,90 @@ func (r *CatalogResolver) resolveSectionSource(ctx context.Context, req CatalogR
...
+	effectiveLibraryIDs, tvScoped, err := ResolveRecentTVLibraryIDs(
+		ctx,
+		r.itemRepo.pool,
+		requested,
+		filters.FilterType == recentTVTypeSeries,
+		access,
+	)
+	if err != nil || !tvScoped {
+		return nil, tvScoped, err
+	}
```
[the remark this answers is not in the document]
Rhainland: No code change here because media_scope "video" is intentionally not an explicit TV scope. MediaScopeItemTypes("video") includes both movies and series, while this feature must leave mixed-media Recently Added sections unchanged. Treating "video" like "series" would incorrectly route a mixed movie/series row through the TV-only resolver and omit its movies. A video-scoped row whose requested libraries are all dedicated TV libraries is already recognized by ResolveRecentTVLibraryIDs through its existing library-type validation, so the exclusively-TV case remains covered.

internal/catalog/recent_tv_db_test.go
```diff
@@ -0,0 +1,313 @@
...
+	if err := pool.QueryRow(ctx, `INSERT INTO media_folders (type, name, enabled) VALUES ('series', $1, true) RETURNING id`, series("tv-folder")).Scan(&tvFolderID); err != nil {
+		t.Fatalf("seed TV folder: %v", err)
+	}
+	if err := pool.QueryRow(ctx, `INSERT INTO media_folders (type, name, enabled) VALUES ('movies', $1, true) RETURNING id`, series("movie-folder")).Scan(&movieFolderID); err != nil {
+		t.Fatalf("seed movie folder: %v", err)
+	}
+	t.Cleanup(func() {
+		_, _ = pool.Exec(ctx, `DELETE FROM media_items WHERE content_id = ANY($1)`, []string{seriesA, seriesB, seriesC, seriesD})
+		_, _ = pool.Exec(ctx, `DELETE FROM media_folders WHERE id = ANY($1)`, []int{tvFolderID, movieFolderID})
+	})
```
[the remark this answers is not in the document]
Rhainland: No code change is needed. scan_runs.media_folder_id is defined with REFERENCES media_folders(id) ON DELETE CASCADE in migration 085_scan_runs.sql. These cleanup callbacks delete the fixture media items and then the media folders; deleting each folder therefore cascades to its seeded scan_runs rows instead of being blocked by them. Explicitly deleting the same runs first would be redundant and there is no accumulating scan-run leak under the current schema.

migrations/sql/20260807094340_add_scan_run_provenance_to_media_availability.sql
```diff
@@ -0,0 +1,30 @@
...
+    ADD COLUMN first_seen_scan_run_id text NULL
+        REFERENCES public.scan_runs(id) ON DELETE SET NULL;
+
+CREATE INDEX idx_media_files_first_seen_scan_run_id
+    ON public.media_files (first_seen_scan_run_id)
+    WHERE first_seen_scan_run_id IS NOT NULL;
+
+CREATE INDEX idx_episode_libraries_first_seen_scan_run_id
+    ON public.episode_libraries (first_seen_scan_run_id)
+    WHERE first_seen_scan_run_id IS NOT NULL;
```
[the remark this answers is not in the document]
Rhainland: Fixed in 0b330d7d. The migration is now non-transactional, adds nullable columns first, installs explicitly named foreign keys as NOT VALID, builds both partial indexes concurrently, and validates the constraints afterward. The down path drops indexes concurrently. It also handles retry after a partial non-transactional failure by using IF NOT EXISTS and removing invalid concurrent-index artifacts before rebuilding. make migrate-validate passes.

web/src/components/GlobalSearch.tsx
```diff
@@ -48,59 +49,73 @@ function GlobalSearchResultRow({
...
+      {item.play_content_id ? (
+        <div className="pointer-events-none absolute top-2 left-3 h-14 w-10">
+          <CardPlayOverlay
+            contentId={item.play_content_id}
+            title={item.title}
+            type={item.type === "movie" ? "movie" : "episode"}
+            size="compact"
+            onPlaybackStart={onPlay}
+          />
+        </div>
```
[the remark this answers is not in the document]
Rhainland: Fixed in 0b330d7d. Global Search now uses a labeled non-listbox result group. Each result's detail button and CardPlayOverlay link are sibling controls rather than interactive descendants of role=option, and the now-invalid aria-activedescendant relationship was removed. Keyboard arrow selection and Enter-to-open behavior are retained. The focused GlobalSearch suite passes 13/13 and includes an assertion that no listbox/option roles wrap the independent button and play link.

Quick104 (commented):
I reviewed the current head (`0b330d7`). Setting aside process and standards concerns, these functional issues should be addressed:

### P1

1. **Playable-target resolution ignores the profile's maximum playback quality** (`internal/catalog/playable_targets.go:105-128`). The resolver filters missing files, disabled folders, and library access, but never applies `Access.MaxPlaybackQuality`. Detail and playback enforce that limit later, so a profile capped at 1080p can receive a Play control and `play_content_id` for media whose only file is 4K, then reach a watch route with no allowed version. Please apply the same quality predicate used by `catalog.FileAllowedByAccess` and add a resolver test for a quality-capped profile.

2. **Library-scoped section catalogs lose their library constraint during play-target enrichment** (`internal/api/handlers/catalog.go:147`). For `source=section&scope=library`, the parser stores the scope in `CatalogRequest.LibraryID`, while this call passes only `req.Query.LibraryIDs`, which is empty for section sources. A series card can therefore select an episode from another accessible library. Please pass the effective source-library scope (including `req.LibraryID`) and cover a series present in multiple libraries.

3. **Keyboard search selection is no longer exposed to assistive technology** (`web/src/components/GlobalSearch.tsx:273-301`). The input's `aria-activedescendant` and the results' listbox/option semantics were removed, but ArrowUp/ArrowDown still update a visual `selectedIndex` on generic elements. Screen-reader users consequently receive no announcement of the selected result. Please restore a valid accessible composite/focus pattern while keeping the nested Play action valid, and add coverage for the resulting ARIA/focus behavior.

### P2

4. **Series/season resume selection bypasses SQLite-backed user progress** (`internal/catalog/playable_targets.go:201-204`). The resolver joins the PostgreSQL `user_watch_progress` table directly, so a supported `userdb.backend=sqlite` deployment cannot use its progress when selecting an episode; an in-progress episode may be ignored. Please resolve progress through the configured user-store abstraction (ideally as a batched overlay) and add backend-conformance coverage.

5. **A crash before a direct scan starts can permanently block that scope** (`internal/scanqueue/repository.go:315-334,517-533`). Direct triggers are excluded from queue workers, and stale cleanup only fails direct runs in `running`. If the process dies after `Create` but before `Start`, the run remains `accepted` indefinitely and the unique active-scope constraint rejects future refreshes. Please include stale accepted direct runs in recovery or make creation/start atomic, with a crash-window test.

6. **The migration's failed-build recovery can block catalog traffic** (`migrations/sql/20260807094340_add_scan_run_provenance_to_media_availability.sql:48-88`). Although the indexes are built concurrently, the recovery blocks use ordinary `DROP INDEX`; retrying after an interrupted concurrent build can therefore wait for an `ACCESS EXCLUSIVE` lock on actively queried/ingested tables. Please move recovery to top-level `DROP INDEX CONCURRENTLY IF EXISTS` statements before the concurrent creates and validate the retry path.

Rhainland:
@Quick104 Thanks for the careful review — all six findings are addressed in [`7c26f9ee`](https://github.com/Silo-Server/silo-server/commit/7c26f9eeb690308157f61f3c40f20cc2e4685772).

Highlights:
- quality ceilings and library-section scope now constrain play-target selection;
- resume ranking uses a single batched read through the configured user store, including SQLite coverage;
- search ArrowUp/ArrowDown now moves real focus through a semantic result list while Play remains an independent control;
- stale direct runs are recovered from both `accepted` and `running`;
- migration retry cleanup uses top-level `DROP INDEX CONCURRENTLY`.

Validation passed: build, affected Go suites, 13 focused search tests, migration validation, frontend lint/format, local-path checks, and changed-line Go lint (0 issues). The new PostgreSQL-backed cases are committed for CI; this checkout did not have `SILO_TEST_DATABASE_URL` configured.

migrations/scan_run_provenance_test.go
```diff
@@ -0,0 +1,35 @@
...
+			t.Fatalf("migration missing concurrent retry cleanup %q", drop)
+		}
+		if createAt < 0 {
+			t.Fatalf("migration missing concurrent create %q", create)
+		}
+		if dropAt > createAt {
+			t.Fatalf("retry cleanup for %s must precede its create", indexName)
+		}
+		if strings.Contains(up, "DROP INDEX public."+indexName) {
+			t.Fatalf("migration contains blocking ordinary drop for %s", indexName)
```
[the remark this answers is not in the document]
Rhainland: Fixed in `4a5fc686`. The test now normalizes SQL whitespace, matches each complete `CREATE INDEX CONCURRENTLY ...;` statement for the expected table and predicate, and rejects both blocking ordinary-drop forms while preserving the drop-before-create assertion. `go test ./migrations` and `make migrate-validate` pass.

Rhainland:
@coderabbitai Addressed the outside-diff stale accepted-run test feedback in [`4a5fc686`](https://github.com/Silo-Server/silo-server/commit/4a5fc686a9573e44da7df75d1f27f40db4b15f61). The setup now backdates `started_at`/`heartbeat_at` only for started runs; the accepted direct run keeps both null and is made stale through `requested_at`, so the test exercises the production `COALESCE(heartbeat_at, started_at, requested_at)` branch. The scanqueue suite passes; the PostgreSQL-backed case remains committed for CI because this checkout has no `SILO_TEST_DATABASE_URL` configured.

web/src/components/GlobalSearch.tsx
```diff
@@ -229,6 +260,17 @@ export function GlobalSearch({
...
     !tmdbDebounceCatchingUp;
   const showError = previewQuery.isError;
+  const moveResultFocus = useCallback(
+    (nextIndex: number) => {
+      if (items.length === 0) {
+        setSelectedIndex(-1);
+        searchInputRef.current?.focus();
+        return;
+      }
+      setSelectedIndex((nextIndex + items.length) % items.length);
```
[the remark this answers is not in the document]
Rhainland: Fixed in `4a5fc686`. `moveResultFocus` now normalizes arbitrary indices, and ArrowUp from the unselected `-1` sentinel explicitly targets `items.length - 1`—double modulo alone would still map `-2` to `0` for a two-result list. The focused test now covers ArrowUp from an unselected two-result list and confirms focus/`aria-current` land on the final result. All 13 GlobalSearch tests pass.

internal/catalog/recent_tv.go
```diff
@@ -0,0 +1,412 @@
...
+		)
+		SELECT page.target_id, page.target_type, page.added_at,
+		       COALESCE(page.single_episode_id, play_target.content_id) AS play_content_id%s
+		%s
+		LEFT JOIN LATERAL (
+			SELECT e_play.content_id
+			FROM episodes e_play
+			WHERE page.target_type = 'series'
+			  AND e_play.series_id = page.series_id
+			  AND e_play.season_number = page.anchor_season_number
```
Quick104, line 474: **Recently-added TV `play_content_id` bypasses the profile playback-quality ceiling**

The `play_target` LATERAL only checks `el_play.media_folder_id = ANY($1)` and `mf_play.missing_since IS NULL`. It never applies `MaxPlaybackQuality`, which `PlayableTargetResolver` deliberately does enforce (`playable_targets.go`, the `CASE UPPER(BTRIM(mf.resolution))` rank clause).

Worse, this value then wins: `catalog.go:222`, `items.go:923` and `sections.go:1444` all only fall back to the resolver when `item.PlayContentID` is empty, so the enforced target is never used for TV recents.

Concretely: a profile capped at 1080p sees a series card whose Play button targets an episode that only exists as a 2160p file, and playback hard-fails at stream-token time.

Note the cache angle too — `PlayContentID` set here rides the process-global resolved-list cache, whose key (`resolvedListCacheKey` -> `WriteAccessScopeCacheKey`) excludes `MaxPlaybackQuality`. So making this value quality-dependent *in the cached path* would leak one profile's ceiling to another. The safe shape is to keep this as a profile-independent anchor **hint** and let the post-cache `PlayableTargetResolver` validate/resolve it per profile.

internal/scanqueue/service.go
```diff
@@ -283,7 +284,17 @@ func (s *Service) requeueStale() {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	requeued, err := s.repo.RequeueStaleRunning(ctx, time.Now().UTC().Add(-s.staleAfter))
+	staleBefore := time.Now().UTC().Add(-s.staleAfter)
+	failed, err := s.repo.FailStaleDirect(ctx, staleBefore)
```
Quick104: **`requeueStale()` aborts the pre-existing stale-run requeue when the new `FailStaleDirect` errors**

The new `FailStaleDirect` call `return`s on error, so a transient failure here disables `RequeueStaleRunning` entirely — behavior that worked before this PR.

A statement timeout or lock contention on `scan_runs` during a large scan is enough: `requeueStale` logs a warning and returns, and queue scan runs abandoned by a crashed worker are never requeued for as long as the error persists. The scan queue silently stalls.

Log and continue to `RequeueStaleRunning` rather than returning.

internal/adminjob/direct_scan_run.go
```diff
@@ -0,0 +1,145 @@
...
+		LibraryID: libraryID,
+		Mode:      scanqueue.ModeSubtree,
+		Path:      path,
+		Trigger:   trigger,
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("create scan run: %w", err)
+	}
+	if !created {
+		return nil, nil, fmt.Errorf("scan scope already has active run %s", run.ID)
```
Quick104: **Admin item/library refresh now hard-fails whenever any active scan run covers the same scope**

When `scanqueue.Repository.Create` reports `created == false` (the partial unique index `idx_scan_runs_active_scope` already has an `accepted`/`running` row for the same `(media_folder_id, mode, path)`), this returns a hard error. Before this PR the same call path just ran `IngestSubtree`.

Failure case: autoscan enqueues a subtree scan for `/media/tv/Show` (status `accepted`, not yet claimed). An admin clicks "Refresh item" on that show and gets `scan scope: scan scope already has active run 01J…`. Same for two concurrent refreshes of one item, or a retry while an earlier direct run is still inside the 2-minute stale window.

A reasonable fix keeps the old behavior: when the scope is already claimed, proceed with ingestion without owning a scan run (empty `scanbatch` run ID) and log it. Those files just get NULL provenance, i.e. the documented legacy/series-card path.

internal/catalog/playable_targets.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+		candidates[contentID] = append(candidates[contentID], playContentID)
+		allCandidateIDs = append(allCandidateIDs, playContentID)
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("iterating playable poster targets: %w", err)
+	}
+	progress := map[string]userstore.WatchProgress{}
+	if q.ProgressStore != nil && len(allCandidateIDs) > 0 {
+		progress, err = q.ProgressStore.ListProgressByMediaItems(ctx, q.ProfileID, allCandidateIDs)
```
Quick104: **Unbounded candidate fan-out becomes one bind parameter per episode**

`Resolve` returns one row per (card, candidate episode) with no per-card cap, then hands the entire flattened list to `ListProgressByMediaItems`. The Postgres implementation (`internal/userstore/pgstore/progress.go:650-656`) emits **one placeholder per ID**.

Some callers reach `BrowseFilters.MaxLimit` of 1000. For 100 long-running series averaging 700 episodes, `allCandidateIDs` holds 70,000 IDs and the generated `IN (...)` exceeds Postgres' 65535 extended-protocol parameter limit — the request fails outright. Below that ceiling it still ships tens of thousands of rows and params on every list response.

At minimum: dedupe `allCandidateIDs` and chunk the progress lookup. Better: bound candidates per card in SQL (`ROW_NUMBER() OVER (PARTITION BY ord …)`) or push the ranking into SQL.

web/src/components/GlobalSearch.tsx
```diff
@@ -209,9 +240,9 @@ export function GlobalSearch({
   // Auto-scroll the selected result into view
   useEffect(() => {
     if (selectedIndex >= 0) {
-      document
-        .getElementById(`search-result-${selectedIndex}`)
-        ?.scrollIntoView({ block: "nearest" });
+      const selectedResult = document.getElementById(`search-result-${selectedIndex}`);
+      selectedResult?.focus();
```
Quick104: **Arrow-key selection steals DOM focus, so the user can no longer type to refine the query**

The selection effect now calls `selectedResult?.focus()`, moving DOM focus off the search input onto a result `<button>`. A button accepts no text, and its `onKeyDown` handles only ArrowUp/ArrowDown.

Repro: type `bre`, press ArrowDown once. `selectedIndex` becomes 0, focus moves to `#search-result-0`, the input blurs. Now type `aking` — nothing happens; the keystrokes go to the button and are discarded. The `useEffect` at line 236 that resets `selectedIndex` on query change can never fire, because `query` can no longer change.

The previous `aria-activedescendant` virtual-focus combobox pattern is the standard here and keeps the Play control independently reachable via Tab.

internal/api/handlers/sections.go
```diff
@@ -1237,6 +1238,28 @@ func (h *SectionHandler) buildSectionsResponse(r *http.Request, withItems []sect
...
+	playTargets := map[string]string{}
+	if h.fetcher != nil {
+		inputs := make([]catalog.PlayableTargetInput, 0, len(allItems))
+		for _, item := range allItems {
+			if item == nil || item.ContentID == "" {
+				continue
+			}
+			inputs = append(inputs, catalog.PlayableTargetInput{ContentID: item.ContentID, Type: item.Type})
+		}
+		resolvedTargets, err := h.fetcher.ResolvePlayableTargets(r.Context(), catalog.PlayableTargetQuery{
```
Quick104: **Section play targets ignore watch progress (no `ProgressStore` passed)**

This `PlayableTargetQuery` omits `ProgressStore`, even though `SectionHandler.StoreProvider` is right there. `preferredPlayableTarget` then sees an empty progress map, ranks every candidate 1, and returns the lowest-numbered available episode.

So a profile watched through S03E05 gets "resume S03E05" from the series card on `/api/v1/catalog` (which does pass the store via `resolvePlayableTargetInputs`) and "play S01E01" from the identical card in a Home rail. Two surfaces, two answers.

internal/adminjob/item_refresh.go
```diff
@@ -494,10 +497,20 @@ func (e *ItemRefreshExecutor) Execute(ctx context.Context, req ItemRefreshReques
...
+	e.publish(cache.EventScanComplete, strconv.Itoa(req.ScanFolderID))
```
Quick104: **`EventScanComplete` now published before the metadata refresh**

This publish moved from after `refreshMetadata` to immediately after the scan completes. With the new `sections.InvalidateResolvedListCache` subscriber wired to this event (`cmd/silo/main.go:659`), the generation can advance and a rail can rebuild from **pre-refresh** data.

That freshly-cached entry then holds stale title/poster/type for the full 15-minute `resolvedListTTL`, because `EventMetadataUpdated` (published one line later) is not wired to the invalidator.

Same ordering question applies to the new publish in `library_refresh.go:435`. Either restore the post-refresh ordering, or also invalidate on `EventMetadataUpdated`.

internal/catalog/recent_tv.go
```diff
@@ -0,0 +1,412 @@
...
+		if err := rows.Scan(&id, &libraryType); err != nil {
+			return nil, false, fmt.Errorf("scanning TV recent library: %w", err)
+		}
+		byID[id] = strings.ToLower(strings.TrimSpace(libraryType))
+	}
+	if err := rows.Err(); err != nil {
+		return nil, false, fmt.Errorf("iterating TV recent libraries: %w", err)
+	}
+
+	if !explicitSeries {
```
Quick104, line 128: **A non-series `filter_type` is hijacked by TV event grouping**

When `explicitSeries` is false the decision is made purely from library types, ignoring a configured `filter_type` of `movie`/`episode`/`season`.

A rail configured `{"filter_type":"episode"}` scoped to library 7 (type `series`) satisfies "every requested library is type series", so `tvScoped` is true and `RecentTVRepository.List` takes over — returning a mix of series cards (multi-episode and legacy events) *and* episode cards, ignoring the episode-only request. The old `buildRecentlyAddedQuery` path ran `applyConfigTypeFilter` and honored it.

Pass the actual `FilterType` through and bail out to the generic path for any non-empty, non-`series` value.

internal/api/handlers/items.go
```diff
@@ -911,9 +916,14 @@ func (h *ItemsHandler) writeCatalogBrowseResponse(w http.ResponseWriter, r *http
 
 	overlaySummaries := h.listOverlaySummaries(r.Context(), result.Items, h.accessFilter(r))
 	userStates := h.listItemUserStates(r, result.Items)
+	playTargets := h.listPlayableTargets(r, result.Items, req.Query.LibraryIDs, h.accessFilter(r))
```
Quick104: **`/browse` never populates `series_id`, so the new episode heading link is dead there**

`writeCatalogBrowseResponse` resolves play targets but builds responses through `toItemListResponseWithOverlay` -> `itemListResponseShell`, which never sets `SeriesID`. Only `CatalogHandler.catalogItemResponses` calls `listEpisodeBrowseMetadata`/`applyEpisodeBrowseMetadata`.

So an episode card rendered from a `/browse` response has `item.series_id === undefined`, and `ItemCard.tsx:232` collapses `headingHref` to `itemHref`. Clicking the series title opens the episode's own detail page — silently diverging from the identical card fed by `/api/v1/catalog`.

internal/sections/resolvedlistcache.go
```diff
@@ -255,17 +310,24 @@ func cloneMediaItems(items []*models.MediaItem) []*models.MediaItem {
...
-// jellyfin-compat /Items/Latest for that same library are built once and reused
-// across both surfaces.
+// entry. Compatibility behavior flags that change membership are included.
 //
 // userID/profileID are still excluded so entries are shared across everyone with
 // the same access; the per-user overlay is always recomputed downstream, so a
 // cached entry never leaks one profile's state to another.
 func resolvedListCacheKey(resolved ResolvedSection, libraryID *int, libraryIDs []int, filter catalog.AccessFilter) string {
 	var b strings.Builder
+	if resolved.SectionType == SectionRecentlyAdded {
```
Quick104, line 371: **Superseded cache generations are retained for the full 15-minute TTL**

Advancing `resolvedListGeneration` only changes *future* keys. The entries it obsoletes stay in `resolvedListCache` until `expiresAt` (15 min) and are only swept opportunistically inside `resolvedListSet`, at most once per minute.

During a large multi-library rescan the generation advances every 30 s, so up to `resolvedListTTL / resolvedListInvalidationInterval` = 30 generations of every (library × access-scope) entry are alive at once. With 20 access scopes that is 600 live entries instead of 20 — a 30× memory multiplier that only decays once scanning stops.

`InvalidateResolvedListCache` should drop the keys it supersedes.

internal/catalog/recent_tv.go
```diff
@@ -0,0 +1,412 @@
...
+			       COALESCE(scan_run_id, '') AS event_id,
+			       series_id,
+			       anchor_season_number,
+			       CASE WHEN scan_run_id IS NOT NULL AND episode_count = 1 THEN episode_id END AS single_episode_id
+			FROM episode_events
+			UNION ALL
+			SELECT series_id, 'series'::text, added_at, ''::text, series_id, NULL::integer, NULL::text
+			FROM series_without_episode_events
+		),
+		deduplicated AS (
```
Quick104: **The `deduplicated` CTE can never remove a row**

`episode_events` is already `GROUP BY e.series_id, el.first_seen_scan_run_id`, so `(target_id, target_type, event_id)` is unique within it; and `series_without_episode_events` is disjoint from it by construction (its `NOT EXISTS` requires no present file, which `episode_events` requires).

Every row therefore has `target_rank = 1`. The window function adds a partition+sort over the whole event set on every Recently Added page load for no behavioral benefit, and a future reader has to re-derive the disjointness proof to know it is safe to touch. `filtered` can select straight from `all_events`.

internal/api/handlers/sections.go
```diff
@@ -1284,6 +1307,17 @@ func (h *SectionHandler) buildSectionsResponse(r *http.Request, withItems []sect
 	return resp
 }
 
+func sectionRequestLibraryIDs(r *http.Request) []int {
```
Quick104: **Library scope re-derived by sniffing `r.URL.Path` for "/library/"**

Every caller of `buildSectionsResponse` already parsed a validated `libraryID` a few lines earlier (`HandleLibrarySections` at 643, `HandleLibrarySectionItems` at 665). Recovering it here by substring-matching the URL and re-parsing `chi.URLParam(r, "id")` is a second, weaker source of truth.

If the route is ever remounted (`/libraries/{id}/sections`) or the param renamed, play targets silently stop being library-scoped — no error, no failing test, no compile break. Pass `libraryID` down instead.

internal/catalog/recent_tv.go
```diff
@@ -0,0 +1,412 @@
...
+		if _, ok := seen[value]; ok {
+			continue
+		}
+		seen[value] = struct{}{}
+		result = append(result, value)
+	}
+	return result
+}
+
+func sortedUniqueInts(values []int) []int {
```
Quick104, line 562: **Hand-rolled insertion sort and duplicate int-set helpers**

`sortedUniqueInts` implements an insertion sort by hand where `slices.Sort` does the job, and `intersectOptionalInts` duplicates the package's existing `intersectInts` (`internal/catalog/query_executor.go:510`).

Four new int-set helpers land in a package that already has one, so the intersection semantics now live in two places and can drift.

internal/api/handlers/catalog_resources.go
```diff
@@ -335,15 +337,17 @@ func (h *CatalogResourceHandler) HandleGetSeason(w http.ResponseWriter, r *http.
...
-				),
-			})
+			resp := h.items.toSeasonResponseFromEpisodes(
+				r,
+				id,
+				season,
+				episodes,
+				h.items.getAggregateUserData(r, episodes),
+			)
+			seasonResponses := []seasonResponse{resp}
```
Quick104: **Single season wrapped in a slice just to enrich it**

Both single-season branches wrap one `seasonResponse` into a slice, call `enrichSeasonPlayTargets`, then unwrap `seasonResponses[0]` — the same three lines at 347-349 and 381-383.

A third single-season path means copying it again, and a reader has to verify `seasonResponses[0]` is still the same value. A `resolveSeasonPlayTarget(r, seriesID, *seasonResponse)` helper (or having the enricher take `[]*seasonResponse`) removes both copies.

web/src/components/GlobalSearch.tsx
```diff
@@ -249,26 +291,22 @@ export function GlobalSearch({
...
-                    items.length === 0 ? -1 : prev < items.length - 1 ? prev + 1 : 0,
-                  );
+                  moveResultFocus(selectedIndex + 1);
                 } else if (e.key === "ArrowUp") {
                   e.preventDefault();
-                  setSelectedIndex((prev) =>
-                    items.length === 0 ? -1 : prev <= 0 ? items.length - 1 : prev - 1,
-                  );
+                  moveResultFocus(selectedIndex < 0 ? items.length - 1 : selectedIndex - 1);
                 } else if (e.key === "Enter" && selectedIndex >= 0 && items[selectedIndex]) {
```
Quick104, line 320: **Unreachable Enter branch on the search input**

This branch is guarded by `selectedIndex >= 0`, but `selectedIndex` only becomes non-negative through `moveResultFocus`, which immediately moves DOM focus to the result button. The input can therefore never receive an Enter event while the guard holds.

`handlePickItem(items[selectedIndex].content_id)` at line 312 is dead: Enter on a selected result is handled by the button's native click, and Enter with no selection falls through to form submission. It reads as a second implementation of result activation that no test can exercise.

(If finding #5 is fixed by restoring `aria-activedescendant` virtual focus, this branch becomes live again and should stay.)

Quick104:
Pushed four performance commits after testing this PR on the shared dev deployment (1.16M `episode_libraries` rows, 794K in the TV library; 1.77M `media_files`; 2.1M `episodes`). Home/section `items` requests were taking 13–50s and `/api/v1/catalog` 5–6s.

**`956d797b0` — playable-target leaf lookups**
The `leaf_targets` CTE matched movie and episode files through one `CASE` expression, which PostgreSQL cannot drive through either `media_files` index — every movie/episode card scanned the table. `EXPLAIN (ANALYZE, BUFFERS)` on dev showed a nested-loop semi join discarding 6.87M rows for an 8-card page, 10.1s. Splitting into type-specific `EXISTS` branches restores index probes: **10,103ms → 2.2ms**, identical results.

**`c820a8ffa` + `39c587fc7` — recently-added TV events**
`RecentTVRepository.List` walked every availability row twice (event aggregate + `series_without_episode_events` anti-join), ~4.7s per uncached request. First commit shares one materialized pass between both branches (4.7s → 3.3s). Second commit splits the no-prefix path in two phases: a narrow `MAX`-only aggregate (fuses into a parallel hash aggregate — no 794K-row sort) finds and ranks every event, then the expensive columns (episode counts, anchor picks) are computed only for the requested page. `rank()` keeps `added_at` ties together so the fully tiebroken page ordering is unchanged. **3.3s → 2.0s**; result-identical on a 500-row page under a pinned snapshot. The name-prefix path keeps the single-pass shape because its per-episode-title filter changes which rows belong to each event.

**`3c82a4820` — parameter typing**
`added_rank <= $n + $m` fails with untyped bind parameters (`operator is not unique: unknown + unknown`, SQLSTATE 42725); both casts to `bigint`. Found live on dev — SQL prototypes with literals can't catch this, which is a good argument for running the `SILO_TEST_DATABASE_URL`-gated tests on this PR before merge.

**Live results on shared dev** (server-side `duration_ms`): section items cache-miss 13–50s → **2.0–2.2s**, warm 170–594ms; `/api/v1/catalog` 5–6s → ~0.6s.

Remaining discussion for this PR: the recent-TV query still makes one full pass over the library's availability rows per cache miss (irreducible for exact results in the current schema). Options if that matters at scale: a covering index (`episodes(content_id) INCLUDE (series_id)`), denormalizing `series_id` into `episode_libraries`, or scanner-maintained incremental event rows. Also observed two concurrent cache misses both recomputing — single-flight on the resolved-list cache would hide scan-time invalidation.

Commands run: `go build`, `go vet`, `go test ./internal/catalog/` (DB-gated tests skip without `SILO_TEST_DATABASE_URL` — **not run**), `golangci-lint run --new-from-merge-base=origin/main ./internal/catalog/...` (0 issues), live deployment verification above. Not run: full `make test-go`, DB-backed catalog tests.

## AI Disclosure

- Tool(s): Claude Code
- Model(s): claude-fable-5
- Involvement: Fully AI-generated, human-directed (maintainer line-by-line review pending)
- Adversarial review: result-identity diffs of old vs. new SQL on live dev data (500-row pages, snapshot-pinned against concurrent scan writes), `EXPLAIN (ANALYZE, BUFFERS)` verification of every plan claim, live deployment measurement of each commit, and a parameterized-path failure (`3c82a4820`) found and fixed during that live verification.

PR #593, opened 2026-08-11
fix(catalog): bound Next Up anchors by distinct series, not rows

### LINKED ISSUE #592: Next Up rail wiped out after bulk mark-watched (row-capped anchor scan evicts all other series)
## Symptom

Reported by user vanbaster (web UI): marking one series watched from Continue Watching made the **entire Next Up row disappear**, and it stayed gone across logout/login.

## Root cause

`buildListNextUpQuery` bounds the global anchor scan with `nextUpAnchorMaxRows = 500` — a cap on **raw completed rows**, introduced in #350 to fix a real 44.7s regression. But the query only needs **one anchor per series**. A bulk mark-watched writes hundreds of completed rows with the newest timestamps; those rows flood the 500-row window and evict every other series.

Production evidence (2026-08-10 ~18:00 UTC, profile `valfam` of user 134):

- Marking SpongeBob SquarePants watched wrote **572 completed rows in <2s** (17:59:59.965 → 18:00:01.718).
- The 500-row anchor window afterwards contained **exactly 1 distinct series** (the fully-watched one, which yields no next episode) → rail returned 0 rows.
- Replaying the query unbounded returned the user's real 16-row rail — no data was lost, only unreachable.

The cap also degrades rails organically: on another prod profile (11k completed rows), the row-capped query returns **2** Next Up rows where a series-bounded scan returns the full **24**.

## Fix direction

Replace the row-capped anchor CTE with a recursive skip-scan (loose index scan) over `idx_uwp_profile_completed` that emits the newest row of each not-yet-seen series and stops after `96` **distinct series**. One series can then never evict others. Compound `(updated_at, media_item_id)` cursor so same-timestamp bulk writes across different series aren't skipped. Series-scoped branch (show-detail tile) unchanged.

Measured on prod (PG 18.4), full generated query: worst profile (23k rows/271 series) ~210ms vs ~84ms old (acceptable; the correctness bug is disqualifying); typical heavy profile 102ms new vs 181ms old, returning 24 rows vs 2.

### AI Disclosure
- Tool(s): Claude Code, Codex CLI
- Model(s): claude-fable-5 (diagnosis, review), gpt-5.6-sol (implementation)
- Involvement: fully AI-generated, human-operated
- Adversarial review: see linked PR

Quick104:
Fixes #592 · Part of the Next Up / sections reliability work (follow-up to #350's perf fixes)

## Problem

Marking one long series watched wiped a user's entire Next Up rail (prod incident 2026-08-10, user report + server-side corroboration in #592). `buildListNextUpQuery` bounds the global anchor scan at `nextUpAnchorMaxRows = 500` **raw completed rows**; a bulk mark-watched wrote 572 completed rows in <2s, all newer than everything else, so the 500-row window contained one series and the rail collapsed to 0 rows. No data was lost — the rail was just unreachable behind the cap.

The cap also silently truncates rails organically: a prod profile with 11k completed rows gets 2 Next Up rows from the capped query vs the full 24 without it.

## Approach

Bound by **distinct series** instead of rows. The anchor CTE becomes a recursive skip-scan (loose index scan) walking `idx_uwp_profile_completed` newest→oldest, emitting the first row of each not-yet-seen series, stopping at `nextUpAnchorMaxSeries = 96` (4× the ~24-item rail). One series can then never evict others, regardless of how many of its rows are newest.

Load-bearing details:

- **Compound `(updated_at, media_item_id)` cursor** in seed + step. Bulk marks and history imports stamp many rows across different series with one timestamp; a bare `updated_at <` cursor would skip those series. Pinned by a DB-backed test.
- Hidden-items filter and `DateCutoff` apply inside **both** seed and recursive step, so excluded rows neither anchor nor advance the cursor.
- The series-scoped branch (show-detail tile) is untouched — intentionally unbounded, anchors on the series' last completed episode regardless of age.
- Downstream SQL (eligible_series exclusion, next-episode LATERAL) unchanged; the walk feeds the same `completed_episodes` CTE name.

## Performance (measured on prod data, PG 18.4, read-only)

| Profile | Old (row cap) | New (series walk) |
|---|---|---|
| Flooded incident profile | ~15ms, **0 rows (the bug)** | 0.5ms, correct rail |
| Heavy organic profile (11k rows, 293 series) | 181ms, **2 rows (truncated)** | 102ms, full 24 rows |
| Worst bulk-import profile (23k rows, 271 series) | 84ms | 210ms |

The worst-profile regression is the cost of actually finding 96 distinct series instead of giving up after 500 rows; both are far under the 44.7s that motivated #350. Worst case with fewer than 96 completed series is one ordered walk of the profile's completed partial index.

## Verification

- `go build ./...`, `go vet`, `gofmt`/`goimports` clean; `golangci-lint run --new-from-merge-base=origin/main ./internal/catalog/...` → 0 issues.
- Unit tests updated: global query asserts `WITH RECURSIVE`, compound cursor, `ANY(w.seen)` skip, series bound; series-scoped test asserts its SQL didn't change shape.
- New DB-backed regression tests (skip without `SILO_TEST_DATABASE_URL`), run against a freshly-migrated Postgres 18 sandbox:
  - `TestNextUpRepository_GlobalAnchorWalkSurvivesFloodedSeries` — 600 same-timestamp completed rows of one series must not evict two older series. **Fails against the old query (returns 0 rows), passes with the walk.**
  - `TestNextUpRepository_GlobalAnchorWalkKeepsSameTimestampSeries` — two series completed at an identical timestamp both anchor (pins the compound cursor).
- Pre-existing failure `TestEpisodeSearchPostgresAndDocumentSource` (`_vectors` opt-out) fails identically on clean `origin/main` against the same DB — unrelated.
- Generated SQL (not hand-transcribed) EXPLAIN ANALYZEd read-only against the production profiles in the table above.

## Risks / follow-ups

- Worst-profile latency regressed 84→210ms. If the tail grows, the escape hatch is a scan-depth cutoff, not a row cap. The `slow query` log is the tripwire.
- Within a bulk-marked series where all timestamps are equal, the anchor's tie-break is `media_item_id`, not episode order — the surfaced episode is still always unwatched (downstream LATERAL skips completed/in-progress), only *which* unwatched episode of that series wins can differ. Noted in a code comment.
- `internal/sections/fetcher.go`'s Continue Watching scan caps (`continueProgressMaxScanned`) share the raw-row-bound shape but have no bulk-write path to flood them; left alone.

### AI Disclosure
- Tool(s): Claude Code (orchestration, diagnosis, review, verification), Codex CLI (implementation)
- Model(s): claude-fable-5, gpt-5.6-sol
- Involvement: fully AI-generated, human-operated
- Adversarial review: Claude reviewed the Codex diff line-by-line against the brief. Found and verified acceptable: the WITH RECURSIVE prefix is emitted via the shared `WITH %s` format string (series-scoped branch stays plain `WITH`); the same `$4` placeholder is legally referenced twice for DateCutoff; confirmed the flood regression test fails against the old query by stashing the fix and re-running. The worst-profile perf regression was measured and judged acceptable versus the correctness bug; documented above rather than hidden.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the global “Next Up” list so series with many recent completions no longer crowd out other series.
  * Hidden and outdated items are now consistently excluded while building recommendations.
  * Results with matching timestamps are ordered consistently for more reliable display.
  * Global results continue scanning eligible history when needed, up to a safe limit.
  * Series-specific “Next Up” views continue to return complete results without the global limit.

### DIFF
--- a/internal/catalog/nextup_repo.go
+++ b/internal/catalog/nextup_repo.go
@@ -109,20 +109,17 @@ func (r *NextUpRepository) ListNextUp(ctx context.Context, q NextUpQuery) ([]Nex
 	return results, nil
 }
 
-// nextUpAnchorMaxRows bounds how many of the profile's most recent completed
-// rows the global next-up query considers when deriving per-series anchors.
-// Without a bound the completed_episodes CTE re-reads the profile's entire
-// completed history per call — 233k rows joined to episodes for the worst
-// bulk-import profile, measured at up to 44.7s. A next-up rail surfaces ~24
-// series; the 500 most recently completed episodes cover every series that can
-// realistically rank on it (a series whose last completed episode is older
-// than 500 completions of other content sorts far past the rail's limit).
-// Progress rows are recency-ordered on idx_uwp_profile_completed, so the
-// bounded anchor scan is an ordered index walk. Series-scoped calls (the
-// show-detail tile) stay unbounded: they must anchor on the series' last
-// completed episode no matter how long ago it was watched, and are naturally
-// bounded by one series.
-const nextUpAnchorMaxRows = 500
+// nextUpAnchorMaxSeries bounds the number of distinct series anchors the global
+// next-up query derives. A row-based cap let a bulk mark-watched operation flood
+// the window with one series and evict every other series; counting distinct
+// series makes that impossible regardless of how many of one series' rows are
+// newest. The rail surfaces about 24 items, so 96 anchors leave ample headroom.
+// Progress rows are recency-ordered on idx_uwp_profile_completed; when a profile
+// has fewer than 96 completed series, the worst case is one full ordered walk of
+// that partial index. Series-scoped calls (the show-detail tile) stay unbounded:
+// they must anchor on the series' last completed episode no matter how long ago
+// it was watched, and are naturally bounded by one series.
+const nextUpAnchorMaxSeries = 96
 
 func buildListNextUpQuery(q NextUpQuery, limit int) (string, []interface{}) {
 	args := []interface{}{q.UserID, q.ProfileID, limit}
@@ -169,11 +166,16 @@ func buildListNextUpQuery(q NextUpQuery, limit int) (string, []interface{}) {
 		sourceTable = "eligible_series"
 	}
 
-	// Global queries derive series anchors from a bounded recent-completions
-	// scan (see nextUpAnchorMaxRows); series-scoped queries keep the unbounded
-	// shape so the anchor is the series' last completed episode regardless of
-	// age. The hidden-items exclusion and date cutoff apply inside the bounded
-	// scan so hidden/old rows never consume the anchor budget.
+	// Global queries use a recursive loose-index scan to find at most
+	// nextUpAnchorMaxSeries distinct series without letting one series' rows
+	// consume the budget. The compound cursor is the total order required when
+	// bulk updates give many rows the same timestamp. media_item_id is only a
+	// deterministic tie-break, not episode order; for equal timestamps the
+	// downstream lookup still skips every completed or in-progress episode.
+	// Series-scoped queries keep the unbounded shape so the anchor is the
+	// series' last completed episode regardless of age. Hidden rows and the date
+	// cutoff are excluded inside every walk step so they neither become anchors
+	// nor advance the cursor.
 	var completedEpisodesCTE string
 	if q.SeriesID != "" {
 		completedEpisodesCTE = fmt.Sprintf(`completed_episodes AS (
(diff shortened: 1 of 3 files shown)


Quick104:
Addressed the P1 in ef7a21b1a — the walk now continues until enough **eligible** series are found, not just visited.

**Approach: batched continuation in Go, not eligibility-in-SQL.** I first prototyped pushing the eligibility test (next-episode LATERAL + newer-partial-progress exclusion) into every walk step so the recursion could stop on eligible-count directly. Measured on the worst production profile it costs **~1.2s** (the eligibility probe runs per visited series, including all the ineligible ones) — worse than the disease. Instead, each batch keeps the cheap 96-visited-series walk but now reports its frontier (compound cursor position + seen-series array + rows walked), and `ListNextUp` loops: run a batch, keep eligible results, resume below the frontier until the requested `Limit` is filled or history is exhausted.

Loop invariants:
- **Pagination works**: jellycompat's `Limit + StartIndex` is filled across batches, not capped at one walk.
- **Termination**: frontier `n < 96` ⇒ the walk ran out of history before its budget ⇒ terminal. No frontier row at all ⇒ empty history ⇒ terminal.
- **Runaway guard**: 10 batches max (960 visited series per call), and hitting the cap with results still short logs a warning with profile/batches/results — no silent truncation.
- **Order preserved**: every batch resumes strictly below the previous compound `(updated_at, media_item_id)` frontier, so appended batches keep the rail's recency order.
- Series-scoped queries (show-detail tile) keep their single-shot shape, byte-identical SQL — pinned by test.

**Verification** (same sandbox DB + read-only prod methodology as the PR body):
- New DB-backed regression: 100 fully-watched (ineligible) series newest in history must not hide 2 older series with eligible next episodes — **fails on the previous commit (returns 0 rows), passes now**; verified by running the new test against `da84cbe` in a scratch worktree.
- Exhaustion test: an all-caught-up profile returns empty without spinning to the batch cap.
- Generated batch-1 SQL EXPLAINed on the worst prod profile: 234ms, frontier row correctly reports `n=96` + 96-series seen array; subsequent batches only run when a page is short.
- `go build`, `go vet`, gofmt clean; `golangci-lint --new-from-merge-base` → 0 issues; full catalog suite green except the pre-existing `TestEpisodeSearchPostgresAndDocumentSource` failure that reproduces on clean main.

One honest caveat: a profile whose recent history is a long run of caught-up series pays one extra ~200ms batch per 96 ineligible anchors until it finds enough eligible ones (bounded at 10). That's the cost of correctness here; the warning log is the tripwire if real profiles ever hit the cap.

*(AI disclosure as in the PR body: claude-fable-5 orchestrating/reviewing, gpt-5.6-sol implementing; the timestamptz cast fix in the exhaustion test's seed SQL was mine.)*

Quick104:
Pushed 8d5f3a269 addressing the remaining review findings. Note ef7a21b1a landed while I was working, which already covers the empty-rail and pagination findings via batching — this commit is scoped to what that didn't reach.

**Anchor ordering (blocking finding).** The walk ordered by `(updated_at, media_item_id)` then `(season, episode)`. Since `media_item_id` is unique, nothing ever ties on that pair, so the season/episode clause was unreachable — for a bulk mark-watched series (all rows one timestamp) the anchor was just the highest `content_id`. With production-shape 18-digit IDs where season 2 was scanned before a season-1 backfill, that surfaced **s01e04 instead of s02e04** — an episode following one already watched.

Fixed by separating the two orderings: the walk still advances on the compound cursor (the total order it needs), and a lateral picks the highest `(season, episode)` among rows sharing `pick.updated_at`. Pinning `updated_at` rather than re-sorting keeps it an index probe; both rows carry the same `updated_at`, so cursor order and `CompletedAt` are unchanged.

**Cursor index.** Added `idx_uwp_profile_completed_cursor`. The existing index stops at `updated_at`, so the `media_item_id` half of the seek was a filter and each step re-read every row tied on one timestamp — exactly the bulk mark-watched shape. Dropping/recreating it around the shipped query on a synthetic 40k-row profile: **341ms → 224ms**, same plan shape.

**Test coverage.** New DB regression seeds the scanned-out-of-order ID shape and fails on the previous ordering (returns s01e04, wants s02e04); a SQL-shape test pins that `media_item_id` can't gate the episode choice again.

All verified against a live Postgres 18.3 sandbox, not inferred. Full `NextUp` suite passes. Two unrelated failures exist on this branch and on `da84cbe3e` alike — `TestEpisodeSearchPostgresAndDocumentSource` and the jellycompat `TestBeginWebOperation*` pair.

**One finding I did not act on:** the per-anchor in-progress suppression scan. In the batched structure it runs once over the ≤96-row CTE rather than per walk step, so the cost I originally measured doesn't apply here. I'd rather leave it than restructure working code on a stale premise — happy to revisit if you'd like it hoisted anyway.

PR #620, opened 2026-08-13
fix(playback): route v3 direct play and remux through proxy nodes

### LINKED ISSUE #619: Protocol v3 direct-play and progressive-remux sessions bypass proxy nodes
## Summary

In a multi-node deployment, protocol v3 direct-play and progressive-remux sessions are served
by the API server itself instead of being delegated to a proxy node. Only the HLS deliveries
(`transcode_hls`, `remux_hls`) consult the node planner, so an operator who deploys dedicated
proxy nodes still sees all direct-play and remux video egress on the API node.

The proxy node already implements the endpoints these routes need, and the Jellyfin-compat
transport already redirects to them — native v3 is the only surface that skips them.

## Source

Internal Discord report from a self-hosting user.

## Steps to Reproduce

1. Deploy a fresh install with three nodes: one API node, one proxy node, one transcode node
   (proxy and transcode grouped together).
2. Disable transcoding on the API node.
3. Start playback of a title whose plan resolves to `original_http` (direct play) or
   `server_remux_progressive` — e.g. an H.264/E-AC-3 MKV, which resolves to a progressive
   remux with an `audio_to_aac` server transformation.
4. Observe the `stream.url` in the v3 plan response and the node actually serving the bytes.

## Expected Behavior

Direct-play and progressive-remux sessions are planned onto a proxy node and the returned
`stream.url` points at that proxy, matching the behavior of the HLS deliveries and of the
Jellyfin-compat transport.

## Actual Behavior

The v3 plan returns an API-relative stream URL (`/stream/{session_id}`) and the API node serves
the video: `ServeDirectPlay` for direct play, and a locally spawned ffmpeg for the progressive
remux. The node planner is never consulted for these deliveries.

Reported plan (abridged) from an API-only node:

```
"delivery": "server_remux_progressive",
"stream": { "url": "/stream/<session-id>?st=...", "protocol": "http_progressive" },
"transformations": [ { "name": "audio_to_aac", "executor": "server", "recipe_version": "1" } ]
```

## Evidence

- `internal/api/handlers/playback_v3.go` — `prepareTransportV3` returns
  `prepareIdentityTransportV3` for every delivery that is not `transcode_hls` or `remux_hls`,
  before the `h.NodePlanner` block is reached. `planNodeSessionV3` is therefore only ever called
  for the two HLS deliveries.
- `internal/api/handlers/playback.go` — `prepareIdentityTransportV3` builds the URL via
  `playbackStreamURL`, which always returns the API-local `/stream/{session_id}` path.
- `internal/api/handlers/stream.go` — `HandleStream` serves `PlayDirect` with
  `playback.ServeDirectPlay` and `PlayRemux` with `playback.ServeRemuxWithOptions`, both in the
  API process.
- `internal/proxy/server.go` — the proxy node already registers `/stream/direct/{token}` and
  `/stream/remux/{token}` with equivalent handlers, including seek and Dolby Vision mode
  handling frozen into the stream token.
- `internal/jellycompat/streams.go` / `handlers_playback.go` — the compat transport calls
  `PlanSession(..., needsTranscode=false, ...)` and redirects to
  `proxyNode.URL + "/stream/direct/"` or `"/stream/remux/"`. The capability exists and is wired
  on that surface only.
- Closed issue #326 fixed the same class of problem for downloads; v3 playback was not covered.

## Secondary Defect

`playback.local_transcode_fallback = false` does not prevent the API node from running ffmpeg
for a progressive remux. `nodepool.LocalTranscodeFallbackAllowed` is only consulted inside the
HLS branch of `prepareTransportV3`, so an audio-converting progressive remux (executor
`server`) still spawns locally on an API-only node even with the setting disabled. Whatever
shape the primary fix takes, this gate should cover the progressive-remux path too.

## Scope

Owning project: Silo Server. No client change should be required — the plan's `stream.url` is
already permitted to be absolute (the HLS deliveries return absolute proxy URLs today), so
clients that follow the plan need no protocol change.

Design notes for whoever picks this up:

- `PlanSession(sessionID, "", needsTranscode=false, estBitrateKbps)` is the existing call shape
  for a proxy-only plan; `internal/jellycompat/streams.go` uses exactly this.
- Bandwidth admission wants a real estimate: the source bitrate for direct play, and the
  effective recipe bitrate for a remux.
- The stream token minted for the proxy must carry the same reconstruction claims the v3
  identity card carries today (audio track, `transcode_audio`, target audio channels/bitrate,
  Dolby Vision remux mode, audio-only flag), or a proxy-served remux will diverge from the
  plan's promised output.
- Falling back to the local path when no proxy is eligible preserves single-node deployments;
  that fallback should honor the `local_transcode_fallback` gate for the remux case.

## Acceptance Criteria

- [ ] A v3 plan with delivery `original_http` returns a proxy-node stream URL when a healthy,
      under-cap proxy exists, and the proxy serves the byte-range transfer.
- [ ] A v3 plan with delivery `server_remux_progressive` returns a proxy-node stream URL when a
      healthy proxy exists, and the remux ffmpeg runs on the proxy, not the API node.
- [ ] Seek (`?seek=`) and Dolby Vision remux mode behave identically on the proxy-served remux
      as on the API-served one.
- [ ] With no eligible proxy, behavior falls back to the current API-local path for direct play;
      for the remux path the fallback is refused when `playback.local_transcode_fallback` is
      `false`.
- [ ] Proxy bandwidth and job caps are respected for these sessions, and the reservation is
      released when transport setup fails.
- [ ] Tests cover proxy selection for both deliveries and the no-proxy fallback.

Quick104:
## Problem

In a multi-node deployment, protocol v3 direct-play and progressive-remux sessions were served by
the API server itself instead of being delegated to a proxy node. `prepareTransportV3` consulted
the node planner only for the two HLS deliveries; `original_http` and `server_remux_progressive`
returned early with an API-local `/stream/{session_id}` URL, and `HandleStream` then served the
bytes (`ServeDirectPlay`) or spawned a local ffmpeg (`ServeRemuxWithOptions`) in the API process.

An operator running a dedicated proxy node still saw all direct-play and remux egress land on the
API node — which is what the reporting user hit on an api/proxy/transcode split with transcoding
disabled on the API node.

## Why this approach

The capability already existed and was already proven on another surface:

- The proxy node implements `/stream/direct/{token}` and `/stream/remux/{token}`
  (`internal/proxy/server.go`).
- The Jellyfin-compat transport already calls `PlanSession(..., needsTranscode=false, ...)` and
  redirects to exactly those routes (`internal/jellycompat/streams.go`).

Native v3 was the only surface skipping it, so on the same deployment a Jellyfin client routed
correctly while Silo's own clients did not. This wires the existing shape into the v3 identity
transport rather than inventing a second node-selection path.

No client change is required: `stream.url` is already allowed to be absolute (the HLS deliveries
return absolute proxy URLs today), and the web player resolves absolute URLs already.

## Token claims

The proxy reconstructs from the stream token alone, so the token now carries three things a
`RecipeCard` does not model:

| Claim | Why it must travel |
| --- | --- |
| `MediaPath` | The proxy has no DB; without it there is nothing to open. |
| `DVProfile` | A Profile 7 remux must strip the dangling RPU. |
| `AudioOnly` | Selects `audio/mp4` over `video/mp4` — the MIME the plan promised. |

None of these fail loudly when missing. They would produce a stream subtly different from the one
the plan promised, which is why they are set explicitly rather than left to `ToClaims()`.

## Related fixes in this PR

**Proxy direct-play parity.** The proxy served direct play with `http.ServeFile`, which sets no
strong ETag. `direct_stream_resume_v1` depends on the ETag `ServeDirectPlay` sets *before*
`ServeContent` — without it `If-Range` never validates and a resumed range silently restarts at
200. Routing direct play to a proxy without this would have broken resumable direct streams on
exactly the deployments this PR targets. The proxy now uses the same serve path (which also brings
the rolling write deadline and stream metrics with it).

**`playback.local_transcode_fallback` gap.** The gate was only consulted inside the HLS branch, so
an audio-converting progressive remux still spawned ffmpeg locally on an API-only node with the
setting disabled. Identity deliveries now honor it too. Direct play deliberately still falls back
locally — moving bytes is not transcode work, and single-node deployments must keep working.

## Risks / follow-up

- When no proxy is eligible (unhealthy, at cap, no planner, no signing secret) behavior falls back
  to the current API-local path, so single-node deployments are unchanged.
- A planner reservation is released whenever the session does not actually reach a proxy, so
  repeated failed starts cannot pin a node's job/bandwidth budget until the reservation ages out.
- Bandwidth admission uses the plan's effective recipe bitrate, falling back to the source bitrate.
- Not covered here: the HLS deliveries already routed correctly and are untouched.

## Verification

```
go build ./...                                    # OK
go test ./internal/api/handlers/ ./internal/proxy/ ./internal/playback/ ./internal/nodepool/
golangci-lint run --new-from-merge-base=origin/main ./internal/api/handlers/... ./internal/proxy/...   # 0 issues
gofmt -l internal/                                # clean
```

Six new tests cover: proxy routing for direct play (incl. reservation session id and bitrate
estimate), proxy routing for progressive remux with seek + DV profile preserved, local fallback
with no eligible proxy, the remux refusal when local fallback is disabled, direct play still
allowed when it is disabled, and reservation release on rollback. **All six were confirmed to fail
against pre-fix behavior** and pass with the change.

One pre-existing failure is present and is **not** from this branch:
`TestHandleReplanPlaybackV3BitmapSubtitleFallsBackFromHDRToSDRVersion` fails identically on a
pristine checkout of `ee9356aab` (#617) with none of this branch's code. Worth a separate look.

Closes #619

## AI-use disclosure

Implemented by Claude Opus 5 (Claude Code) on behalf of a maintainer. The report was triaged from
an internal Discord thread; the root cause, the token-claim gaps, and the two related defects were
identified by reading the code, and every claim above was verified by execution as shown.

## Summary by CodeRabbit

* **New Features**
  * Improved direct playback with resumable downloads, stronger cache validation, stream metrics, and rolling write timeouts.
  * Added proxy routing for direct playback and progressive remux sessions.
  * Added signed playback URLs containing media, seek, Dolby Vision, and audio-transcoding details.
  * Added bitrate-aware admission, capability validation, and automatic local fallback.
  * Added hardware and transformation capability reporting for authenticated proxy requests.
  * Extended remote playback sessions’ idle grace period while retaining expiration safeguards.

* **Bug Fixes**
  * Proxy reservations are now released when playback setup fails or falls back locally.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 811c85d5016fb1532c074a73b57c7e5df37cb5e1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/design/schemas/playback-v3/v3/fixtures/valid/capability_response.json
+++ b/docs/design/schemas/playback-v3/v3/fixtures/valid/capability_response.json
@@ -10,6 +10,7 @@
     "playback_route_diagnostics",
     "device_quirks_v1",
     "seek_reanchor_v1",
+    "output_change_v1",
     "direct_stream_resume_v1",
     "plan_source_duration_v1"
   ],
--- a/docs/design/schemas/playback-v3/v3/fixtures/valid/decision_response.json
+++ b/docs/design/schemas/playback-v3/v3/fixtures/valid/decision_response.json
@@ -7,6 +7,7 @@
     "playback_route_diagnostics",
     "device_quirks_v1",
     "seek_reanchor_v1",
+    "output_change_v1",
     "direct_stream_resume_v1",
     "plan_source_duration_v1"
   ],
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -49,6 +49,7 @@ type SessionManagerInterface interface {
 	TouchActivity(sessionID string) error
 	BeginTransport(sessionID string) error
 	EndTransport(sessionID string) error
+	SetRemoteTransport(sessionID string, remote bool) error
 	SetEffectiveMediaFileID(sessionID string, fileID int) error
 	SetTranscodeNodeURL(sessionID, url string) error
 	SetTranscodeRoute(sessionID string, route playback.TranscodeRoute) error
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -396,12 +396,20 @@ const streamTokenParam = "st"
 // reconstruction recipe. Returns "" when no signing secret is configured
 // (reconstruct effectively disabled, e.g. in tests).
 func (h *PlaybackHandler) signSessionToken(card playback.RecipeCard) string {
+	return h.signStreamClaims(card.ToClaims())
+}
+
+// signStreamClaims mints a stream token from claims that are already assembled.
+// Callers serving a session from another node use it to add the claims a
+// RecipeCard does not model (the file's Dolby Vision profile, audio-only flag),
+// which a remote executor cannot look up for itself.
+func (h *PlaybackHandler) signStreamClaims(claims streamtoken.Claims) string {
 	if h.JWTSecret == "" {
 		return ""
 	}
-	token, err := streamtoken.Sign(card.ToClaims(), h.JWTSecret, playback.MaxTokenTTL)
+	token, err := streamtoken.Sign(claims, h.JWTSecret, playback.MaxTokenTTL)
 	if err != nil {
-		slog.Warn("sign stream token failed", "error", err, "session", card.SessionID, "playback_session_id", card.SessionID)
+		slog.Warn("sign stream token failed", "error", err, "session", claims.SessionID, "playback_session_id", claims.SessionID)
 		return ""
 	}
 	return token
--- a/internal/api/handlers/playback_test.go
+++ b/internal/api/handlers/playback_test.go
@@ -181,6 +181,8 @@ func (failingSessionManager) BeginTransport(string) error { return nil }
 
 func (failingSessionManager) EndTransport(string) error { return nil }
 
+func (failingSessionManager) SetRemoteTransport(string, bool) error { return nil }
+
 func (failingSessionManager) SetEffectiveMediaFileID(string, int) error { return nil }
 
 func (failingSessionManager) SetTranscodeNodeURL(string, string) error { return nil }
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -672,7 +672,7 @@ func (h *PlaybackHandler) prepareTransportV3(r *http.Request, session *playback.
 		return preparedTransportV3{}, timelineErr
 	}
 	if result.Plan.Delivery != playback.DeliveryTranscodeHLSV3 && result.Plan.Delivery != playback.DeliveryRemuxHLSV3 {
-		return h.prepareIdentityTransportV3(session, result, timeline), nil
+		return h.prepareIdentityTransportV3(r, session, file, result, timeline)
 	}
 	if h.NodePlanner != nil {
 		plan := h.planNodeSessionV3(r.Context(), session, result)
@@ -777,7 +777,7 @@ func planRequiresServerTransformationsV3(plan *playback.PlanV3) bool {
 	return false
 }
 
-func (h *PlaybackHandler) prepareIdentityTransportV3(session *playback.Session, result playback.PlannerResultV3, timeline preparedTimelineV3) preparedTransportV3 {
+func (h *PlaybackHandler) prepareIdentityTransportV3(r *http.Request, session *playback.Session, file *models.MediaFile, result playback.PlannerResultV3, timeline preparedTimelineV3) (preparedTransportV3, *transportErrorV3) {
 	routeSession := *session
 	routeSession.PlayMethod = result.PlayMethod
 	routeSession.BasePlayMethod = result.PlayMethod
@@ -788,11 +788,31 @@ func (h *PlaybackHandler) prepareIdentityTransportV3(session *playback.Session, 
 	routeSession.TargetAudioChannels = result.TargetAudioChannels
 	routeSession.TargetAudioBitrateKbps = result.TargetAudioBitrateKbps
 	routeSession.RemuxDVMode = remuxDVModeForPlanV3(result.Plan)
+
+	proxyNode, proxyErr := h.planIdentityProxyV3(r, session.ID, result)
+	if proxyErr != nil {
+		return preparedTransportV3{}, proxyErr
+	}
+	streamURL, servedByProxy := h.identityStreamURLV3(&routeSession, file, proxyNode)
+	releaseProxyReservation := func() {
+		if releaser, ok := h.NodePlanner.(sessionReservationReleaserV3); ok {
+			releaser.ReleaseSession(session.ID)
+		}
+	}
+	if proxyNode != nil && !servedByProxy {
+		// A planned proxy that could not be addressed (no signable token, no
+		// file record) falls back to the local path, so its reservation must be
+		// dropped now rather than pinning that node's budget until it ages out.
+		releaseProxyReservation()
+		if err := h.refuseLocalIdentityWorkV3(r, result); err != nil {
+			return preparedTransportV3{}, err
+		}
+	}
+
 	previousNodeURL := session.TranscodeNodeURL
 	previousTransportID := remoteTransportID(session)
 	unlock := h.tm.LockSessionLifecycle(session.ID)
 	committed := false
-	streamURL := h.playbackStreamURL(&routeSession)
 	if result.Plan != nil && result.Plan.Delivery == playback.DeliveryRemuxProgressiveV3 {
 		if seek := timeline.seekSeconds; seek > 0 {
 			streamURL = appendPlaybackQueryV3(streamURL, "seek", strconv.FormatFloat(seek, 'f', -1, 64))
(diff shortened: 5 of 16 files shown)


Quick104:
All three P1 findings were valid and are fixed in 811c85d5. Thanks — the capability one in particular was a genuine hole in the design, not just a hardening nit.

**Do proxy nodes even have remux capabilities?** Yes, but not *provably*, which was the actual bug. The proxy really does run ffmpeg (`handleRemux` → `playback.ServeRemuxWithOptions` on `cfg.Playback.FFmpegPath`), and in the standard deployment it is the same image and the same `jellyfin-ffmpeg7` as the API node — mode is only a config switch. But the proxy exposed no `/hw-capabilities` endpoint, so unlike the HLS offload path nothing verified the selected proxy could execute the transformations the plan froze. On a mismatched build the failure is real and ugly: a missing `aac` encoder 500s mid-stream, and a missing `dovi_rpu` filter is refused outright by the remux (`remux.go:265`).

Fixed by giving the proxy the same endpoint, in the same shape and at the same path as a transcode node, and validating the frozen recipe against the selected proxy before committing to it. Two deliberate choices:

- A proxy that does not answer the probe is treated as **incapable**, not assumed good. An older proxy predating the endpoint is precisely the mismatched build this check exists to catch, so failing open would defeat it.
- Direct play skips the probe entirely — it copies bytes and needs no recipe, so it should not pay a fetch or be rejected by one. There is a test asserting zero probes on that path.

**Write deadlines.** Confirmed: `meteredResponseWriter` implemented neither `Unwrap` nor `SetWriteDeadline`, and the standalone proxy runs `WriteTimeout: 0`, so there was no server-level guard behind the degraded wrapper. Added `Unwrap`. Worth noting this affected *every* proxy stream, not only the new routes.

**Session liveness.** Also confirmed, with one clarification: this exposure is pre-existing rather than introduced here — `buildProxyManifestURL` on unmodified `main` already returns an absolute proxy URL for HLS that never touches `HandleStream`/`BeginTransport`. This change widens it to direct play and remux, so it is fair to fix here. Sessions are now marked as remotely transported.

I did **not** implement it as unconditional immunity. `SessionManager` has no absolute session lifetime cap, so a session that can never be reaped would leak forever whenever a client disappears without calling stop — trading a reaped-too-early bug for a leaked-forever one. Instead the mark widens the idle windows to a 5-minute floor: long enough to outlast a heartbeat gap on a healthy stream, short enough that an abandoned session still goes away. The mark is always set on commit (not only when true), so a re-plan that moves a session back onto the API clears a stale one.

Both CodeRabbit nitpicks applied as well — exported transformation constants in the tests, and coverage for the effective-recipe bitrate branch, which matters because a downmixing remux egresses less than its source.

New tests: proxy rejected when it lacks the recipe (asserting the reservation is released), direct play skips the probe, effective-vs-source bitrate selection, proxy capability endpoint + its auth, and four in `internal/playback` covering the widened grace — including one proving an abandoned proxy session is still reaped, and one proving clearing the mark restores normal reaping.

Verification: `go build ./...`, affected packages green, `golangci-lint --new-from-merge-base=origin/main ./internal/...` → 0 issues, `gofmt` clean.

Unrelated, but flagging since it will show up in CI: `TestHandleReplanPlaybackV3BitmapSubtitleFallsBackFromHDRToSDRVersion` fails on a pristine checkout of `ee9356aab` (#617) with none of this branch's code. Not from this PR, but it is broken on `main` right now.

Quick104:
Round two, in `d86e7e59`. Also fixed the CI failure, which turned out not to be from this branch.

**CI: stale playback fixtures.** `internal/playback/testdata/protocol_v3` was missing `output_change_v1`, added by #613/#617 without regenerating. It is stale on `origin/main` too, so it fails every branch — but it blocks this PR, so `make playback-fixtures` is included here. The diff is one line per file, nothing else.

**Capable sibling proxies** (Cursor + Codex, same finding). Valid, and the more interesting of the two framings: on the api/proxy split this PR targets, with `local_transcode_fallback` disabled, a single mismatched round-robin pick would refuse playback outright while a capable proxy sat idle.

Fixed by narrowing *before* selection instead of rejecting after, which is what the HLS path already does for transcode nodes. `PlanSessionWith` now applies its eligibility predicate to the proxy on proxy-only plans — the proxy is the executor there — and the planner grows `ProxyNodeURLs()` to match `TranscodeNodeURLs()`. A nice side effect: no reservation is ever taken against an incapable proxy, so there is nothing to release on that path (the earlier test asserting a release was encoding the old design and has been updated).

**Stale remote-transport mark** (P2). Valid. Only the identity commit called `SetRemoteTransport`, so a replan from proxy to integrated HLS left the mark set and the widened grace held the session's stream/transcode slots for five minutes after a local disconnect. Every committed route now records locality — including the remote HLS route, which also hands the client an absolute proxy URL that never reaches this server.

**CORS validators** (P2). Valid and a genuinely good catch: the proxy allowed `If-Range`/`Range` as *request* headers but exposed no *response* headers, so cross-origin JS could never read the `ETag` it was supposed to echo. `direct_stream_resume_v1` silently degraded to a full restart whenever the proxy was on a different origin — the normal deployment. Added `ETag`, `Accept-Ranges`, `Content-Range`, `Content-Length`, `Content-Encoding`, `Last-Modified` to `ExposedHeaders`.

Two I have **not** changed, with reasoning:

**"Preserve reconstruction when returning proxy stream URLs."** The premise that the API-local URL previously survived a restart does not hold for the calls named. `HandleUpdateProgress` does a bare `sessionMgr.GetSession` with no reconstruct path — on `origin/main` as well — so progress/stop return session-not-found after an API restart regardless of which URL the client holds. The `?st=` reconstruct only ever applied to `HandleStream` itself, i.e. the media request, which on a proxy-served session is served by the proxy from the same token. So this is a real gap in v3 session durability, but a pre-existing one that predates this PR and wants its own issue rather than a rushed fix here.

**"Keep lazy proxy reservations until the stream starts."** Real, and also pre-existing: the identical window exists today for HLS, where `PlanSession` reserves at plan time and the proxy reports nothing until the client opens the manifest. Fixing it properly means admission at GET time or a proxy-confirmed reservation — a change to the reservation protocol shared by every route, which does not belong in a bug fix for #619. Worth its own issue.

The three P1s re-posted from commit `2716856` (write deadlines, session liveness, proxy capability validation) were all fixed in `811c85d5`; those comments were against the pre-fix commit.

New tests: capable-sibling selection end to end, mark cleared on a proxy→local transition, planner-level proxy filtering (including that an unsatisfiable plan leaves no reservation), and `ProxyNodeURLs`. Both new behavioral tests were confirmed to fail against pre-fix behavior.

Verification: `go build ./...`, `make verify-playback-fixtures` clean, affected packages green, `golangci-lint --new-from-merge-base=origin/main ./internal/...` → 0 issues, `gofmt` clean.

Still failing on `main` independently of this branch, for whoever picks it up: `TestHandleReplanPlaybackV3BitmapSubtitleFallsBackFromHDRToSDRVersion` and the two `TestBeginWebOperation*` tests in `jellycompat`. CI's Go job does not run the Go suite, so they are not gating anything today.

Quick104:
Go is fixed in `0dbca895`. Two corrections to what I said earlier, both mine.

**I was wrong that CI does not run the Go suite.** It does — `make test-go` is `go test ./...`. I inferred otherwise from the previous run, where the fixture step failed *before* the Test step executed, so the suite never appeared in the log. Anything I said about those failures "not gating" was wrong.

**I committed debug instrumentation by mistake.** While diagnosing the pre-existing `TestHandleReplanPlaybackV3BitmapSubtitleFallsBackFromHDRToSDRVersion` failure I added a `t.Fatalf` printing the terminal reason. I reverted it in the wrong checkout, so the revert silently did nothing and the instrumentation shipped in `d86e7e59`. Removed; the assertion is back to its original form.

**The underlying failure turned out to be a real bug, not a stale test.** #617 changed which terminal a burn-in-forced adaptation reports: when the subtitle burn requirement is the sole trigger, an HDR source that cannot be re-encoded now returns `subtitle_conversion_unsupported` rather than `hdr_transcode_unsupported`, so the refusal names what the viewer can act on. Good change.

But `terminalAllowsAlternateFileV3` gates the alternate-version retry on the old reason strings and was not updated. That silently retired the fallback for precisely the case its own comment describes:

> a bitmap subtitle can require video burn-in that an HDR source cannot support while an SDR alternate can

So selecting a PGS/VOBSUB track on an HDR source with an SDR alternate available refused playback outright instead of switching to the version that can serve it. The test was not asserting stale behavior — it was correctly reporting a regression that had been failing on `main` since #617 merged.

Fixed by adding the new reason to the gate, plus a direct unit test on `terminalAllowsAlternateFileV3` so a future rename of a refusal reason fails on the gate itself rather than only through an end-to-end replan test.

Full `go test ./...` now passes for `internal/api/handlers`. Five failures remain in my local run only — `jellycompat` `TestBeginWebOperation*`, `playback` `TestFFmpegSupportsNVENC*`, and `transcodenode` `TestHandleDownloadPrepareTracking*`. All three packages passed in CI's own run of this branch, so they are environment-specific here (NVENC probing, process-lock semantics) and not something this branch introduces. Correspondingly, my earlier note about `TestBeginWebOperation*` being "broken on main" was also wrong — they are green in CI.

`golangci-lint --new-from-merge-base=origin/main ./internal/...` → 0 issues, `gofmt` clean, fixtures current.

PR #621, opened 2026-08-13
perf(literaryworks): match candidates via UNION branches instead of cross-table OR

RXWatcher:
## Problem

`listMatchCandidateIDs` OR-s its three match signals into a single `WHERE`:

```go
matchWhere = " AND (" + strings.Join(matchFilters, " OR ") + ")"
```

Those predicates live on three different tables (`media_items.title`, `media_item_provider_ids`, `{ebook,audiobook}_series`). PostgreSQL can bitmap-OR predicates on one relation, but not across relations — so it discards the partial index `idx_media_items_books_title_lower` and walks every opposite-format book, applying the OR as a post-filter.

Plan on a real 434k-book library (`EXPLAIN (ANALYZE, BUFFERS)`):

```
Index Scan using idx_media_items_type_year on media_items mi (actual rows=1.00 loops=1)
  Filter: (... AND ((lower(title) = 'golden margins') OR (ANY (content_id = (hashed SubPlan 2).col1))))
  Rows Removed by Filter: 434175
  Buffers: shared hit=346929
Execution Time: 578.718 ms
```

This matters because `AutoLinkContent` calls it once per **unlinked** book on every rescan. Its cheap guard only skips books already linked to a work. On my library 412,434 of 434,176 books (95%) are unlinked — most books exist in only one format, so they never link, and they are re-scanned in full every pass. That is ~65 CPU-hours per pass, indefinitely.

## Approach

Each match signal becomes its own `UNION` branch inside `mi.content_id IN (...)`, so each rides its own index. The title branch repeats `type IN ('ebook','audiobook')` so it matches the *partial* index.

No schema change and no new index — every index this needs already exists.

Semantics are unchanged: same result set, same `ORDER BY title, content_id`, same limit, same `ignored`-decision anti-join. `EXISTS (... x.content_id = mi.content_id ...)` becomes a `content_id` projection inside `IN (...)`, which is equivalent; `UNION` dedupes, and `IN` cannot multiply rows.

## Results

Same query, same data, before vs after:

| | before | after |
|---|---:|---:|
| Execution time | 578.718 ms | **0.135 ms** |
| Shared buffers hit | 346,932 | **11** |
| Rows removed by filter | 434,175 | 0 |
| Indexes used | `idx_media_items_type_year` | `idx_media_items_books_title_lower`, `audiobook_series_name_lower`, pkey |

A full pass over 412,434 unlinked books goes from ~65.5 CPU-hours to ~56 seconds.

## How it was tested

New `internal/literaryworks/match_candidates_test.go`:

- `TestMatchCandidateQueryUsesTitleIndex` — seeds 30k books, `ANALYZE`s, and asserts the plan names `idx_media_items_books_title_lower`. **Written first and watched fail** against the old query, where the plan showed `Seq Scan on media_items` with the OR as a post-filter.
- Four semantic tests written *before* the change and passing against the old code, pinning behavior across the rewrite: opposite-format case-insensitive title match, series-position match despite differing titles, `ignored`-decision suppression, and ordering/limit with no duplicate rows from multiple series rows.

Verification run locally against a `pgvector/pgvector:pg17` container with this repo's migrations applied:

- `go test ./internal/literaryworks` — 16/16 pass
- `gofmt -l` — clean
- `go vet` — clean
- `golangci-lint run ./internal/literaryworks/...` — **3 issues, identical to the pre-change baseline on the same commit.** No regression.

Two honest caveats:

1. **I did not run the app end-to-end.** I verified the query plan and timings directly against a real 434k-book production database, and ran the package tests, but I did not spin up the UI and click through book matching.
2. **`go build ./...` fails on `web/embed.go: pattern all:dist: no matching files found`** — this reproduces identically on unmodified `main` without a frontend build, so it is pre-existing and unrelated.

Also note: this repo's CI does not set `SILO_TEST_DATABASE_URL`, so these tests (like the existing DB tests) **skip in CI**. They only run against a local database.

## Watch out for

- `source.ExternalIDs` is a Go map, so provider branches are emitted in **non-deterministic order**. That varies the SQL text between calls, which fragments `pg_stat_statements` and the prepared-statement cache. This is pre-existing, and I deliberately left it out to keep one thing per MR — happy to send a follow-up that sorts the providers.
- This does not address the missing negative cache. Unmatched books are still re-scanned every pass; they are just ~4,300× cheaper now. A "no match found" marker would remove the work entirely, but that is a behavior change worth its own discussion.

### AI Disclosure
- **Tool(s):** Claude Code
- **Model(s):** `claude-opus-5[1m]` (Opus 5, 1M context)
- **Involvement:** fully AI-generated, human-directed and human-reviewed
- **Adversarial review:** The self-review found one real defect: the new test file introduced 9 `"ebook"` string literals, which pushed `goconst` over threshold and added a 4th lint issue versus baseline — a lint regression. Resolved by switching the tests to the existing `FormatEbook`/`FormatAudiobook` constants, matching the convention already used in `matcher_test.go`, `service_test.go`, and `repository_test.go`; lint then matched baseline exactly at 3 pre-existing issues. The review also checked the `EXISTS`→`IN` rewrite for NULL semantics (`content_id` is `NOT NULL` in all three source tables, so `IN` and `EXISTS` behave identically here) and for row multiplication (`UNION` dedupes; the ordering/limit test seeds a candidate with a series row specifically to catch duplicates). Two earlier hypotheses were investigated and **disproved** before landing on this one: table bloat/autovacuum starvation (`episodes` is correctly sized; autovacuum has legitimately not hit its threshold) and a missing index (all required indexes already exist). Both were discarded on evidence rather than carried into the diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved literary-work matching across formats using titles, provider IDs, and series information.
  * Matching now handles case differences and series positions more reliably.
  * Excludes ignored decisions, duplicate results, and items from the same source.
  * Preserves deterministic title ordering and result limits.

* **Tests**
  * Added comprehensive coverage for matching behavior, ordering, limits, duplicate prevention, and database query performance.

### DIFF
--- a/internal/literaryworks/match_candidates_test.go
+++ b/internal/literaryworks/match_candidates_test.go
@@ -1,0 +1,211 @@
+package literaryworks
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// The candidate lookup fans out across three independent match signals (title,
+// provider IDs, series position). These tests pin the result semantics so the
+// query shape can change underneath them without altering what comes back.
+
+func seedCandidateFixture(t *testing.T, pool *pgxpool.Pool) (folderID int, cleanup func()) {
+	t.Helper()
+	ctx := context.Background()
+	folderID = seedLiteraryFolder(t, pool)
+	cleanup = func() {
+		_, _ = pool.Exec(ctx, `DELETE FROM media_items WHERE content_id LIKE 'cand-%'`)
+		_, _ = pool.Exec(ctx, `DELETE FROM media_folders WHERE id = $1`, folderID)
+	}
+	t.Cleanup(cleanup)
+	return folderID, cleanup
+}
+
+func seedSeries(t *testing.T, pool *pgxpool.Pool, table, contentID, seriesName string, index float64) {
+	t.Helper()
+	if _, err := pool.Exec(context.Background(), fmt.Sprintf(`
+		INSERT INTO %s (content_id, series_name, series_index)
+		VALUES ($1, $2, $3)
+	`, table), contentID, seriesName, index); err != nil {
+		t.Fatalf("seed %s for %s: %v", table, contentID, err)
+	}
+}
+
+func TestListMatchCandidateIDsMatchesOppositeFormatByTitle(t *testing.T) {
+	pool := newLiteraryWorksTestPool(t)
+	folderID, _ := seedCandidateFixture(t, pool)
+	repo := NewRepository(pool)
+
+	seedLiteraryMediaItem(t, pool, "cand-src", FormatEbook, "Golden Margins", folderID)
+	seedLiteraryMediaItem(t, pool, "cand-hit", FormatAudiobook, "golden margins", folderID)
+	seedLiteraryMediaItem(t, pool, "cand-miss", FormatAudiobook, "Different Book", folderID)
+
+	ids, err := repo.listMatchCandidateIDs(context.Background(), MatchItem{
+		ContentID: "cand-src",
+		Type:      FormatEbook,
+		Title:     "Golden Margins",
+	}, 20)
+	if err != nil {
+		t.Fatalf("listMatchCandidateIDs: %v", err)
+	}
+
+	if len(ids) != 1 || ids[0] != "cand-hit" {
+		t.Fatalf("ids = %v, want [cand-hit] (case-insensitive title match, opposite format only)", ids)
+	}
+}
+
+func TestListMatchCandidateIDsMatchesBySeriesPosition(t *testing.T) {
+	pool := newLiteraryWorksTestPool(t)
+	folderID, _ := seedCandidateFixture(t, pool)
+	repo := NewRepository(pool)
+
+	seedLiteraryMediaItem(t, pool, "cand-src", FormatEbook, "Book One Title", folderID)
+	seedLiteraryMediaItem(t, pool, "cand-hit", FormatAudiobook, "Totally Other Title", folderID)
+	seedSeries(t, pool, "audiobook_series", "cand-hit", "Vaz", 1)
+
+	index := 1.0
+	ids, err := repo.listMatchCandidateIDs(context.Background(), MatchItem{
+		ContentID:   "cand-src",
+		Type:        FormatEbook,
+		Title:       "Book One Title",
+		SeriesName:  "vaz",
+		SeriesIndex: &index,
+	}, 20)
+	if err != nil {
+		t.Fatalf("listMatchCandidateIDs: %v", err)
+	}
+
+	if len(ids) != 1 || ids[0] != "cand-hit" {
+		t.Fatalf("ids = %v, want [cand-hit] matched on series despite differing titles", ids)
+	}
+}
+
+func TestListMatchCandidateIDsExcludesIgnoredDecisions(t *testing.T) {
+	pool := newLiteraryWorksTestPool(t)
+	folderID, _ := seedCandidateFixture(t, pool)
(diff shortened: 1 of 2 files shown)


RXWatcher:
The `verify-playback-fixtures` failure on this PR is **pre-existing on `main`**, not caused by these changes.

This PR touches only `internal/literaryworks/`. The failing job is in `internal/playback/`.

I reproduced it on a clean detached checkout of `ee9356aa` (this PR's base, none of these changes applied):

```
+            "output_change_v1",
::error::internal/playback/testdata/protocol_v3 is stale; run make playback-fixtures
```

Same three fixtures, same single added feature string. `output_change_v1` was added to the playback feature list without regenerating the checked-in fixtures, so the job fails for every open PR right now.

Opened #622 to regenerate them — 5 insertions, 0 deletions, pure generator output. Kept separate from this PR rather than folded in, per the one-thing-per-MR guidance in CONTRIBUTING.

Happy to rebase this branch once #622 (or an equivalent fix) lands so CI here goes green.

---

Separately, on the CodeRabbit review points — both look correct to me:

- The ordering assertion in `TestListMatchCandidateIDsOrdersByTitleAndHonoursLimit` checks length and source-exclusion but not the actual order, despite the test being named for it.
- There's no coverage of the provider-ID UNION branch via `source.ExternalIDs` — I rewrote three branches and only tested two.

I'll push both fixes shortly.

One thing worth flagging for reviewers: this repo's CI does not set `SILO_TEST_DATABASE_URL`, so the new tests here (like the existing DB-backed tests) **skip in CI**. Green CI means they compiled, not that they ran. I verified them locally against `pgvector/pgvector:pg17` with this repo's migrations applied — 16/16 pass, and the index-usage test was confirmed to fail against the old query before the fix.

RXWatcher:
Both review points addressed in `700efb8a` — test-only change, `repository.go` is byte-identical to the previous commit.

**Ordering assertion.** `TestListMatchCandidateIDsOrdersByTitleAndHonoursLimit` now asserts the exact slice rather than just the length and source-exclusion. Fair catch — the test was named for ordering and wasn't checking it.

**Provider-ID branch.** Added `TestListMatchCandidateIDsMatchesBySharedProviderID`: titles differ, neither side has series data, so a shared non-ASIN provider ID is the only signal that can match. Also added `TestListMatchCandidateIDsIgnoresASINProviderID`, since `asin` is explicitly skipped when building the branches — ASIN identifies an edition rather than a work, and that exclusion had no coverage either.

Both new assertions were verified to actually catch regressions rather than just passing:

| Deliberate break | Result |
|---|---|
| Provider branch disabled | `ids = []`, want `[cand-hit]` → FAIL |
| Candidate query sort flipped to `DESC` | `ids = [cand-c cand-b]`, want `[cand-a cand-b]` → FAIL |

Worth noting the first attempt at that verification was itself wrong — I patched the `ORDER BY` in `ListMatchCandidates`' hydration query instead of the one in `buildMatchCandidateQuery`, and the test passed. Retargeting to the right statement produced the expected failure. Mentioning it because "I checked the test has teeth" is only worth anything if the check hit the right code.

Local verification against `pgvector/pgvector:pg17` with this repo's migrations: full package 18/18 pass, `gofmt` clean, `golangci-lint` at 3 issues — identical to the pre-change baseline on this commit, so no new lint.

Reminder for reviewers: this repo's CI does not set `SILO_TEST_DATABASE_URL`, so none of these DB-backed tests actually execute in CI. They were run locally.

The remaining red check is `verify-playback-fixtures`, which is pre-existing on `main` and unrelated to this PR — see #622.

Quick104:
Closing this duplicate in favor of #1288, which addressed the same OR-to-UNION candidate query linked to issue #1262. #1288 has since merged, so the optimization is now in main.

Issue #630, opened 2026-08-13
Surface exact client app version and build on the Activity page


Quick104:
## Problem

The admin Activity page cannot tell you which build of a client app a session is streaming from.

Two separate causes:

1. **The version we already had was being destroyed.** Android has been sending `X-Silo-Client-Version` on every request, and the server stored it intact in `playback_sessions_sync.client_version`. But the UI renders `client_label`, composed by `playbackClientDisplayName`, which routes the version through `shortPlaybackClientVersion` — that strips non-`[0-9.]` runes, truncates to two components, and drops a trailing `.0`. A client reporting `1.0.0` rendered as **`Silo Android TV 1`**.
2. **The Apple clients sent nothing.** No client name, no app version — only device id/name/platform/family. Every iOS, tvOS, and macOS session fell back to user-agent sniffing.

Neither platform had any notion of a build number on the session path, so two builds of the same marketing version were indistinguishable.

## Scope

- Report app version **and** build number from the Android and Apple clients.
- Show the exact build on the Activity page and include it in diagnostic logs.
- Distinguish official builds from sideload/dev builds.

## Approach

Two additive, **opaque** wire fields alongside the existing client headers: `X-Silo-Client-Build` (≤64) and `X-Silo-Client-Channel` (≤32), with `client_playback_context.app_build` / `app_channel` as the v3 fallback.

Opaque is the load-bearing decision. Apple uses a per-platform TestFlight sequence; Android uses a per-marketing-version counter. Keeping the server from parsing, comparing, or enum-validating either value lets both coexist without inventing a shared scheme. Displaying build alongside version and platform — `Silo Android TV 1.0.0 (build 5)` — is unambiguous and matches what Play Console and TestFlight show.

Consequence worth recording: the server cannot sort or compare builds. Any future minimum-client-version gating must key on `client_version`, which is semver.

## Out of scope

`SiloMac` has no release lane at all (`release.yml` has only `setup`/`build-numbers`/`ios`/`tvos`; the Fastfile has no `SiloMac` lane), so Mac sessions will report `1.0.0 (build 1)` until one exists. Deferred deliberately: `SiloMac` has no entitlements file, and the Mac App Store path requires App Sandbox around bundled FFmpeg frameworks and arbitrary local-server access. Developer ID + notarized DMG is the lighter path if revisited. Tracked separately.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: fully AI-generated
[...]

PR #631, opened 2026-08-13
feat(activity): report exact client app version, build, and channel

Quick104:
Part of #630

## Problem

The Activity page could not name the build a session was streaming from — and the more interesting half of that was self-inflicted.

Android has been sending `X-Silo-Client-Version` on every request, and the server stored it intact in `playback_sessions_sync.client_version`. But the UI renders `client_label`, composed by `playbackClientDisplayName`, which routed the version through `shortPlaybackClientVersion`: that strips every non-`[0-9.]` rune, truncates to two components, and drops a trailing `.0`. So a client reporting `1.0.0` rendered as **`Silo Android TV 1`**. The exact version was on the wire and in the database, just unreachable by the UI.

Separately, the Apple clients sent no client name or version at all (fixed in Silo-Server/silo-apple#153), and no platform had any build-number concept on the session path.

## Approach

Two additive, **opaque** wire fields alongside the existing client headers — `X-Silo-Client-Build` (≤64) and `X-Silo-Client-Channel` (≤32) — with `client_playback_context.app_build` / `app_channel` as the v3 fallback. That fallback is also where the previously-received-and-discarded `app_version` finally gets used.

**Why opaque, and why a separate header.** Not folded into `X-Silo-Client-Version` because `shortPlaybackClientVersion` would eat any non-numeric suffix, and because build numbers are not semver components: Apple uses per-platform TestFlight sequences, Android a per-marketing-version counter. The server never parses, compares, or enum-validates build or channel, which is what lets both schemes coexist without inventing a shared one. The cost is real and worth stating: **the server cannot sort or compare builds**, so any future minimum-client-version gating must key on `client_version`, which is semver. Naming matches the diagnostics contract's existing `app_version`/`app_build` pair so the two admin surfaces read the same.

**The label split.** Only the *named-client* branch of `playbackClientDisplayName` stops truncating. The user-agent branch keeps `shortPlaybackClientVersion`, so browser labels stay `Chrome 120` instead of becoming a full UA version string — there is a test pinning this. `client_label` therefore becomes `Silo Android TV 1.0.0` (compact but exact), and the build lives in the new `client_label_full`.

**Where it shows.** The compact row is unchanged in width — `getSessionClientLabel` is shared with `AdminDashboard`, `AdminStats`, and `HouseholdStreamsPanel`, none of which are touched. The exact string lands in the row tooltip and a new **Client** card in the expanded panel.

**Logging.** `client_name`/`version`/`build`/`channel` on both `playback plan decided` lines and on session expiry. `opslog` stores an open `attrs` JSONB, so this surfaces at `/admin/logs` with no migration. `activity_log` is deliberately untouched — highest-volume table, and the value is constant per device.

## v1 API compliance

Additive only. `client_build`, `client_channel`, `client_label_full` are all new and `omitempty`; no existing field changes type or meaning. `GET /admin/sessions/capabilities` advertises `client_build`/`client_channel` for feature detection, matching the existing `effective_play_method` / `is_jellyfin_client` pattern.

The one behavioral change is the truncation fix, which is a bug fix rather than a contract change — `Silo Android TV 1` was never the intent.

## Backward compatibility

- **Older clients** send no build; `client_label_full` degrades to `Silo Android TV 0.3.11`, and the UI renders the parenthetical only when present — no "(build unknown)" noise.
- **Jellyfin compat clients** keep an empty build. The `MediaBrowser` auth header vocabulary has `Client`/`Device`/`DeviceId`/`Version` and no build concept; synthesizing one from a user agent would be a guess, so `internal/jellycompat/auth.go` is untouched.
- **Older servers, newer clients** — the extra header is ignored, and the new JSON fields are safe: the v3 start decoder does not use `DisallowUnknownFields` (verified; the strict decoders are all elsewhere).
- **Web sessions** keep `Chrome 120` — the UA branch is unchanged.

## Verification

Docker was unavailable, so the migration was applied for real against a throwaway database on a scratch Postgres, and each hand-edited statement was then `PREPARE`d against the resulting live schema — the class of column/placeholder mismatch that fails at runtime rather than compile time:

```
playback_route_events   client_build    text  YES
playback_route_events   client_channel  text  YES
playback_sessions_sync  client_build    text  YES
playback_sessions_sync  client_channel  text  YES
goose_db_version: 20260813160942 | true
```

- Reconciler upsert (29 columns / 29 value slots / 28 placeholders + `NOW()`) — `PREPARE` OK
- Route-events insert — `PREPARE` OK
- Loader `SELECT` — `PREPARE` OK, **48 columns against 48 `rows.Scan` args**

Then:

```
gofmt -l internal cmd        # clean
go build ./...               # OK
go test ./internal/api/handlers/ ./internal/worker/ ./internal/playback/planstore/
  ok  internal/api/handlers        87.644s
  ok  internal/worker               1.003s
  ok  internal/playback/planstore   1.192s
```

Label tests, including the regression guard:

```
--- PASS: TestPlaybackClientDisplayNameKeepsNamedClientVersionExact
--- PASS: TestPlaybackClientFullDisplayName
--- PASS: TestPlaybackClientDisplayNameAndroidDevices
      (incl. chrome_remains_browser_label)
--- PASS: TestSessionsCapabilitiesAdvertisesActivityFields
```

Web: `tsc -b` clean, `prettier --check` clean, `vitest src/pages/adminActivityPresentation.test.ts` 13/13 pass. `golangci-lint run --new-from-rev=HEAD ./...` reports **0 issues** (a whole-tree run surfaces 167 pre-existing findings, none from this branch — CI runs `--new-from-merge-base`). `make verify-local-paths` passes.

Pre-existing failures, reproduced at pristine `HEAD` in a throwaway worktree and not caused by this branch: `internal/jellycompat` `TestBeginWebOperation{RecoversDeadProcessLock,RejectsLiveProcessLock}` (deterministic at HEAD), `internal/playback` gpudetect NVENC tests (flaky at HEAD, pass on rerun), and 4 localStorage-dependent web tests. Nothing was added to `WEBTEST_KNOWN_FAILURES`.

## Risks

- The migration is two nullable `ADD COLUMN IF NOT EXISTS` pairs — no rewrite, no default backfill, safe on a live table.
- Build/channel are set once at session creation; no stream-state update path clears them. `SessionStreamState` and `RecipeCard` were deliberately left alone (`RecipeCard`'s client fields are populated only by jellycompat, which has no build concept).
- No v2 session-start log line exists to extend — the legacy start endpoint returns 426 and routes into v3 — so the attrs went to the v3 plan lines and session expiry instead.

## Follow-ups

- The web app reports the literal `"web"` as its app version; teaching it a real version from the git tag is an easy separate change.
- `SiloMac` has no release lane, so Mac sessions will report `1.0.0 (build 1)` — see #630.

## Client-side companions

- Silo-Server/silo-android#226
- Silo-Server/silo-apple#153

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5
- Adversarial review: review of the diff caught two real defects before this landed. (1) In the Apple companion PR, the sideload build-number counter had been rewritten as a count of existing releases, which regresses if any release is deleted and lets two different IPAs ship as the same version+build — reverted to max-based and fixture-tested. (2) Unstamped Android builds would have reported a placeholder build `0`, which the opaque-string contract means the server would render verbatim as "(build 0)"; fixed on the client so an unstamped build reports the build as absent. Review also specifically re-verified that `shortPlaybackClientVersion` survives on the user-agent path only, since collapsing both branches would have silently regressed browser labels.

## Review follow-up (b43b7ef06, 143a7e622, e61c705c4)

An independent review of this branch raised 14 findings; all are addressed in a
second commit. Full mapping in [this comment](https://github.com/Silo-Server/silo-server/pull/631#issuecomment-5285975312).
Four of them revise claims made above, so they belong in the description:

- **The body fallback now applies only to a client that sent `X-Silo-Client`.**
  As originally written it took `client_playback_context.app_version` whenever
  the header was absent — and the web player sends the literal `"web"` there
  while sending no client-name header, so every browser session would have
  stamped `client_version="web"`. That is the field this description reserves as
  the semver key for future minimum-version gating. `client_playback_context`
  carries no app name, so nothing in the body can identify a nameless client
  anyway.
- **Over-long `app_build`/`app_channel` are clamped, not rejected.**
  `validateCapabilitiesV3` was failing the whole start request with 400 while the
  header route silently clamped the same value — an opaque diagnostic label could
  refuse playback. Both routes now use `normalizeClientMetadataValue`, which is
  what the settings-api doc already promised.
- **The full label keeps build and channel for user-agent-labelled clients.** It
  previously dropped both whenever the client reported no name, so the new Client
  card could never show a build for a browser or a `Client`-less jellycompat
  session. `client_label` (the compact one) is unchanged.
- **`replan-request.schema.json` gained the same two properties.**
  `ReplanRequestV3` reuses `ClientPlaybackContextV3` and validates the same
  bounds, so shipping them in the start schema alone left the replan contract
  describing a type the server no longer has. A new contract test asserts every
  `$def` the two request schemas share is identical.

Also in that commit: rune-safe clamping in `normalizeClientMetadataValue` (a
mid-rune byte cut yields invalid UTF-8, which fails the whole per-node session
sync transaction), route events completing their identity from the session,
`ClientInfo.LogAttrs()` as the single definition of the four log keys — omitting
fields the client did not report rather than writing empty keys into opslog —
`client_label_full` omitted when it would repeat `client_label`, a test for the
header/body precedence rule, and the Activity search matching the exact label so
a build number is findable.

A later bot-review round added two more fixes ([summary](https://github.com/Silo-Server/silo-server/pull/631#issuecomment-5286119623)):
the client identity is now clamped at the request boundary rather than only where
the session stamps it — the decision logs and `playback_route_events` are written
from the resolved value, so an oversized header reached both — and the clamp
counts runes, matching the `maxLength` the schemas publish. Separately, a JSON NUL
escape in the start body survived the UTF-8 repair (NUL is valid UTF-8) and would
have failed the whole per-node session-sync transaction on a `text` column;
control characters are now stripped outright.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit ce81a3b3b8ed82e477ad8095267d231a5d573da4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Playback sessions now capture and display client build and distribution channel details.
  * Added full client identity labels, including exact versions, builds, and channels.
  * Playback requests accept optional app build and channel information.
  * Admin activity details now show expanded client identity and user-agent information.
  * Client metadata is included in playback events, session synchronization, and diagnostics.
  * Metadata values are safely limited to supported lengths.

* **Documentation**
  * Documented supported client identity headers, fallback behavior, and value limits.

### DIFF
--- a/docs/design/schemas/playback-v3/v3/start-request.schema.json
+++ b/docs/design/schemas/playback-v3/v3/start-request.schema.json
@@ -531,6 +531,14 @@
           "type": "string",
           "maxLength": 64
         },
+        "app_build": {
+          "type": "string",
+          "maxLength": 64
+        },
+        "app_channel": {
+          "type": "string",
+          "maxLength": 32
+        },
         "device": {
           "type": "object",
           "description": "Manufacturer and model stay first-class because the device-quirk registry matches on them. Everything platform-specific travels in platform_details as opaque bounded strings.",
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -1078,15 +1078,18 @@ func playbackClientInfoFromRequest(r *http.Request) playback.ClientInfo {
 	if r == nil {
 		return playback.ClientInfo{}
 	}
-	// Build and channel are opaque: they are trimmed and length-clamped when the
-	// session stamps them, never parsed or validated against an enum here.
+	// Clamped here, at the boundary, rather than only where the session stamps
+	// them: the decision logs and playback_route_events are written from this
+	// value directly, so a client sending a header-sized build would otherwise
+	// reach both despite the published bound. Values stay opaque — trimmed and
+	// length-clamped, never parsed or validated against an enum.
 	return playback.ClientInfo{
-		Name:      strings.TrimSpace(r.Header.Get("X-Silo-Client")),
-		Version:   strings.TrimSpace(r.Header.Get("X-Silo-Client-Version")),
-		Build:     strings.TrimSpace(r.Header.Get("X-Silo-Client-Build")),
-		Channel:   strings.TrimSpace(r.Header.Get("X-Silo-Client-Channel")),
+		Name:      r.Header.Get("X-Silo-Client"),
+		Version:   r.Header.Get("X-Silo-Client-Version"),
+		Build:     r.Header.Get("X-Silo-Client-Build"),
+		Channel:   r.Header.Get("X-Silo-Client-Channel"),
 		UserAgent: r.UserAgent(),
-	}
+	}.Normalized()
 }
 
 // HandleUpdateProgress handles POST /playback/{session_id}/progress.
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -952,7 +952,9 @@ func (h *PlaybackHandler) handleExpiredSession(session *playback.Session) {
 	}
 	sessionCopy := *session
 	go func() {
-		slog.Info("expired inactive playback session", "session", sessionCopy.ID, "playback_session_id", sessionCopy.ID)
+		slog.Info("expired inactive playback session", "session", sessionCopy.ID, "playback_session_id", sessionCopy.ID,
+			"client_name", sessionCopy.ClientName, "client_version", sessionCopy.ClientVersion,
+			"client_build", sessionCopy.ClientBuild, "client_channel", sessionCopy.ClientChannel)
 		// Expiry is a liveness reap, not a user stop — keep the recipe card so a
 		// resume reconstructs under the same id (the card's own TTL reaps it if
 		// the session is truly abandoned).
@@ -1076,9 +1078,13 @@ func playbackClientInfoFromRequest(r *http.Request) playback.ClientInfo {
 	if r == nil {
 		return playback.ClientInfo{}
 	}
+	// Build and channel are opaque: they are trimmed and length-clamped when the
+	// session stamps them, never parsed or validated against an enum here.
 	return playback.ClientInfo{
 		Name:      strings.TrimSpace(r.Header.Get("X-Silo-Client")),
 		Version:   strings.TrimSpace(r.Header.Get("X-Silo-Client-Version")),
+		Build:     strings.TrimSpace(r.Header.Get("X-Silo-Client-Build")),
+		Channel:   strings.TrimSpace(r.Header.Get("X-Silo-Client-Channel")),
 		UserAgent: r.UserAgent(),
 	}
 }
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -43,6 +43,11 @@ import (
 type SessionManagerInterface interface {
 	StartSession(userID int, profileID string, fileID int, method playback.PlayMethod, transcodeAudio bool) (*playback.Session, error)
 	StartSessionWithFiles(userID int, profileID string, effectiveFileID int, requestedFileID int, method playback.PlayMethod, transcodeAudio bool) (*playback.Session, error)
+	// StartSessionWithFilesContext is required rather than probed for at run
+	// time: the context is how the reporting client's identity reaches the new
+	// session, so an implementation without it would start sessions that
+	// silently carry no client name, version, build or channel.
+	StartSessionWithFilesContext(ctx context.Context, userID int, profileID string, effectiveFileID int, requestedFileID int, method playback.PlayMethod, transcodeAudio bool) (*playback.Session, error)
 	UpdateProgress(sessionID string, position float64, isPaused bool) error
 	UpdateAudioTrack(sessionID string, audioTrackIndex int, method playback.PlayMethod) error
 	UpdateStreamState(sessionID string, state playback.SessionStreamState) error
@@ -61,10 +66,6 @@ type SessionManagerInterface interface {
 	SetProgressPersistenceDisabled(sessionID string, disabled bool) error
 	StopSession(sessionID string) error
 	GetSession(sessionID string) (*playback.Session, error)
-}
-
-type sessionStarterWithFilesContext interface {
-	StartSessionWithFilesContext(ctx context.Context, userID int, profileID string, effectiveFileID int, requestedFileID int, method playback.PlayMethod, transcodeAudio bool) (*playback.Session, error)
 }
 
 type transcodePermissionChecker interface {
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -952,9 +952,9 @@ func (h *PlaybackHandler) handleExpiredSession(session *playback.Session) {
 	}
 	sessionCopy := *session
 	go func() {
-		slog.Info("expired inactive playback session", "session", sessionCopy.ID, "playback_session_id", sessionCopy.ID,
-			"client_name", sessionCopy.ClientName, "client_version", sessionCopy.ClientVersion,
-			"client_build", sessionCopy.ClientBuild, "client_channel", sessionCopy.ClientChannel)
+		slog.Info("expired inactive playback session", append([]any{
+			"session", sessionCopy.ID, "playback_session_id", sessionCopy.ID,
+		}, sessionCopy.ClientInfo().LogAttrs()...)...)
 		// Expiry is a liveness reap, not a user stop — keep the recipe card so a
 		// resume reconstructs under the same id (the card's own TTL reaps it if
 		// the session is truly abandoned).
--- a/internal/api/handlers/playback_sessions.go
+++ b/internal/api/handlers/playback_sessions.go
@@ -48,7 +48,10 @@ type playbackSessionRow struct {
 	ClientIP                 string    `json:"client_ip,omitempty"`
 	ClientName               string    `json:"client_name,omitempty"`
 	ClientVersion            string    `json:"client_version,omitempty"`
+	ClientBuild              string    `json:"client_build,omitempty"`
+	ClientChannel            string    `json:"client_channel,omitempty"`
 	ClientLabel              string    `json:"client_label,omitempty"`
+	ClientLabelFull          string    `json:"client_label_full,omitempty"`
 	ClientUserAgent          string    `json:"client_user_agent,omitempty"`
 	AudioTrackIndex          int       `json:"audio_track_index"`
 	TranscodeAudio           bool      `json:"transcode_audio"`
@@ -89,6 +92,11 @@ type playbackSessionsCapabilitiesResponse struct {
 	EffectivePlayMethodValues []string `json:"effective_play_method_values"`
 	// IsJellyfinClient reports that rows carry is_jellyfin_client.
 	IsJellyfinClient bool `json:"is_jellyfin_client"`
+	// ClientBuild reports that rows carry client_build (and the exact-version
+	// client_label_full derived from it).
+	ClientBuild bool `json:"client_build"`
+	// ClientChannel reports that rows carry client_channel.
+	ClientChannel bool `json:"client_channel"`
 }
 
 // HandleGetSessionsCapabilities exposes additive feature support for the live
@@ -98,6 +106,8 @@ func (h *AdminHandler) HandleGetSessionsCapabilities(w http.ResponseWriter, _ *h
 		EffectivePlayMethod:       true,
 		EffectivePlayMethodValues: []string{"direct", "remux", "transcode", "audio"},
 		IsJellyfinClient:          true,
+		ClientBuild:               true,
+		ClientChannel:             true,
 	})
 }
 
@@ -182,6 +192,8 @@ func (l *PlaybackSessionsLoader) Load(
 			COALESCE(HOST(s.client_ip), ''),
 			COALESCE(s.client_name, ''),
 			COALESCE(s.client_version, ''),
+			COALESCE(s.client_build, ''),
+			COALESCE(s.client_channel, ''),
 			COALESCE(s.client_user_agent, ''),
 			COALESCE(s.audio_track_index, 0),
 			COALESCE(s.transcode_audio, FALSE),
@@ -239,6 +251,7 @@ func (l *PlaybackSessionsLoader) Load(
 			&posterPath,
 			&s.PlayMethod, &s.ReportingNode, &s.NodeDisplayName, &s.FileDuration, &s.StartedAt, &s.UpdatedAt,
 			&s.PositionSeconds, &s.IsPaused, &s.HasPlaybackControl, &s.ClientIP, &s.ClientName, &s.ClientVersion,
+			&s.ClientBuild, &s.ClientChannel,
 			&s.ClientUserAgent, &s.AudioTrackIndex, &s.TranscodeAudio, &streamBitrateKbps,
 			&s.TranscodeNodeURL, &s.TargetResolution, &s.TargetVideoCodec, &s.TargetAudioCodec, &targetBitrateKbps,
 			&s.TranscodeHWAccel, &s.SourceContainer, &sourceBitrateKbps, &s.SourceVideoCodec, &s.SourceVideoResolution,
@@ -253,6 +266,7 @@ func (l *PlaybackSessionsLoader) Load(
 		s.SourceBitrateKbps = sourceBitrateKbps
 		s.SourceAudioChannels = sourceAudioChannels
 		s.ClientLabel = playbackClientDisplayName(s.ClientName, s.ClientVersion, s.ClientUserAgent)
+		s.ClientLabelFull = playbackClientFullDisplayName(s.ClientName, s.ClientVersion, s.ClientBuild, s.ClientChannel, s.ClientUserAgent)
 		enrichPlaybackSessionRow(&s, audioTracksJSON)
 		sessions = append(sessions, s)
 	}
@@ -447,11 +461,50 @@ func firstNonEmptyValue(values ...string) string {
 	return ""
 }
 
+// playbackClientFullDisplayName renders the client's exact identity — version,
+// opaque build, and non-default channel — for surfaces that can afford the
+// width (expanded session details, tooltips). Clients that report no name fall
+// back to the compact user-agent label, which is all that can be derived there.
+func playbackClientFullDisplayName(name, version, build, channel, userAgent string) string {
+	name = strings.TrimSpace(name)
+	if name == "" {
+		return playbackClientDisplayName(name, version, userAgent)
+	}
+
+	label := name
+	if version = strings.TrimSpace(version); version != "" {
+		label += " " + version
+	}
+
+	build = strings.TrimSpace(build)
+	channel = strings.TrimSpace(channel)
+	// "release" is the assumed channel, so naming it adds width without adding
+	// information; every other channel is worth calling out.
+	if strings.EqualFold(channel, "release") {
+		channel = ""
+	}
+	var qualifiers []string
+	if build != "" {
+		qualifiers = append(qualifiers, "build "+build)
+	}
+	if channel != "" {
+		qualifiers = append(qualifiers, channel)
+	}
+	if len(qualifiers) > 0 {
+		label += " (" + strings.Join(qualifiers, ", ") + ")"
+	}
+	return label
+}
+
+// playbackClientDisplayName renders the compact label the session lists show.
+// A client that reports its own name keeps its version verbatim — truncating
+// "1.0.0" to "1" hid the only field that identifies the running build. Labels
+// derived from a user agent stay shortened, where a full browser version
+// ("120.0.6099.109") is noise.
 func playbackClientDisplayName(name, version, userAgent string) string {
 	name = strings.TrimSpace(name)
-	version = shortPlaybackClientVersion(version)
 	if name != "" {
-		if version != "" {
+		if version = strings.TrimSpace(version); version != "" {
 			return name + " " + version
 		}
 		return name
--- a/internal/api/handlers/playback_sessions.go
+++ b/internal/api/handlers/playback_sessions.go
@@ -266,7 +266,13 @@ func (l *PlaybackSessionsLoader) Load(
 		s.SourceBitrateKbps = sourceBitrateKbps
 		s.SourceAudioChannels = sourceAudioChannels
 		s.ClientLabel = playbackClientDisplayName(s.ClientName, s.ClientVersion, s.ClientUserAgent)
-		s.ClientLabelFull = playbackClientFullDisplayName(s.ClientName, s.ClientVersion, s.ClientBuild, s.ClientChannel, s.ClientUserAgent)
+		// Only worth the bytes when it says more than the compact label — which
+		// it does not for any client without a build or a non-release channel,
+		// i.e. most of a 200-row page. Clients read client_label when
+		// client_label_full is absent, so omitting the duplicate costs nothing.
+		if full := playbackClientFullDisplayName(s.ClientName, s.ClientVersion, s.ClientBuild, s.ClientChannel, s.ClientUserAgent); full != s.ClientLabel {
+			s.ClientLabelFull = full
+		}
 		enrichPlaybackSessionRow(&s, audioTracksJSON)
 		sessions = append(sessions, s)
 	}
@@ -463,17 +469,14 @@ func firstNonEmptyValue(values ...string) string {
 
 // playbackClientFullDisplayName renders the client's exact identity — version,
 // opaque build, and non-default channel — for surfaces that can afford the
-// width (expanded session details, tooltips). Clients that report no name fall
-// back to the compact user-agent label, which is all that can be derived there.
+// width (expanded session details, tooltips). The name-and-version half is
+// whatever the compact label resolved, so a client identified only by its user
+// agent still gets its build named: a client that bothered to report a build
+// has earned having it displayed, whether or not it also named itself.
 func playbackClientFullDisplayName(name, version, build, channel, userAgent string) string {
-	name = strings.TrimSpace(name)
-	if name == "" {
-		return playbackClientDisplayName(name, version, userAgent)
-	}
-
-	label := name
-	if version = strings.TrimSpace(version); version != "" {
-		label += " " + version
+	label := playbackClientDisplayName(name, version, userAgent)
+	if label == "" {
+		return ""
 	}
 
 	build = strings.TrimSpace(build)
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -462,16 +462,13 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
 	// user which version they are running.
 	clientInfo := playbackClientInfoForStartV3(r, req.ClientPlaybackContext)
 	if result.Terminal != nil {
-		slog.InfoContext(r.Context(), "playback plan decided", "component", "playback",
+		slog.InfoContext(r.Context(), "playback plan decided", append([]any{
+			"component", "playback",
 			"outcome", "terminal",
 			"reason", result.Terminal.Reason,
 			"file_id", effectiveFile.ID,
 			"quality_preference", req.QualityPreference,
-			"client_name", clientInfo.Name,
-			"client_version", clientInfo.Version,
-			"client_build", clientInfo.Build,
-			"client_channel", clientInfo.Channel,
-		)
+		}, clientInfo.LogAttrs()...)...)
 		response, persistErr := h.persistTerminalStartDecisionV3(r.Context(), userID, profileID, req, requestDigests, requestedFile.ID, effectiveFile.ID, playback.NewTerminalResponseV3(result.Terminal.Reason, result.Terminal.Message, result.Terminal.Retryable))
 		if persistErr != nil {
 			writeStartAttemptPersistenceErrorV3(w, persistErr)
@@ -486,7 +483,8 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
 	// One line per plan decision so route selection is reconstructible from
 	// server logs alone (finding a mis-planned route previously required
 	// correlating client logcat, ffmpeg commands, and session rows).
-	slog.InfoContext(r.Context(), "playback plan decided", "component", "playback",
+	slog.InfoContext(r.Context(), "playback plan decided", append([]any{
+		"component", "playback",
 		"outcome", "plan",
 		"decision_reason", result.Plan.DecisionReason,
 		"delivery", result.Plan.Delivery,
@@ -499,13 +497,9 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
 		"target_bitrate_kbps", result.TargetBitrateKbps,
 		"quality_preference", req.QualityPreference,
 		"bandwidth_estimate_kbps", intOrZeroHandlerV3(req.BandwidthEstimateKbps),
-		"client_name", clientInfo.Name,
-		"client_version", clientInfo.Version,
-		"client_build", clientInfo.Build,
-		"client_channel", clientInfo.Channel,
-	)
+	}, clientInfo.LogAttrs()...)...)
 	result.Plan.DegradationWarnings = append(result.Plan.DegradationWarnings, warnings...)
-	response, statusErr := h.startPlannedPlaybackV3(r, userID, profileID, req, requestDigests, requestedFile, effectiveFile, audioIndex, result)
+	response, statusErr := h.startPlannedPlaybackV3(r, userID, profileID, req, requestDigests, requestedFile, effectiveFile, audioIndex, result, clientInfo)
 	if statusErr != nil {
 		if statusErr.reason == "playback_attempt_reused" {
 			writeError(w, http.StatusConflict, "playback_attempt_reused", statusErr.message)
@@ -535,8 +529,19 @@ type playbackStartRequestDigestsV3 struct {
 // every request; the start body's client_playback_context is the fallback for
 // clients that report their app identity only there. All values stay opaque —
 // they are trimmed and length-clamped when the session stamps them.
+//
+// The fallback applies only to a client that named itself. client_playback_context
+// carries no app name, so nothing in the body can identify a nameless client
+// anyway — it is labelled from its user agent, and its app_version is a
+// free-form platform string rather than the marketing version client_version
+// promises. The web player, for one, reports the literal "web" there; taking it
+// unconditionally would write "web" into the one field that is contractually
+// semver, on every browser session.
 func playbackClientInfoForStartV3(r *http.Request, clientContext playback.ClientPlaybackContextV3) playback.ClientInfo {
 	info := playbackClientInfoFromRequest(r)
+	if info.Name == "" {
+		return info
+	}
 	if info.Version == "" {
 		info.Version = strings.TrimSpace(clientContext.AppVersion)
 	}
@@ -545,6 +550,38 @@ func playbackClientInfoForStartV3(r *http.Request, clientContext playback.Client
 	}
 	if info.Channel == "" {
 		info.Channel = strings.TrimSpace(clientContext.AppChannel)
+	}
+	return info
+}
+
+// playbackClientInfoWithSessionFallbackV3 completes a header-derived identity
+// from the session the event belongs to. Route events posted out of band carry
+// no client_playback_context, so without this a client that reports its build
+// only in the start body would attribute its plan_selected event to a build and
+// every later event of the same attempt to none.
+func (h *PlaybackHandler) playbackClientInfoWithSessionFallbackV3(sessionID string, info playback.ClientInfo) playback.ClientInfo {
+	if sessionID == "" || h.sessionMgr == nil {
+		return info
+	}
+	if info.Name != "" && info.Version != "" && info.Build != "" && info.Channel != "" {
+		return info
+	}
+	session, err := h.sessionMgr.GetSession(sessionID)
+	if err != nil || session == nil {
+		return info
+	}
+	stamped := session.ClientInfo()
+	if info.Name == "" {
+		info.Name = stamped.Name
+	}
+	if info.Version == "" {
+		info.Version = stamped.Version
+	}
+	if info.Build == "" {
+		info.Build = stamped.Build
+	}
+	if info.Channel == "" {
+		info.Channel = stamped.Channel
 	}
 	return info
 }
@@ -569,7 +606,7 @@ func (d playbackStartRequestDigestsV3) matches(stored string) bool {
 	return stored == "" || stored == d.current || stored == d.legacy
 }
 
-func (h *PlaybackHandler) startPlannedPlaybackV3(r *http.Request, userID int, profileID string, req playback.StartRequestV3, requestDigests playbackStartRequestDigestsV3, requestedFile, effectiveFile *models.MediaFile, audioIndex int, result playback.PlannerResultV3) (playback.DecisionResponseV3, *transportErrorV3) {
+func (h *PlaybackHandler) startPlannedPlaybackV3(r *http.Request, userID int, profileID string, req playback.StartRequestV3, requestDigests playbackStartRequestDigestsV3, requestedFile, effectiveFile *models.MediaFile, audioIndex int, result playback.PlannerResultV3, clientInfo playback.ClientInfo) (playback.DecisionResponseV3, *transportErrorV3) {
 	if result.Plan == nil {
 		return playback.DecisionResponseV3{}, &transportErrorV3{reason: "internal_error", message: "The server produced no playback plan."}
 	}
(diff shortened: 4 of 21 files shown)


Quick104 (commented):
Automated review of this branch (xhigh recall pass). 14 findings, ranked most-severe first in the inline comments below.

Verification done while reviewing, for what it's worth: `go vet` is clean on every touched package, `internal/playback/contract` passes (the schema conformance gate), and the new label tests pass. I also hand-checked the three SQL statements — the reconciler upsert is 29 columns / 28 placeholders + `NOW()` / 28 args, the route-events insert is 17/17/17, and the loader is 48 SELECT columns against 48 `rows.Scan` args. All correct.

The one I would fix before merge is the first: the web player already sends `app_version: "web"`, so the new body fallback writes the literal `"web"` into `client_version` for every browser session — the field this PR designates as the semver key for future version gating.

<sub>Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5).</sub>

internal/api/handlers/playback_v3.go
```diff
@@ -518,6 +530,25 @@ type playbackStartRequestDigestsV3 struct {
...
 
+// playbackClientInfoForStartV3 resolves the client's app identity for a v3
+// start request. The X-Silo-Client-* headers win because they are present on
+// every request; the start body's client_playback_context is the fallback for
+// clients that report their app identity only there. All values stay opaque —
+// they are trimmed and length-clamped when the session stamps them.
+func playbackClientInfoForStartV3(r *http.Request, clientContext playback.ClientPlaybackContextV3) playback.ClientInfo {
+	info := playbackClientInfoFromRequest(r)
+	if info.Version == "" {
+		info.Version = strings.TrimSpace(clientContext.AppVersion)
```
Quick104, line 546: **The web player's literal `"web"` now lands in `client_version`** _(correctness)_

`web/src/player/client-context-v3.ts:26` sets `WEB_APP_VERSION = "web"` and line 243 sends it as `client_playback_context.app_version` on every v3 start. The web app sends no `X-Silo-Client-Version` header (`web/src/api/client.ts:198` sends only `X-Silo-Client-Family`), so before this PR the field was empty for browser sessions.

With this fallback, `playback_sessions_sync.client_version`, the `plan_selected` row in `playback_route_events`, and every `playback plan decided` log line now record `client_version="web"` for the largest client population.

That is exactly the field the PR description reserves as the semver key: "any future minimum-client-version gating must key on `client_version`, which is semver." This change is what makes that untrue.

internal/playback/protocol_v3.go
```diff
@@ -887,7 +892,7 @@ func validateCapabilitiesV3(c *ClientCodecCapabilitiesV3, ctx *ClientPlaybackCon
 	if !validCapabilityEvidenceV3(c.AudioEvidence) {
 		return errors.New("audio_evidence is required and must be exact, platform_attested, or declared")
 	}
-	if len(c.CodecsVideo) > 64 || len(c.CodecsVideoHardware) > 64 || len(c.CodecsAudio) > 64 || len(c.Containers) > 64 || len(c.VideoDecode) > 64 || len(ctx.Deliveries) > 16 || len(ctx.Device.Platform) > 32 || len(ctx.FormFactor) > 32 || len(ctx.AppVersion) > 64 {
+	if len(c.CodecsVideo) > 64 || len(c.CodecsVideoHardware) > 64 || len(c.CodecsAudio) > 64 || len(c.Containers) > 64 || len(c.VideoDecode) > 64 || len(ctx.Deliveries) > 16 || len(ctx.Device.Platform) > 32 || len(ctx.FormFactor) > 32 || len(ctx.AppVersion) > 64 || len(ctx.AppBuild) > 64 || len(ctx.AppChannel) > 32 {
```
Quick104: **Over-long `app_build`/`app_channel` 400s the start request instead of clamping** _(correctness)_

`validateCapabilitiesV3` now rejects `len(ctx.AppBuild) > 64 || len(ctx.AppChannel) > 32` with "capability list exceeds supported size", which surfaces as a 400 on `POST /playback/start` — nothing plays.

The header path never reaches this check: `normalizeClientMetadataValue` (`internal/playback/session.go:314`) just clamps. So a client whose dev-channel build string is a 70-char `git describe` output plays fine when it sets `X-Silo-Client-Build`, and cannot play at all when it puts the same value in `client_playback_context.app_build`.

`docs/settings-api.md` (added in this PR) states "Values are trimmed and truncated to the clamp above" with no mention of rejection. An opaque, purely cosmetic diagnostic identifier should not be able to block playback — clamp it here (the function's own doc says it "validates and normalizes"), or document the hard failure.

internal/api/handlers/playback_sessions.go
```diff
@@ -447,11 +461,50 @@ func firstNonEmptyValue(values ...string) string {
 	return ""
 }
 
+// playbackClientFullDisplayName renders the client's exact identity — version,
+// opaque build, and non-default channel — for surfaces that can afford the
+// width (expanded session details, tooltips). Clients that report no name fall
+// back to the compact user-agent label, which is all that can be derived there.
+func playbackClientFullDisplayName(name, version, build, channel, userAgent string) string {
+	name = strings.TrimSpace(name)
+	if name == "" {
```
Quick104: **The full label silently discards build and channel when the client reports no name** _(correctness)_

Any client that sets `X-Silo-Client-Build` (or `app_build`) but not `X-Silo-Client` — the web player, and every jellycompat client whose `MediaBrowser` header omits `Client` — has `client_build` stored in `playback_sessions_sync` and advertised through the new `client_build` capability, yet this early return drops both build and channel on the floor.

The new "Client" card in the expanded panel — the surface added specifically to answer "which build is this?" — renders `Chrome 120` for such a session while the same row's `client_build` field says `5`.

The added test pins the behaviour (`no client name falls back to user agent label`, build `5` / channel `dev` → `Chrome 120`), so the gap is locked in rather than flagged. Appending the qualifiers to the user-agent label costs nothing.

internal/api/handlers/playback_v3.go
```diff
@@ -2708,7 +2739,7 @@ func (h *PlaybackHandler) HandlePlaybackRouteEventV3(w http.ResponseWriter, r *h
 	}
 	event.Diagnostics = sanitizeDiagnosticsV3(event.Diagnostics)
 	client := playbackClientInfoFromRequest(r)
-	h.enqueueRouteEventV3(playback.RouteEventRecordV3{RouteEventV3: event, UserID: userID, ProfileID: profileID, ClientName: client.Name, ClientVersion: client.Version, ClientModel: event.Diagnostics["device_model"]})
+	h.enqueueRouteEventV3(playback.RouteEventRecordV3{RouteEventV3: event, UserID: userID, ProfileID: profileID, ClientName: client.Name, ClientVersion: client.Version, ClientBuild: client.Build, ClientChannel: client.Channel, ClientModel: event.Diagnostics["device_model"]})
```
Quick104, line 2774: **The route-event endpoint has no body fallback, so build/channel disagree within one session** _(correctness)_

The body fallback in `playbackClientInfoForStartV3` exists for "clients that cannot set the headers on every request". For such a client, `plan_selected` (line 692, resolved `clientInfo`) and the replan events (line 2130, from `session.ClientBuild`) carry build/channel — but every out-of-band event posted here (`first_frame`, `plan_failed`, `stopped`, `terminal`) goes through `playbackClientInfoFromRequest(r)` and writes NULL.

Querying `playback_route_events` for "which build produced this failure" then returns a build for the selection event and nothing for the failure event of the same attempt — the opposite of the attributability this PR is for. Sourcing build/channel from the session (as `executeReplanV3` does) would make all events agree.

docs/design/schemas/playback-v3/v3/start-request.schema.json
```diff
@@ -531,6 +531,14 @@
           "type": "string",
           "maxLength": 64
         },
+        "app_build": {
```
Quick104, line 534: **`replan-request.schema.json` did not get the same two properties** _(contract-drift)_

`ReplanRequestV3` reuses `ClientPlaybackContextV3` (`internal/playback/protocol_v3.go:439`) and `validateReplanV3` calls `validateCapabilitiesV3` (line 869), so the new `app_build`/`app_channel` bounds are enforced on replan requests too — but only the start schema gained the properties. The replan schema's `$defs.client_playback_context` still lists only `protocol_version`/`form_factor`/`app_version`/`device`/`output`/`deliveries`.

`internal/playback/contract/contract_test.go` exists precisely because "a schema is data, and data does not fail to compile when a struct tag or a bound moves" — but it has no struct-vs-schema parity entry for `client_playback_context`, so nothing catches this. Android/Apple/web vendor these schemas; a client generated from the replan schema will not know the fields exist on that path.

internal/api/handlers/playback_v3.go
```diff
@@ -518,6 +530,25 @@ type playbackStartRequestDigestsV3 struct {
 	legacy  string
 }
 
+// playbackClientInfoForStartV3 resolves the client's app identity for a v3
+// start request. The X-Silo-Client-* headers win because they are present on
+// every request; the start body's client_playback_context is the fallback for
+// clients that report their app identity only there. All values stay opaque —
+// they are trimmed and length-clamped when the session stamps them.
+func playbackClientInfoForStartV3(r *http.Request, clientContext playback.ClientPlaybackContextV3) playback.ClientInfo {
```
Quick104, line 540: **The only new behavioural rule in the PR is untested** _(test-coverage)_

This function encodes three independent decisions: the header wins per field (not per struct), the body is consulted only when the header is empty, and nothing falls back for `Name`. None is exercised — `TestPlaybackClientDisplayNameKeepsNamedClientVersionExact` and `TestPlaybackClientFullDisplayName` cover only the pure string formatters, and no handler test posts a v3 start with `app_build`/`app_channel` and asserts what the session was stamped with.

That absence is why the `"web"` version-fallback regression above ships silently: a table test with (header empty, body `"web"`) would have made the stored value visible during review.

internal/playback/session.go
```diff
@@ -453,6 +463,8 @@ func newSession(
 		IsPaused:             false,
 		ClientName:           normalizeClientMetadataValue(clientInfo.Name, 128),
 		ClientVersion:        normalizeClientMetadataValue(clientInfo.Version, 64),
+		ClientBuild:          normalizeClientMetadataValue(clientInfo.Build, 64),
```
Quick104: **Byte-wise truncation of client metadata can break a whole node's session sync** _(correctness)_

`normalizeClientMetadataValue` (line 314) slices by bytes, not runes. Go's HTTP server accepts bytes `>= 0x80` in header values, so `X-Silo-Client-Build` may carry multi-byte UTF-8; `value[:maxLen]` can cut mid-rune (e.g. a 70-byte build whose 64th byte lands inside a 3-byte sequence), producing an invalid UTF-8 string.

That string is stored on the `Session`, carried through `buildLiveSessionSync`, and passed to the `playback_sessions_sync` upsert. Postgres rejects it with `invalid byte sequence for encoding "UTF8"`, `ReconcileNodeSessions` returns an error, and because the upserts and the stale-row DELETE share one transaction (`internal/worker/reconciler.go:132-223`) the **entire node's** session snapshot fails to sync — not just the offending session.

The flaw predates this PR for `ClientName`/`ClientVersion`/`ClientUserAgent`, but this hunk widens the exposure to two more client-controlled fields; a rune-boundary truncation fixes all five at once.

web/src/pages/adminActivityPresentation.ts
```diff
@@ -245,6 +245,31 @@ export function getSessionClientLabel(session: AdminSession): string {
   return clientName || "";
 }
 
+/**
+ * The exact client identity — version, build, and non-release channel — for
+ * surfaces that can afford the width (expanded session details, tooltips).
+ * `getSessionClientLabel` stays the compact list label; this one must never
+ * replace it in a fixed-width row.
+ */
+export function getSessionClientLabelFull(session: AdminSession): string {
```
Quick104, line 254: **This re-implements `getSessionClientLabel` verbatim** _(simplification)_

Lines 260-270 duplicate lines 235-245 character for character — the `client_label` branch, the name+version branch, and the `clientName || ""` fallback. The whole function is:

```ts
return session.client_label_full?.trim() || getSessionClientLabel(session);
```

As written, any future change to the compact fallback chain has to be made twice or the two labels silently diverge in the same row. Repo `CLAUDE.md`: "Prefer extracting shared logic over duplicating it, and prefer changing existing code over bolting a local workaround onto it." The server-side twin already delegates — `playbackClientFullDisplayName`'s empty-name branch calls `playbackClientDisplayName` — so only this copy duplicates.

internal/api/handlers/playback_v3.go
```diff
@@ -457,20 +457,28 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
...
+	// terminal reported against one build is attributable without asking the
+	// user which version they are running.
+	clientInfo := playbackClientInfoForStartV3(r, req.ClientPlaybackContext)
 	if result.Terminal != nil {
 		slog.InfoContext(r.Context(), "playback plan decided", "component", "playback",
 			"outcome", "terminal",
 			"reason", result.Terminal.Reason,
 			"file_id", effectiveFile.ID,
 			"quality_preference", req.QualityPreference,
+			"client_name", clientInfo.Name,
```
Quick104: **The four client-identity attrs are pasted at three call sites** _(reuse)_

The identical four key/value pairs appear here (terminal), at line 502 (plan), and at `internal/api/handlers/playback.go:955-957` (session expiry). `playback.ClientInfo` already owns exactly these four fields, so a `func (c ClientInfo) LogAttrs() []any` (or a `slog.Group("client", …)`) would keep the key names in one place.

Today, renaming `client_build` to `app_build` for consistency with the diagnostics contract — a rename the PR description explicitly cites as the naming precedent — means finding three separate literal blocks in two packages, and missing one produces logs where the same session is keyed two different ways.

internal/api/handlers/playback_v3.go
```diff
@@ -551,7 +582,7 @@ func (h *PlaybackHandler) startPlannedPlaybackV3(r *http.Request, userID int, pr
 			return playback.DecisionResponseV3{}, &transportErrorV3{reason: reason, message: "The selected server adaptation is disabled for this user."}
 		}
 	}
-	clientInfo := playbackClientInfoFromRequest(r)
+	clientInfo := playbackClientInfoForStartV3(r, req.ClientPlaybackContext)
```
Quick104: **Client identity is resolved twice per start request** _(efficiency)_

`handleStartPlaybackV3` already computed `clientInfo` at line 463 for the decision log and the terminal route event; this discards that work and recomputes the identical value from the same `*http.Request`. Every v3 start therefore does 8 `Header.Get` calls and 12 `TrimSpace` calls where 4 and 6 would do.

`startPlannedPlaybackV3` already takes 9 parameters — passing the resolved `clientInfo` (which ends up on the context at line 586 anyway) removes the second parse.

internal/api/handlers/playback_sessions.go
```diff
@@ -253,6 +266,7 @@ func (l *PlaybackSessionsLoader) Load(
 		s.SourceBitrateKbps = sourceBitrateKbps
 		s.SourceAudioChannels = sourceAudioChannels
 		s.ClientLabel = playbackClientDisplayName(s.ClientName, s.ClientVersion, s.ClientUserAgent)
+		s.ClientLabelFull = playbackClientFullDisplayName(s.ClientName, s.ClientVersion, s.ClientBuild, s.ClientChannel, s.ClientUserAgent)
```
Quick104: **`client_label_full` is emitted even when byte-identical to `client_label`** _(efficiency)_

For every session with no build and no non-release channel — every browser, every jellycompat client, every pre-update native client — `playbackClientFullDisplayName` returns exactly what `playbackClientDisplayName` returned, so the two fields are byte-identical. `omitempty` cannot help because the value is non-empty.

On the `LIMIT 200` admin sessions response that is up to 200 redundant strings per poll, on an endpoint the Activity page refreshes on a timer. Setting `ClientLabelFull` only when it differs keeps the wire contract — the web fallback at `adminActivityPresentation.ts:260` already falls through to `client_label`.

internal/api/handlers/playback_v3.go
```diff
@@ -491,6 +499,10 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
 		"target_bitrate_kbps", result.TargetBitrateKbps,
 		"quality_preference", req.QualityPreference,
 		"bandwidth_estimate_kbps", intOrZeroHandlerV3(req.BandwidthEstimateKbps),
+		"client_name", clientInfo.Name,
```
Quick104: **Empty client attrs are persisted to `opslog` on every plan decision** _(efficiency)_

For browser and jellycompat sessions `clientInfo.Name`/`Version`/`Build`/`Channel` are all empty (or, after the fallback, three of four). `slog` records them anyway, and the `opslog` handler (`cmd/silo/main.go:237`) persists the attrs map, so `/admin/logs` accumulates rows carrying `client_build=""` and `client_channel=""` forever.

The PR description's own reason for leaving `activity_log` untouched is this volume argument — "highest-volume table, and the value is constant per device." Appending the attrs only when non-empty keeps the diagnostic value and drops the noise.

web/src/pages/AdminActivity.tsx
```diff
@@ -528,6 +529,13 @@ function StreamRow({
   const streamMeta = [sourceContainer, streamBitrate].filter(Boolean).join(" · ");
   const clientIP = session.client_ip?.trim() || "";
   const clientLabel = getSessionClientLabel(session);
+  // The row stays compact; the exact version, build, and channel live in the
+  // tooltip and in the expanded panel's Client card.
+  const clientLabelFull = getSessionClientLabelFull(session);
```
Quick104, line 537: **The Activity search cannot match the build this PR surfaces** _(search-coverage)_

The filter at line 153 still checks `getSessionClientLabel(s)` — the compact label, which by design excludes build and channel — plus `client_user_agent` and `client_ip`. `client_build` and `client_channel` are added to the `AdminSession` type (`web/src/api/types.ts:2448-2449`) but are never read anywhere in `web/src`.

So the primary workflow this feature enables — "a bug report came in from build 5, show me those sessions" — returns zero rows. Swapping in `getSessionClientLabelFull(s)`, or adding the two fields to the predicate, makes the newly surfaced identity searchable.

internal/playback/protocol_v3.go
```diff
@@ -331,12 +331,17 @@ type DeliveryCapabilityV3 struct {
...
-	Device          DeviceContextV3                 `json:"device"`
-	Output          OutputContextV3                 `json:"output"`
-	Deliveries      map[string]DeliveryCapabilityV3 `json:"deliveries"`
+	ProtocolVersion int    `json:"protocol_version"`
+	FormFactor      string `json:"form_factor"`
+	AppVersion      string `json:"app_version"`
+	// AppBuild and AppChannel are the request-body fallback for the
+	// X-Silo-Client-Build / X-Silo-Client-Channel headers. Both are opaque
+	// strings the server stores verbatim.
+	AppBuild   string                          `json:"app_build,omitempty"`
```
Quick104, line 340: **The web player's TypeScript mirror of this type was not updated** _(contract-drift)_

`web/src/player/protocol-v3.ts:251-263` declares `ClientPlaybackContextV3` and did not gain `app_build`/`app_channel`, while this struct and `start-request.schema.json` did.

The web player is one of the three clients the checked-in schemas exist for, and `web/src/player/client-context-v3.ts:241-250` is the single place that builds the context. The follow-up that teaches the web app a real version from the git tag (called out in the PR description) will need a type change here first. Nothing fails today; the cost is that the drift stays invisible until someone tries to use the field.

Quick104:
All 14 review findings are addressed in b43b7ef06. Mapping, in the order they were raised:

| # | Finding | Fix |
| - | ------- | --- |
| 1 | Web player's `"web"` lands in `client_version` | The body fallback now applies only to a client that sent `X-Silo-Client`. `client_playback_context` has no app-name field, so nothing in the body can identify a nameless client anyway — it is labelled from its user agent, and its `app_version` is a free-form platform string rather than the marketing version `client_version` promises. Two test cases pin the `"web"` case specifically. |
| 2 | Over-long `app_build`/`app_channel` 400s the start | `validateCapabilitiesV3` clamps both with `normalizeClientMetadataValue` — the same helper the header path uses — instead of rejecting. An opaque identity label can no longer fail a playback start, and the doc's "trimmed and truncated" claim is now true on both routes. |
| 3 | Full label drops build/channel with no client name | `playbackClientFullDisplayName` now qualifies whatever the compact formatter resolved, so a user-agent-labelled session reads `Chrome 120 (build 5, dev)`. This also deletes its duplicated name+version assembly. Test expectation updated, plus a case for the no-qualifiers path. |
| 4 | Route-event endpoint has no fallback | `HandlePlaybackRouteEventV3` fills empty identity fields from the session the event belongs to (after the existing ownership check), matching what `executeReplanV3` already did. All events of one attempt now agree. |
| 5 | `replan-request.schema.json` missing the properties | Added. Plus `TestRequestSchemasShareIdenticalDefs`, which asserts every `$def` the two request schemas share is byte-identical — they deserialize the same Go types through the same validator, so this cannot drift again. (Every shared `$def` was identical before this PR; `client_playback_context` was the only one that had diverged.) |
| 6 | `playbackClientInfoForStartV3` untested | `TestPlaybackClientInfoForStartV3`: 6 cases covering header-wins, per-field fallback, the nameless-client rule, and whitespace-only headers. |
| 7 | Byte-wise truncation can break a node's session sync | `normalizeClientMetadataValue` clamps on a rune boundary via `strings.ToValidUTF8`. Fixes it for `client_name`/`client_version`/`client_user_agent` at the same time. |
| 8 | `getSessionClientLabelFull` duplicates `getSessionClientLabel` | Now one line delegating to it. |
| 9 | Four log attrs pasted at three call sites | `ClientInfo.LogAttrs()` is the single definition of those keys; all three sites use it. Added `Session.ClientInfo()` so the expiry path and the route-event fallback share one accessor. |
| 10 | Identity resolved twice per start | `startPlannedPlaybackV3` takes the resolved `clientInfo` instead of re-parsing headers. |
| 11 | `client_label_full` duplicated on every row | Set only when it differs from `client_label`. Clients already fall through to `client_label` when it is absent. |
| 12 | Empty attrs persisted to opslog | `LogAttrs()` omits fields the client did not report, so browser and jellycompat decisions no longer write four empty keys. |
| 13 | Activity search can't match a build | The filter uses `getSessionClientLabelFull`. |
| 14 | Web `ClientPlaybackContextV3` type missing the fields | Added as optional, documented as omitted by the web player. |

**Behavioural changes worth calling out for reviewers of the original PR**, since they revise claims in the description:

- `client_version` stays empty for web sessions rather than becoming `"web"`. The follow-up that teaches the web app a real version is still worth doing, but it is no longer load-bearing for the semver contract.
- The compact label for a nameless client is unchanged (`Chrome 120`); only the *full* label gained the qualifiers.
- `client_label_full` is now absent rather than duplicated when it would equal `client_label` — additive and `omitempty`, so no client sees a field change.

## Verification

```
gofmt -l internal cmd                         # clean
go vet ./internal/... ./cmd/...               # clean
go test ./internal/api/handlers/              # ok   86.755s
go test ./internal/playback/...               # ok   (incl. contract, planstore)
go test ./internal/worker/                    # ok
```

Web (deps installed fresh in the worktree):

```
pnpm exec tsc -b                              # clean
pnpm run format:check                         # clean
pnpm run lint                                 # 0 errors (151 pre-existing warnings, all vendor/react-refresh)
pnpm exec vitest run \
  src/pages/adminActivityPresentation.test.ts \
  src/player/client-context-v3.test.ts \
  src/player/playback-info.test.ts \
  src/player/hooks/usePlaybackSession.test.ts # 65 passed
make verify-local-paths                       # clean
```

`golangci-lint` is not installed on this machine, so `make lint` was not run; CI's `--new-from-merge-base` pass still gates it.

Nothing was added to `WEBTEST_KNOWN_FAILURES`, and no test carries a new `t.Skip`.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: review and follow-up commit fully AI-generated. The review was an independent xhigh-recall pass over the diff; findings 1, 2, 4 and 7 are defects that would have shipped, and finding 5 is a contract divergence now guarded by a test rather than by care.

internal/api/handlers/playback_v3.go
```diff
@@ -491,6 +499,10 @@ func (h *PlaybackHandler) handleStartPlaybackV3(w http.ResponseWriter, r *http.R
...
+		"client_version", clientInfo.Version,
+		"client_build", clientInfo.Build,
+		"client_channel", clientInfo.Channel,
```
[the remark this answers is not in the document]
Quick104: Fixed in 143a7e622. `ClientInfo.Normalized()` is now the single definition of the 128/64/64/32/512 bounds and runs at the request boundary in `playbackClientInfoFromRequest` (and at the end of `playbackClientInfoForStartV3` for body-sourced values), so the plan-decision log and `playback_route_events` observe the clamp instead of only `newSession`. `TestPlaybackClientInfoFromRequestClampsHeaders` covers oversized headers.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/api/handlers/playback.go
```diff
@@ -1076,9 +1078,13 @@ func playbackClientInfoFromRequest(r *http.Request) playback.ClientInfo {
 	if r == nil {
 		return playback.ClientInfo{}
 	}
+	// Build and channel are opaque: they are trimmed and length-clamped when the
+	// session stamps them, never parsed or validated against an enum here.
 	return playback.ClientInfo{
 		Name:      strings.TrimSpace(r.Header.Get("X-Silo-Client")),
 		Version:   strings.TrimSpace(r.Header.Get("X-Silo-Client-Version")),
+		Build:     strings.TrimSpace(r.Header.Get("X-Silo-Client-Build")),
+		Channel:   strings.TrimSpace(r.Header.Get("X-Silo-Client-Channel")),
```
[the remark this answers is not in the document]
Quick104: Fixed in 143a7e622 — same root cause as the Codex note on `playback_v3.go`. The clamp moved into `ClientInfo.Normalized()` applied in this helper, so logs and `RouteEventRecordV3` no longer bypass the documented 64/32 limits; `newSession` still normalizes because identities also arrive from the jellycompat and Audiobookshelf surfaces. Coverage: `TestPlaybackClientInfoFromRequestClampsHeaders`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/api/handlers/playback_v3.go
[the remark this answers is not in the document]
Quick104: Already fixed in b43b7ef06, before this review landed — this was finding 4 of the human-requested review pass. `HandlePlaybackRouteEventV3` now calls `playbackClientInfoWithSessionFallbackV3`, which fills empty identity fields from the session the event belongs to (after the existing ownership check), matching what `executeReplanV3` already did.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/playback/protocol_v3.go
```diff
@@ -887,7 +892,7 @@ func validateCapabilitiesV3(c *ClientCodecCapabilitiesV3, ctx *ClientPlaybackCon
 	if !validCapabilityEvidenceV3(c.AudioEvidence) {
 		return errors.New("audio_evidence is required and must be exact, platform_attested, or declared")
 	}
-	if len(c.CodecsVideo) > 64 || len(c.CodecsVideoHardware) > 64 || len(c.CodecsAudio) > 64 || len(c.Containers) > 64 || len(c.VideoDecode) > 64 || len(ctx.Deliveries) > 16 || len(ctx.Device.Platform) > 32 || len(ctx.FormFactor) > 32 || len(ctx.AppVersion) > 64 {
+	if len(c.CodecsVideo) > 64 || len(c.CodecsVideoHardware) > 64 || len(c.CodecsAudio) > 64 || len(c.Containers) > 64 || len(c.VideoDecode) > 64 || len(ctx.Deliveries) > 16 || len(ctx.Device.Platform) > 32 || len(ctx.FormFactor) > 32 || len(ctx.AppVersion) > 64 || len(ctx.AppBuild) > 64 || len(ctx.AppChannel) > 32 {
 		return errors.New("capability list exceeds supported size")
```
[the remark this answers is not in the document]
Quick104: Partly fixed, partly by design, in b43b7ef06 + 143a7e622. The rejection is gone: `validateCapabilitiesV3` now clamps `app_build`/`app_channel` instead of failing the request, so a schema-valid value can no longer 400. And `normalizeClientMetadataValue` now counts runes, so the clamp matches the schema's `maxLength` exactly — your 20-character emoji channel is kept whole. `app_version`'s byte-length rejection is unchanged: it predates this PR and altering an existing accept/reject boundary belongs in its own change.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/playback/session.go
```diff
@@ -453,6 +463,8 @@ func newSession(
...
+		ClientChannel:        normalizeClientMetadataValue(clientInfo.Channel, 32),
```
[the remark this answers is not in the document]
Quick104: Not changed — investigated and this is pre-existing and broader than build/channel, so fixing it here would be the wrong scope.

The only writers of a card's client fields are jellycompat: `streams.go:1432-1433` and `handlers_playback.go:555-556`. Jellyfin clients have no build concept at all — the `MediaBrowser` auth vocabulary is `Client`/`Device`/`DeviceId`/`Version` — so adding `ClientBuild`/`ClientChannel` to `RecipeCard` would be dead fields on both paths.

Native sessions reconstruct from `identityRecipeCard` (`playback.go:507`), which sets no client metadata whatsoever. So a native session that resumes after a restart already loses `client_name`, `client_version` and `client_user_agent` too — build and channel are not specially dropped, they are dropped identically to the fields that shipped before this PR. That gap is real and worth its own issue, but it is not this PR's regression and closing it means deciding what a native card should carry, not appending two strings.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/playback/session.go
```diff
@@ -304,7 +354,11 @@ func (m *SessionManager) SetExpirationHook(fn func(*Session)) {
 func normalizeClientMetadataValue(value string, maxLen int) string {
 	value = strings.TrimSpace(value)
 	if maxLen > 0 && len(value) > maxLen {
-		value = value[:maxLen]
+		// Clamp on a rune boundary. Header values may carry multi-byte UTF-8, and
+		// a mid-rune byte slice produces a string Postgres refuses outright — the
+		// per-node session upserts share one transaction, so a single malformed
+		// client string would fail that whole node's sync rather than one row.
+		value = strings.ToValidUTF8(value[:maxLen], "")
```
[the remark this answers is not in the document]
Quick104: Good catch — fixed in e61c705c4. You are right that NUL is valid UTF-8, so `utf8.ValidString` accepts it, `strings.ToValidUTF8` leaves it, and `TrimSpace` does not treat it as whitespace; a JSON NUL escape in a v3 start body would have reached the text column and failed the whole per-node reconcile transaction. `normalizeClientMetadataValue` now strips control characters outright rather than NUL alone, since none of them belong in a label rendered in the admin UI and written to structured logs. `TestPlaybackClientInfoStripsControlCharacters` covers it through the body path. (Confirmed headers cannot carry it — net/http rejects bytes below 0x20 — so the body path this PR added was the only exposure.)

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

internal/playback/session.go
```diff
@@ -302,9 +369,27 @@ func (m *SessionManager) SetExpirationHook(fn func(*Session)) {
 }
 
 func normalizeClientMetadataValue(value string, maxLen int) string {
+	// Header values are raw bytes, and a text column will not take invalid
+	// UTF-8: Postgres rejects the whole statement, and the per-node session
+	// upserts share one transaction, so one malformed client string would fail
+	// that entire node's sync rather than a single row.
+	if !utf8.ValidString(value) {
```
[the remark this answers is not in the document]
Quick104: Confirmed and fixed in e61c705c4 — same issue Codex raised on the previous commit. `normalizeClientMetadataValue` now strips control characters outright (not just NUL), since NUL is valid UTF-8 and so passed both `utf8.ValidString` and `TrimSpace`. `TestPlaybackClientInfoStripsControlCharacters` pins it through the v3 body path, which was the only exposure — net/http rejects bytes below 0x20 in headers.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Quick104:
Bot review round addressed — two more commits.

**143a7e622** — the 64/32 clamp only ran where `newSession` stamped its fields, but the resolved `ClientInfo` is written straight to the plan-decision log and to `playback_route_events`, so a header-sized build reached both (raised by both Codex and CodeRabbit). `ClientInfo.Normalized()` is now the single definition of those bounds and runs at the request boundary. It also clamps by **runes** rather than bytes, which closes Codex's separate point that the published `maxLength` counts characters — a 32-character emoji channel was 128 bytes.

**e61c705c4** — a JSON NUL escape in the v3 start body's `app_version`/`app_build`/`app_channel` decodes to a real NUL, which is valid UTF-8 and therefore survived both the UTF-8 repair and `TrimSpace`, and Postgres refuses NUL in `text`. Since the per-node session upserts share one transaction, one such start would have stopped every live session on that node from reconciling. Control characters are now stripped outright. Found independently by Codex and Cursor; headers were never exposed (net/http rejects bytes below 0x20), only the body path this PR introduced.

Two suggestions were **not** taken:

- **Persist build/channel through recipe-card reconstruction.** The only writers of a card's client fields are jellycompat (`streams.go:1432-1433`, `handlers_playback.go:555-556`), and Jellyfin clients have no build concept, so the fields would be dead on both paths. Native sessions reconstruct from `identityRecipeCard`, which sets no client metadata at all — so `client_name`/`client_version`/`client_user_agent` are already lost on that path today. Build and channel are not specially dropped; that gap is real but pre-existing and wants its own change. Thread left open for a maintainer call.
- **Rune-count `app_version`'s bound.** Its byte-length rejection predates this PR; changing an existing accept/reject boundary belongs in its own change. The two fields this PR adds now clamp by runes and never reject.

### Verification (at e61c705c4)

```
gofmt -l internal cmd                    # clean
go vet ./internal/... ./cmd/...          # clean
go test ./internal/api/handlers/         # ok   86.611s
go test ./internal/playback/...          # ok   (incl. contract, planstore)
go test ./internal/worker/               # ok
make verify-local-paths                  # clean
```

`internal/jellycompat` `TestBeginWebOperation{RecoversDeadProcessLock,RejectsLiveProcessLock}` fail — **verified they fail identically at the merge base `ebc99ac09`**, so they are pre-existing and unrelated, matching what the description already noted. `golangci-lint` is not installed on this machine; CI's `--new-from-merge-base` pass still gates lint.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: fully AI-generated. Each bot finding was verified against current `HEAD` before acting — two were already fixed by the earlier review pass, one was declined with reasoning above.

Quick104:
CI `Go` failure fixed, plus the CodeRabbit nitpick.

**4993d7b76 — the actual CI failure.** `golangci-lint --new-from-merge-base` flagged two things on lines my earlier commit touched:

- `goconst`: reformatting the two decision-log calls into slice literals pulled their `"component"` key into the changed-lines window, where it collides with the existing `logComponentKey` constant. Both now use the constant. The file's other 16 occurrences are left alone — CI only requires the lines a branch touches to be clean, and rewriting them would bury this change in unrelated churn.
- `misspell`: `labelled` → `labeled`, in the Go comment and in `docs/settings-api.md`.

I have now installed `golangci-lint v2.12.2` locally and run **the exact command CI runs** (`--new-from-merge-base=origin/main ./...`, with `origin/main` freshly fetched): **0 issues**. That gap is what let the first failure through, and it is closed.

**44f8474e2 — CodeRabbit's nitpick on `startPlannedPlaybackV3`.** Taken. The type assertion on `sessionStarterWithFilesContext` fell back to the context-free `StartSessionWithFiles`, and the context is precisely how the reporting client's identity reaches the new session — so a non-conforming implementation would start sessions with no client name, version, build or channel, silently. `SessionManagerInterface` now requires the method, making that a compile error instead. Production already implemented it; the one test double gained a three-line method. It is pre-existing code, but it is this PR's data that the fallback drops, so it was worth taking rather than deferring.

The earlier Codex and Cursor NUL reports were already fixed in e61c705c4 and their threads replied to and resolved.

### Verification (at 44f8474e2)

```
golangci-lint run --new-from-merge-base=origin/main ./...   # 0 issues (v2.12.2, same as CI)
gofmt -l internal cmd                                       # clean
go vet ./internal/... ./cmd/...                             # clean
go test ./internal/api/handlers/                            # ok  86.615s
go test ./internal/playback/... ./internal/worker/           # ok
```

`internal/jellycompat` `TestBeginWebOperation{RecoversDeadProcessLock,RejectsLiveProcessLock}` still fail — verified they fail identically at the merge base `ebc99ac09`, so pre-existing and unrelated.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: fully AI-generated.

Issue #599, opened 2026-08-11
[bug] CI runs no PostgreSQL, so 171 DB-backed tests silently skip — 7 are broken on main and 2 packages are flaky


jasonthagerty:
### What happened

> **Scope, for `/issue-to-pr`:**
> - **Type:** bug, infrastructure → `fix/` branch prefix.
> - **Client surface:** none. No API, auth, playback, session, library, or metadata behavior
>   changes; no coordinated `silo-apple` / `silo-android` follow-up.
> - **Linked scope:** not tied to a v1 capability epic — this is repo infrastructure, not a
>   user-facing capability. **Related to #385**, which proposed the CI lane this issue's evidence
>   bears on; see "Prior art" immediately below. Happy to file it under an epic instead if you'd
>   rather it were tracked there.
> - **Sign-off required before implementation.** This issue asks for a CI change and an
>   architectural decision about test isolation, which are two of `issue-to-pr`'s own stop-and-ask
>   triggers ("a test/CI change would be needed to *pass* — get sign-off, don't do it silently"
>   and "the change balloons into something large or architectural — propose a plan first"). An
>   agent should stop here and get a decision rather than start. A phased plan is in
>   [Technical notes](#technical-notes).

**Prior art: this overlaps #385, and I'd rather build on it than re-file it.** #385 (2026-07-11) already noted that "DB tests actually skip on a missing `SILO_TEST_DATABASE_URL`", and listed "a Postgres service container to unlock the 28 self skipping DB test files" among its optional extras. I found it after doing this work, and it deserves the credit for spotting the gap first.

Two things have changed since, which is why I think this is still worth its own report:

1. **Its premise has partly been overtaken.** #385 was written when "nothing runs on pull requests at all". `.github/workflows/ci.yml` landed on 2026-07-30 in `dc4b9a09` (part of #479, the settings contract) — so the test lane now exists. The Postgres container did not come with it.
2. **Unlocking these tests is not just provisioning.** #385 reasonably assumed supplying a database would let them run. It does not: 7 fail deterministically, 2 packages are flaky, and one failure is a production defect rather than test rot. That's what this issue documents, and it changes the "optional extra" framing — the extra has to happen *and* it uncovers work.

The count has also grown: 28 self-skipping DB test files then, **48** now.

I'm not claiming the implementation. #385's author offered to do this work and should keep it if they want it; treat this as evidence for that effort rather than competition with it.

---

[...]

PR #645, opened 2026-08-15
fix(watchstate): make series mark-watched atomic and unload-proof

Quick104:
## Problem

Marking a large series as watched in the web UI only marked *some* episodes when the user pressed back or closed the tab quickly.

The cause is server-side. `POST /watched/{id}` expands a series to every episode and then does **four sequential DB round-trips per episode**, all on the request context with no transaction:

- `episodeTargets` → `contentDurationSeconds` — one file query per episode (`internal/api/handlers/items.go`)
- `recordMarkWatched` loops per target: `MarkWatched` upsert → `ResolveHistoryIdentity` (itself an episode lookup **plus** a series provider-ID lookup) → history insert (`internal/watchstate/service.go`)

A 200-episode series is roughly 800 sequential queries. Each iteration commits independently, so when the browser cancels the request — closing the tab is the reliable trigger — the loop dies partway and everything already committed stays committed. That is exactly the partial mark users see.

The client compounded it: the watched mutation used a plain fetch with no `keepalive` and no optimistic update, so the button sat pending for the whole slow round-trip and the request died with the document.

Season marks took the identical path.

## Approach

**Batch the reads.** `episodeTargets` now resolves durations through the existing batched `listEpisodeFiles` helper rather than one query per episode. Extracted `mediaFileDurationSeconds` and reused it in `contentDurationSeconds` so the batched and single-item paths cannot drift.

**Bulk identity resolution.** New `ResolveHistoryIdentities` resolves a whole series in one episode query plus one provider-ID query per distinct series. Both batch lookups sit behind optional-capability interfaces with per-ID fallbacks, mirroring the existing `VisibleHistoryAdder` pattern — the narrow test fakes still compile unchanged.

**One transactional write.** New `userstore.WatchedBatchWriter` capability with a package-level helper that falls back to today's per-target loop. Implemented transactionally in both Postgres and SQLite. `recordMarkWatched` and jellycompat's `recordMarkWatchedBatch` both route through it, so the two share one implementation.

Notably this needs **no `context.WithoutCancel`**. With a single transaction, a cancelled request rolls back to a clean "nothing marked" state — all-or-nothing is the correct semantic here, and strictly better than a detached goroutine racing the response.

**Client.** `keepalive: true` so the request survives navigation and tab close, plus an optimistic `played` flip with rollback following the established `favorites.ts` pattern. Covers series, season, and episode in both directions, since all three call sites share the one hook.

## Two things worth reviewer attention

1. **jellycompat duration semantics.** Routing `recordMarkWatchedBatch` through the shared path would have clobbered known durations with `0`, because that caller supplies none and its previous `MarkProgressBatch` left the column untouched. The batch upsert now only ever *advances* a duration, never clears it. Pinned by a conformance assertion.

2. **SQLite context handling.** `internal/userdb` ignores `ctx` by package convention. My cancellation test failed at first because of it. I plumbed context through this one function only — cancellation safety is the entire point of the transaction — and said so in a comment at the call site.

## Testing

New coverage:

- `internal/userstore/storetest/suite.go` — shared conformance case for `MarkWatchedBatch`, run by **both** backends: per-target durations land, exactly one history row per target, the hidden-history watermark still pushes a re-mark back into visibility, a zero duration does not clear a known one, and blank IDs are compacted out.
- `internal/watchstate/service_test.go` — a 25-episode series resolves identities in **exactly one** episode lookup and one provider-ID lookup (asserted via call counters, so an O(n) regression fails the test), and a 40-episode mark on a cancelled context leaves **zero** rows.
- `web/src/hooks/queries/items.test.ts` — `keepalive` is sent, and the optimistic `played` flip applies then rolls back on error.

**The regression test was verified to actually bite.** With the transactional path stubbed back out to the per-item fallback, `TestManualMarkWatchedSeriesIsAtomicOnCancel` fails:

```
--- FAIL: TestManualMarkWatchedSeriesIsAtomicOnCancel (0.00s)
    service_test.go:816: RecordManualMarkWatchedWithResult on a cancelled context returned nil error
FAIL
```

Rather than let the Postgres tests skip, I stood up a `pgvector/pgvector:pg18` container, applied all migrations, and ran the real Postgres path:

```
=== RUN   TestPostgresMarkWatchedBatch/MarkWatchedBatch
--- PASS: TestPostgresMarkWatchedBatch (0.05s)
ok  	github.com/Silo-Server/silo-server/internal/userstore/pgstore	0.163s
=== RUN   TestSQLiteMarkWatchedBatch/MarkWatchedBatch
--- PASS: TestSQLiteMarkWatchedBatch (0.01s)
ok  	github.com/Silo-Server/silo-server/internal/userdb	0.096s
--- PASS: TestManualMarkWatchedSeriesResolvesIdentitiesInBulk (0.01s)
--- PASS: TestManualMarkWatchedSeriesIsAtomicOnCancel (0.00s)
ok  	github.com/Silo-Server/silo-server/internal/watchstate	0.158s
```

Touched packages, with a live database:

```
ok  	github.com/Silo-Server/silo-server/internal/watchstate
ok  	github.com/Silo-Server/silo-server/internal/userdb
ok  	github.com/Silo-Server/silo-server/internal/userstore
ok  	github.com/Silo-Server/silo-server/internal/userstore/pgstore	0.919s
ok  	github.com/Silo-Server/silo-server/internal/api/handlers	86.738s
```

Gates:

```
$ golangci-lint run --new-from-merge-base=origin/main ./...
0 issues.

$ cd web && pnpm run lint
✖ 151 problems (0 errors, 151 warnings)      # all pre-existing, none in changed files

$ make verify-local-paths
scripts/check-local-path-leaks.sh            # passes

$ make test-web
Test Files  276 passed (276)
     Tests  1927 passed (1927)
```

## Not green, and not mine

`go test ./internal/...` has **9 pre-existing failures** across `auth`, `catalog` search, `jellycompat` playback, and `notifications`. I verified these are byte-identical on a clean stash of this branch, so they are not from this change — and they are the same rot documented in #599. I did not fix them.

Without a test database the count is 4 (`jellycompat`), also identical on a clean tree.

## Not verified

I have **not** run this against a live server in a browser. The end-to-end check — mark a long series, close the tab within a second, reload, confirm every episode is watched — is the exact user-reported scenario and is still worth doing before merge. No screenshots for that reason; the UI itself is unchanged apart from the button responding immediately.

## Risks

- The two new batch SQL statements are the substantive risk; their hidden-watermark CTEs must match the single-row versions they mirror, which is what the shared conformance case pins.
- jellycompat's stored `duration_seconds` now carries a real value where the batch path previously wrote nothing. That is a fix, but it is a behaviour change in jellycompat.
- A very large series now holds a transaction longer than the previous drip of independent commits. These are per-user progress/history rows so contention should be negligible, but it is worth a look on the largest available series.

## Scope

No linked capability epic — this is a bug fix to existing behaviour, not new scope. Happy to file it under one if you would rather it were tracked there. No API contract change: same endpoints, same response shape, additive-only.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Adversarial review: My own review of the diff found three real defects, all fixed before this PR. (1) In the Postgres batch, entries with a blank media ID are dropped, so `entries[i]` could misalign with the returned rows — replaced positional indexing with an explicit index map. (2) The client `onMutate` cancelled queries on only one of the two item-detail key shapes while snapshotting both, so an in-flight query on the other shape could land after the optimistic write and revert the button — both now use the same predicate. (3) Routing jellycompat through the shared path would have zeroed known durations, as described above. Separately, `tsc -b` caught four type errors in my new test file that vitest alone did not surface, and `golangci-lint` caught 8 findings on my lines (3 errcheck, 5 misspell) — all fixed, lint is now clean at 0 issues. What I could not close: no live browser verification, and the 9 pre-existing suite failures remain.

## Summary by CodeRabbit

- **New Features**
  - Marking multiple episodes as watched now updates progress and watch history together.
  - Series watch actions resolve episode identities in bulk for more consistent results.
  - Watch durations are selected more reliably from media details, with catalog runtime fallback.

- **Bug Fixes**
  - Watched-state updates now appear immediately across item and catalog views.
  - Failed updates restore the previous state and display an error.
  - Watch actions remain reliable when navigating away during a request.

### DIFF
--- a/internal/api/handlers/items.go
+++ b/internal/api/handlers/items.go
@@ -2264,15 +2264,40 @@ func (h *ItemsHandler) resolveWatchedTargets(ctx context.Context, contentID stri
 	}}, nil
 }
 
+// episodeTargets resolves the watched leaf targets for a season or series.
+// Durations come from one batched file lookup rather than a query per episode:
+// marking a long series used to issue one GetByEpisodeID per episode before it
+// wrote anything.
 func (h *ItemsHandler) episodeTargets(ctx context.Context, episodes []*models.Episode) []watchedLeafTarget {
+	episodeIDs := make([]string, 0, len(episodes))
+	for _, episode := range episodes {
+		episodeIDs = append(episodeIDs, episode.ContentID)
+	}
+	filesByEpisode := h.listEpisodeFiles(ctx, episodeIDs)
+
 	targets := make([]watchedLeafTarget, 0, len(episodes))
 	for _, episode := range episodes {
 		targets = append(targets, watchedLeafTarget{
 			ContentID:       episode.ContentID,
-			DurationSeconds: h.contentDurationSeconds(ctx, "", episode.ContentID, episode.Runtime),
+			DurationSeconds: mediaFileDurationSeconds(filesByEpisode[episode.ContentID], episode.Runtime),
 		})
 	}
 	return targets
+}
+
+// mediaFileDurationSeconds picks a playable duration from an item's files,
+// falling back to the catalog runtime. Mirrors contentDurationSeconds's
+// preference order so the batched and single-item paths agree.
+func mediaFileDurationSeconds(files []*models.MediaFile, fallbackRuntimeMinutes int) float64 {
+	for _, file := range files {
+		if file != nil && file.Duration > 0 {
+			return float64(file.Duration)
+		}
+	}
+	if fallbackRuntimeMinutes > 0 {
+		return float64(fallbackRuntimeMinutes * 60)
+	}
+	return 0
 }
 
 func (h *ItemsHandler) contentDurationSeconds(ctx context.Context, contentID, episodeID string, fallbackRuntimeMinutes int) float64 {
@@ -2294,18 +2319,10 @@ func (h *ItemsHandler) contentDurationSeconds(ctx context.Context, contentID, ep
 	case contentID != "":
 		files, err = h.fileRepo.GetByContentID(ctx, contentID)
 	}
-	if err == nil {
-		for _, file := range files {
-			if file != nil && file.Duration > 0 {
-				return float64(file.Duration)
-			}
-		}
+	if err != nil {
+		files = nil
 	}
-
-	if fallbackRuntimeMinutes > 0 {
-		return float64(fallbackRuntimeMinutes * 60)
-	}
-	return 0
+	return mediaFileDurationSeconds(files, fallbackRuntimeMinutes)
 }
 
 // isNotFound checks if an error is a "not found" sentinel.
--- a/internal/userdb/conformance_test.go
+++ b/internal/userdb/conformance_test.go
@@ -27,6 +27,14 @@ func TestSQLiteProgressSince(t *testing.T) {
 	storetest.RunProgressSince(t, newConformanceStore)
 }
 
+// TestSQLiteMarkWatchedBatch runs the batch mark-watched conformance test
+// (series/season mark-watched) against the real SQLite backend. The Postgres
+// backend runs the same suite in internal/userstore/pgstore, which is what
+// keeps the two transactional implementations from drifting.
+func TestSQLiteMarkWatchedBatch(t *testing.T) {
+	storetest.RunMarkWatchedBatch(t, newConformanceStore)
+}
+
 // TestSQLiteSettingValues runs the canonical settings-contract storage
 // conformance tests against the per-user SQLite backend. The Postgres backend
 // runs the same suite in internal/userstore/pgstore, which is what keeps the two
--- a/internal/userdb/progress.go
+++ b/internal/userdb/progress.go
@@ -1,6 +1,7 @@
 package userdb
 
 import (
+	"context"
 	"database/sql"
 	"encoding/json"
 	"fmt"
@@ -171,14 +172,9 @@ func SetProgressIfNewer(db *sql.DB, profileID, mediaItemID string, position, dur
 	return rows > 0, nil
 }
 
-// MarkWatched creates or replaces progress with a completed entry.
-func MarkWatched(db *sql.DB, profileID, mediaItemID string, duration float64) error {
-	if duration < 0 {
-		duration = 0
-	}
-
-	now := nowUTC()
-	query := `
+// markWatchedProgressSQL upserts one completed progress row, honoring the
+// hidden-history watermark.
+const markWatchedProgressSQL = `
 		INSERT INTO watch_progress (profile_id, media_item_id, position_seconds, duration_seconds, completed, updated_at)
 		SELECT ?, ?, 0, ?, 1, ` + visibleTimestampSQL + `
 		FROM (SELECT 1) seed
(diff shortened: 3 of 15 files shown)


Quick104:
## Live verification on the dev server

Deployed this branch to the dev server (`make deploy-dev`, local source build) and tested in a real browser as `admin`. This closes the "not verified" gap in the PR description.

**Confirmed the running container had the new code** before testing — image changed `9bb3537124eb` → `a7ae1568ea5b`, and the new error strings are present in the deployed binary:

```
$ docker exec silo-silo-1 grep -ac "begin mark watched batch" /usr/local/bin/silo
1
$ docker exec silo-silo-1 grep -ac "marking watched batch" /usr/local/bin/silo
1
```

**Test subject: EastEnders — 42 seasons, 6,375 local episodes.** Far larger than the 200-episode case the PR describes, so the old code path would have been ~25,000 sequential queries.

### 1. Baseline mark — instant UI, one request

Verified via a `fetch` wrapper that the real request carries the flag:

```
POST keepalive=true status=200 ms=5929
```

6,375 episodes in **5.9s**, and the DB confirms a complete write:

```
 progress_completed | history_rows
--------------------+--------------
               6375 |         6375
```

### 2. Abort mid-flight (the reported bug)

Reset to 0/0, then aborted the in-flight request **1 second into a ~6 second operation**:

```
{"keepalive":true,"abortedAtMs":1000,"error":"AbortError"}
```

The client gave up, but the server still finished:

```
POST /api/v1/watched/series-tvdb-71753 status=200 duration_ms=13235
```

Result — **all 6,375, not a partial subset**:

```
 progress_completed | history_rows
--------------------+--------------
               6375 |         6375
```

### 3. Navigate away mid-flight (closest to the user report)

Reset to 0/0, clicked **Mark Series Watched**, then hard-navigated away after **800 ms**, destroying the document:

```
POST /api/v1/watched/series-tvdb-71753 status=200 duration_ms=13854
```

```
 progress_completed | history_rows
--------------------+--------------
               6375 |         6375
```

All 6,375 episodes marked despite leaving the page. Reloading the series shows **"Mark Series Unwatched"**, **"Play Latest"**, and a green watched check on every one of the 42 season tiles.

EastEnders fully marked watched after navigating away mid-request

*(Screenshot: EastEnders detail page — 42 SEASONS / 6375 EPISODES, "Mark Series Unwatched", every season tile check-marked.)*

### 4. Unmark

`DELETE` also sends `keepalive: true` and correctly cleared everything back to 0/0.

## One pre-existing issue this surfaced (not from this PR)

**Unmarking a 6,375-episode series took 96 seconds** (`DELETE ... status=200 duration_ms=96374`), versus ~6s to mark.

This is **not** a regression from this branch — `recordMarkUnwatched` is untouched by the diff, and `git blame` dates its paging loop to May/June. The cause looks like `completedHistoryForTargets` paging `ListCompletedHistory` 500 rows at a time across 6,375 IDs before `RemoveHistoryItems` runs — ~13 round-trips each passing the full ID array.

It is out of scope here (this PR is about mark-watched atomicity, one concern per PR), but it's the mirror image of the problem this PR fixes and probably deserves its own issue. Happy to file one.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Adversarial review: All evidence above is real captured output from the dev deployment — no synthesized logs. Deployment scope was deliberately narrowed to `make deploy-dev` (server only) rather than the full `make dev-deploy`, since the ns12/ns17 transcode-node redeploy is unrelated to this change. Test data was reset between each scenario and fully cleaned up afterward (EastEnders back to 0 progress / 0 history).

PR #647, opened 2026-08-15
fix(notifications): restore store capabilities and batch series resolution

### LINKED ISSUE #646: [bug] Unmarking a large series takes ~2 minutes (~16M index scans on `episodes` per request)
### What happened

Unmarking a large series (`DELETE /api/v1/watched/{id}`) takes **~2 minutes** on a 6,375-episode series, versus ~5s to mark the same series watched. The request holds a connection the whole time and the UI button stays pending.

Measured on the dev server against EastEnders (42 seasons, 6,375 file-backed episodes, 8,421 history rows), same series, back to back:

```
--- POST (mark)   --- http=200 total=5.152s
--- DELETE (unmark) --- http=000 total=143.622s   (client timeout at 300s not reached; curl gave up on the response)
```

Repeated runs of the DELETE: **96.4s**, **116.3s**, **143.6s**. Marking the same series is consistently 5–6s.

> Note on provenance: I found this while verifying #645 (which fixes the *mark* path). The unmark path is untouched by that PR — `recordMarkUnwatched` does not appear in its diff, and `git blame` dates `completedHistoryForTargets` to 2026-05-22/06-18. This is pre-existing, not a regression from #645.

### Root cause is NOT the obvious one

My first guess was that `completedHistoryForTargets` (`internal/watchstate/service.go`) is slow because it pages `ListCompletedHistory` 500 rows at a time — ~17 round-trips for 8,421 rows, each re-passing the full ID array. **That guess is wrong**, and I want to save the next person the detour.

Every individual SQL statement on this path is fast. Measured with `EXPLAIN (ANALYZE, BUFFERS)` against the real dev data:

- one `ListCompletedHistory` page (offset 0, 500 rows, full 7,398-ID array): **52 ms**
- the `user_history_hidden_items` watermark upsert in `RemoveHistoryItems`: **177 ms** (incl. 37 ms FK trigger × 7,398)
- the `user_watch_progress` delete: **33 ms**

17 pages × 52 ms is about 0.9s. That does not explain 143s.

### The actual signal: ~16 million index scans on `episodes` for one request

Diffing `pg_stat_user_tables.idx_scan` immediately before and after a single DELETE:

```
episodes:                  +15,989,221
user_history_hidden_items:    +352,236
media_files:                   +25,605
user_watch_history:             +8,046
user_watch_progress:            +1,705
media_item_provider_ids:          +348
```

**~16M index scans on `episodes` for ~8.4k history rows is ~1,900 scans per row.** Statement logging during the request confirms the shape: 2,343 logged statements, with the slowest at 907 ms and almost everything else in the 1–10 ms range. This is a Go-side N+1 (or N×M) fan-out, not one slow query — which is why `EXPLAIN` on any single statement looks healthy and the endpoint still takes two minutes.

The same measurement during a **mark** on the identical series shows `episodes: +2,584,217` — so both paths fan out, the unmark roughly 6× worse. Whatever is doing the per-item episode lookup is shared, and #645 only removed it from the mark path's identity resolution.

### Why this matters

- It is the mirror image of the bug #645 fixes. That PR made mark-watched atomic and unload-proof; unmark is still slow enough that a user will navigate away mid-request. Unmark *is* transactional (`RemoveHistoryItems` wraps its statements), so the failure mode is a hung UI rather than partial data — but a 2-minute button is its own bug.
- A ~16M-scan request is a real load concern on a shared server, not just a latency annoyance.

### Reproduction

Commands assume the repository root is the cwd, and a Silo deployment with a large series.

1. Pick a series with thousands of file-backed episodes:
   ```sql
   SELECT e.series_id, mi.title, COUNT(*) AS eps
   FROM episodes e JOIN media_items mi ON mi.content_id = e.series_id
   WHERE EXISTS (SELECT 1 FROM media_files f WHERE f.episode_id = e.content_id)
   GROUP BY e.series_id, mi.title ORDER BY eps DESC LIMIT 5;
   ```
2. Mark it watched, then time the unmark:
   ```bash
   curl -s -o /dev/null -w "%{http_code} %{time_total}s\n" -X POST   -H "Authorization: Bearer $TOKEN" -H "X-Profile-Id: $PROFILE" "$BASE/api/v1/watched/$SERIES_ID"
   curl -s -o /dev/null -w "%{http_code} %{time_total}s\n" -X DELETE -H "Authorization: Bearer $TOKEN" -H "X-Profile-Id: $PROFILE" "$BASE/api/v1/watched/$SERIES_ID"
   ```
3. To see the fan-out, snapshot `SELECT relname, idx_scan FROM pg_stat_user_tables WHERE relname='episodes'` immediately before and after the DELETE.

### Suggested direction (not yet validated)

Find the per-episode lookup behind the `episodes` fan-out and batch it, the way #645 batched `ResolveHistoryIdentities` and the duration lookups for the mark path. Likely candidates worth instrumenting first, in order:

1. `resolveWatchedTargets` / `episodeTargets` in `internal/api/handlers/items.go` — the unmark branch calls the same target expansion as mark.
2. Whatever populates `ManualMarkResult.Entries` for the local-watch-provider dispatch on unmark.
3. `internal/notifications/interest_hooks.go` and any completion observers reachable from the unmark request.

Worth confirming with a CPU profile or targeted instrumentation rather than by inspection — my first read of this code produced a wrong answer, and the SQL-level evidence above contradicts the intuitive explanation.

### Scope

- **Type:** bug, performance → `fix/` branch prefix.
- **Client surface:** none. No API contract change intended — same endpoint, same response shape.
- **Related:** #645 (fixes the mark-watched half). Not a v1 capability epic; happy to file it under one if preferred.

### Environment

- Dev server, this branch's build (`silo:dev-local`, image `a7ae1568ea5b`), Postgres 18 (pgvector image).
- Test data was seeded and fully cleaned up afterward; the series is back to 0 progress / 0 history rows.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Adversarial review: I filed this only after disproving my own initial hypothesis. My first explanation (the 500-row paging loop in `completedHistoryForTargets`) is stated above and refuted with `EXPLAIN (ANALYZE)` timings — the whole loop accounts for under a second of a 143-second request. Every number here is real captured output from the dev deployment; no synthesized logs or unexecuted repros. What I have *not* done is isolate the exact call site of the `episodes` fan-out, so the "suggested direction" is explicitly labelled unvalidated rather than presented as the fix.

Quick104:
## Problem

Fixes #646. Marking or unmarking a large series spent minutes inside the interest-tracking layer. Investigating that turned up **two independent defects**, both in `internal/notifications`, and the first one is the more serious of the two.

### 1. The decorator silently dropped every optional store capability

`interestTrackingStore` embeds the **interface** `userstore.UserStore`:

```go
type interestTrackingStore struct {
	userstore.UserStore
	...
}
```

Embedding an interface promotes only the methods *in that interface*. Any optional capability the backing store implements is invisible through the decorator. Minimal repro:

```go
type Base interface{ A() string }
type Extra interface{ B() string }
type real struct{}
func (real) A() string { return "a" }
func (real) B() string { return "b" }
type deco struct{ Base }

var d Base = deco{Base: real{}}
_, ok := d.(Extra)   // false
```

`cmd/silo/main.go:1680` wraps the provider unconditionally when notifications are enabled. So in production `userstore.MarkWatchedBatch`'s `store.(WatchedBatchWriter)` assertion **failed and took the per-target fallback** — the exact path #645 was written to remove. Same for `AddVisibleHistory`, `VisibleHistoryTimestamps`, and jellycompat's `SeriesEpisodeWatchCounts` rollup.

This is the nastiest property of the optional-capability pattern: callers all have a correct fallback, so the failure mode is silent. No error, no failing test — just a much slower server. #645's tests passed because they exercise the bare store; only the wired-up binary hits the decorator.

**Fix:** forward all four capabilities explicitly, plus compile-time assertions on both decorator types so a future capability is a build error instead of a silent regression:

```go
var _ userstore.WatchedBatchWriter = (*interestTrackingStore)(nil)
var _ userstore.VisibleHistoryAdder = (*interestTrackingStore)(nil)
var _ userstore.HistoryVisibilityStore = (*interestTrackingStore)(nil)
var _ userstore.SeriesEpisodeRollupStore = (*interestTrackingStore)(nil)
```

### 2. The interest flush resolved item→series one query at a time

`InterestUpdater.flush` called `resolveSeriesID` **per queued item**, then deduped to one recompute per series. A whole-series mark queues one mutation per episode, so thousands of lookups all collapsed to a single series — after paying for every one of them. The dedupe ran *after* the per-item query, not before.

**Fix:** resolve the whole batch in one query and dedupe from the result. `resolveSeriesID` had no other callers, so it is deleted rather than left as a second path that can drift from the batched one.

## Results

Measured on the dev server against EastEnders (6,375 file-backed episodes), three clean runs with the async flush allowed to drain between requests:

```
run1 MARK 0.905s   UNMARK 8.506s
run2 MARK 0.793s   UNMARK 9.762s
run3 MARK 1.042s   UNMARK 7.655s
```

| | before (#646) | after | |
|---|---|---|---|
| mark | 5.2s | **~0.9s** | ~5× |
| unmark | 96s / 116s / 143s | **~8.6s** | ~13× |
| `episodes` index scans | 18.5M | **19,298** | ~960× |

## A correction to my own measurement in #646

#646 reports unmark at 96–143s. Re-measuring carefully, **part of that was measurement error on my part**: I ran mark and unmark back to back, and the async interest flush kicked off by the *mark* was still saturating the pool when the unmark started. Spacing the requests shows the true post-fix figure is ~8.6s, and it also means the pre-fix numbers were inflated by the same effect.

The defects are real and the improvement is real — the fan-out and the dropped capabilities are both independently verified — but the headline "2 minutes" in the issue overstates the isolated unmark cost. I would rather correct that than let a flattering number stand.

## Testing

New tests:

- `TestInterestTrackingStorePreservesWatchStateCapabilities` — asserts every capability the bare store advertises survives wrapping, then drives a real `MarkWatchedBatch` through the wrapper to prove pass-through. **Fails on main** with all three capabilities reported dropped, which is how the bug was found.
- `TestResolveSeriesIDsBatchesLookups` — pins batched resolution semantics against Postgres (episodes, seasons, series-type items resolve; movies, unknown IDs, and blanks are absent so the flush loop still skips them). Uses temp tables with a single-connection pool, matching `push_devices_test.go`.

There was already a `TestInterestTrackingStorePreservesSettingCapabilities` guarding this exact class of bug for the settings capabilities — it just predates the watch-state ones. The new test extends that idea rather than inventing a pattern.

Gates:

```
$ golangci-lint run --new-from-merge-base=origin/main ./...
0 issues.

$ go test ./internal/notifications/ ./internal/watchstate/ ./internal/userstore/... ./internal/userdb/
ok  	internal/watchstate
ok  	internal/userstore
ok  	internal/userstore/pgstore
ok  	internal/userdb
```

Backend-only change; no frontend files touched, so `make test-web` is unaffected.

## Pre-existing failures (not from this PR)

`internal/notifications` has 3 failing tests (`TestDispatchOperationalEnqueuesApplePushAttempts`, `TestPushDeviceRepositoryUpsertApple*`) and `internal/jellycompat` has 2–4 depending on whether a test DB is present. I verified these are byte-identical on a clean stash of this branch — they are the rot documented in #599, not this change.

## Risks

- `SeriesEpisodeWatchCounts` on the decorator returns an error when the inner store lacks the capability (the per-user SQLite backend has no catalog tables). Every caller already treats an error as "use the per-episode fallback", so behaviour is unchanged there — but it is worth a reviewer's eye, since the wrapper now always *satisfies* the interface even when the inner store does not.
- The batched resolver keeps the first row when an ID matches more than one branch, mirroring the old `UNION ALL ... LIMIT 1`. If an ID could legitimately exist as both an episode and a season, ordering matters — I do not believe that is possible, but I have not proven it.
- Unmark at ~8.6s is much better, not instant. #645's `keepalive` covers the user-facing half; further work on `RemoveHistoryItems` would be a separate concern.

### AI Disclosure
- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Adversarial review: I disproved two of my own hypotheses before landing this. First, #646's suggested cause (the 500-row paging loop in `completedHistoryForTargets`) is wrong — `EXPLAIN (ANALYZE)` puts the whole loop under a second. Second, I initially attributed the remaining post-fix cost to `LIMIT/OFFSET` quadratic re-scanning; profiling page 1 vs page 13 showed 21ms vs 10ms, refuting that too, and led to the async-flush contention finding that corrects the issue's headline number. I also verified the interface-embedding behaviour with a standalone Go program rather than trusting my reading of the spec, and confirmed the "idle" background scan rate (449 scans/45s) to make sure the 16M figure was really attributable to the request. Not done: no browser testing (backend-only change, no UI surface).

Part of #646.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved watch-state tracking for batch updates and visible watch history.
  * Preserved series watch counts and progress updates during notification processing.
  * Improved retry handling when interest updates cannot be resolved.

* **Performance**
  * Batched series and episode lookups to reduce database queries.

* **Tests**
  * Added coverage for watch-state forwarding, completed progress records, batched resolution, duplicates, and unknown media.

### DIFF
--- a/internal/notifications/interest_hooks.go
+++ b/internal/notifications/interest_hooks.go
@@ -333,9 +333,19 @@ func (s *interestTrackingStore) MarkWatchedBatch(
 ) ([]userstore.WatchHistoryEntry, error) {
 	writer, ok := s.UserStore.(userstore.WatchedBatchWriter)
 	if !ok {
-		return userstore.MarkWatchedBatch(ctx, s.UserStore, profileID, targets, entries)
+		// The helper runs against the inner store, so this decorator's own
+		// MarkWatched hook never fires; queue here or the recompute is lost.
+		// A mid-loop error still leaves earlier targets written, so queue
+		// regardless of err — a redundant queue only costs one recompute,
+		// while a missing one leaves profile_series_interest stale until some
+		// unrelated mutation or the rebuild task touches the series.
+		written, err := userstore.MarkWatchedBatch(ctx, s.UserStore, profileID, targets, entries)
+		s.queueTargetMutations(profileID, targets)
+		return written, err
 	}
 	written, err := writer.MarkWatchedBatch(ctx, profileID, targets, entries)
+	// The batch write is one transaction: on error nothing landed, so there is
+	// nothing to recompute.
 	if err == nil {
 		s.queueItemMutations(profileID, written)
 	}
@@ -347,6 +357,14 @@ func (s *interestTrackingStore) MarkWatchedBatch(
 func (s *interestTrackingStore) queueItemMutations(profileID string, entries []userstore.WatchHistoryEntry) {
 	for _, entry := range entries {
 		s.updater.QueueItemMutation(s.userID, profileID, entry.MediaItemID)
+	}
+}
+
+// queueTargetMutations queues by requested target rather than written entry,
+// for the non-transactional path where a partial write may have occurred.
+func (s *interestTrackingStore) queueTargetMutations(profileID string, targets []userstore.MarkWatchedTarget) {
+	for _, target := range targets {
+		s.updater.QueueItemMutation(s.userID, profileID, target.MediaItemID)
 	}
 }
 
--- a/internal/notifications/interest_hooks.go
+++ b/internal/notifications/interest_hooks.go
@@ -70,6 +70,19 @@ var _ userstore.SettingMutationTransactioner = (*interestTrackingStore)(nil)
 var _ userstore.SettingValueCompareAndSetter = (*interestTrackingStoreWithDevices)(nil)
 var _ userstore.SettingMutationTransactioner = (*interestTrackingStoreWithDevices)(nil)
 
+// Optional store capabilities must survive the decorator. Embedding the
+// UserStore interface promotes only that interface's methods, so each of these
+// needs an explicit forward below; the assertions make a missing one a compile
+// error instead of a silent production slowdown.
+var _ userstore.WatchedBatchWriter = (*interestTrackingStore)(nil)
+var _ userstore.VisibleHistoryAdder = (*interestTrackingStore)(nil)
+var _ userstore.HistoryVisibilityStore = (*interestTrackingStore)(nil)
+var _ userstore.SeriesEpisodeRollupStore = (*interestTrackingStore)(nil)
+var _ userstore.WatchedBatchWriter = (*interestTrackingStoreWithDevices)(nil)
+var _ userstore.VisibleHistoryAdder = (*interestTrackingStoreWithDevices)(nil)
+var _ userstore.HistoryVisibilityStore = (*interestTrackingStoreWithDevices)(nil)
+var _ userstore.SeriesEpisodeRollupStore = (*interestTrackingStoreWithDevices)(nil)
+
 // WithPreferenceSettingsTransaction preserves the optional atomic-settings
 // capability of the wrapped store. Preference writes do not affect interest
 // signals, so the transaction can pass through unchanged; keeping the method
@@ -296,6 +309,80 @@ func (s *interestTrackingStore) RemoveHistoryItems(ctx context.Context, profileI
 		}
 	}
 	return err
+}
+
+// --- Optional store capabilities.
+//
+// This decorator embeds the userstore.UserStore *interface*, which promotes
+// only the methods declared on that interface. Any optional capability the
+// backing store implements is invisible through the wrapper unless it is
+// forwarded explicitly here — and because callers reach these via type
+// assertion with a working fallback, a missing forward is silent: no error, no
+// test failure, just a much slower path in production. Marking a series
+// watched regressed exactly this way. When adding a new optional capability to
+// userstore, forward it here and extend the compile-time assertions below.
+
+// MarkWatchedBatch forwards the transactional batch mark-watched write, whose
+// whole purpose is that a large series commits as one unit. Falling through to
+// the per-target loop would restore the partial-write window it removes.
+func (s *interestTrackingStore) MarkWatchedBatch(
+	ctx context.Context,
+	profileID string,
+	targets []userstore.MarkWatchedTarget,
+	entries []userstore.WatchHistoryEntry,
+) ([]userstore.WatchHistoryEntry, error) {
+	writer, ok := s.UserStore.(userstore.WatchedBatchWriter)
+	if !ok {
+		return userstore.MarkWatchedBatch(ctx, s.UserStore, profileID, targets, entries)
+	}
+	written, err := writer.MarkWatchedBatch(ctx, profileID, targets, entries)
+	if err == nil {
+		s.queueItemMutations(profileID, written)
+	}
+	return written, err
+}
+
+// queueItemMutations records one interest mutation per written entry, matching
+// what the per-target path queues through MarkWatched.
+func (s *interestTrackingStore) queueItemMutations(profileID string, entries []userstore.WatchHistoryEntry) {
+	for _, entry := range entries {
+		s.updater.QueueItemMutation(s.userID, profileID, entry.MediaItemID)
+	}
+}
+
+// AddVisibleHistory forwards the watermark-aware history insert. The generic
+// fallback needs two round-trips (timestamp lookup, then insert) to do the
+// same job.
+func (s *interestTrackingStore) AddVisibleHistory(ctx context.Context, entry userstore.WatchHistoryEntry) (userstore.WatchHistoryEntry, error) {
+	adder, ok := s.UserStore.(userstore.VisibleHistoryAdder)
+	if !ok {
+		return userstore.AddVisibleHistory(ctx, s.UserStore, entry)
+	}
+	return adder.AddVisibleHistory(ctx, entry)
+}
+
+// VisibleHistoryTimestamps forwards the batched watermark lookup; without it
+// callers assume a single wall-clock timestamp for every item.
+func (s *interestTrackingStore) VisibleHistoryTimestamps(ctx context.Context, profileID string, mediaItemIDs []string, at time.Time) (map[string]string, error) {
+	visibility, ok := s.UserStore.(userstore.HistoryVisibilityStore)
+	if !ok {
+		return userstore.VisibleHistoryTimestamps(ctx, s.UserStore, profileID, mediaItemIDs, at)
+	}
+	return visibility.VisibleHistoryTimestamps(ctx, profileID, mediaItemIDs, at)
+}
+
+// SeriesEpisodeWatchCounts forwards the SQL-side series rollup used by
+// jellycompat; without it that path materializes every episode of every
+// series. Stores that lack the capability (the per-user SQLite backend has no
+// catalog tables) return an error, which every caller already treats as
+// "use the per-episode fallback" — the same outcome as before this forward
+// existed.
+func (s *interestTrackingStore) SeriesEpisodeWatchCounts(ctx context.Context, profileID string, seriesIDs []string) (map[string]userstore.SeriesWatchCounts, error) {
+	rollup, ok := s.UserStore.(userstore.SeriesEpisodeRollupStore)
+	if !ok {
+		return nil, fmt.Errorf("series episode rollup is not supported by this store")
+	}
+	return rollup.SeriesEpisodeWatchCounts(ctx, profileID, seriesIDs)
 }
 
 func (s *interestTrackingStore) DeleteHistoryBySource(ctx context.Context, profileID string, mediaItemIDs []string, source userstore.WatchHistorySource) error {
--- a/internal/notifications/interest_hooks.go
+++ b/internal/notifications/interest_hooks.go
@@ -39,11 +39,27 @@ func (p *interestTrackingProvider) ForUser(ctx context.Context, userID int) (use
 		return store, err
 	}
 	tracked := &interestTrackingStore{UserStore: store, userID: userID, system: p.system, updater: p.system.Interest}
-	// Preserve the DeviceRegistry interface upgrade some callers probe for.
-	if registry, ok := store.(userstore.DeviceRegistry); ok {
+	// Preserve the interface upgrades callers probe for. Both are conditional
+	// on the backing store: advertising a capability it does not have would
+	// send callers down a fast path that can only fail.
+	registry, hasDevices := store.(userstore.DeviceRegistry)
+	rollup, hasRollup := store.(userstore.SeriesEpisodeRollupStore)
+	switch {
+	case hasDevices && hasRollup:
+		return &interestTrackingStoreWithDevicesAndRollup{
+			interestTrackingStore:    tracked,
+			DeviceRegistry:           registry,
+			SeriesEpisodeRollupStore: rollup,
+		}, nil
+	case hasDevices:
 		return &interestTrackingStoreWithDevices{
 			interestTrackingStore: tracked,
 			DeviceRegistry:        registry,
+		}, nil
+	case hasRollup:
+		return &interestTrackingStoreWithRollup{
+			interestTrackingStore:    tracked,
+			SeriesEpisodeRollupStore: rollup,
 		}, nil
 	}
 	return tracked, nil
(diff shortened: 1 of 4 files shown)


Quick104:
Good catch — this was a real bug, and I've fixed it in 7b3a9abe.

Confirmed the mechanism before fixing: the fallback calls `userstore.MarkWatchedBatch(ctx, s.UserStore, ...)` against the **inner** store, so the decorator's own `MarkWatched` hook never fires and nothing queues a recompute. Marking a series watched on such a backend would update progress and history but leave `profile_series_interest` stale until an unrelated mutation or the rebuild task happened to touch the series.

I took the substance of the suggestion, with one deliberate difference: rather than inlining the loop, I added `queueTargetMutations` alongside the existing `queueItemMutations`. The two paths queue from genuinely different sources and I wanted that visible at the call site:

- **fallback** — queue from `targets`, regardless of error. It's a per-target loop, so a mid-loop failure still leaves earlier targets written.
- **transactional** — queue from `written`, only on success. The whole batch is one transaction, so on error nothing landed and there is nothing to recompute.

Your point about queuing even on error is right for the fallback specifically, and that reasoning is now in a comment so it doesn't get "tidied up" later.

Also added the regression test you asked for — `TestInterestTrackingStoreQueuesMutationsOnBatchFallback`, using a `storeWithoutBatchWriter` wrapper that hides the capability from an otherwise-working SQLite store. It fails on the previous commit:

```
--- FAIL: TestInterestTrackingStoreQueuesMutationsOnBatchFallback
    no interest mutation queued for ep-1 on the batch fallback path; interest goes stale
    no interest mutation queued for ep-2 on the batch fallback path; interest goes stale
```

and passes now.

While in there I audited the other three capability forwards for the same gap. They're clean: `AddVisibleHistory` and `VisibleHistoryTimestamps` are pass-throughs, and history-only writes were never hooked for interest in the first place — `AddHistory` is deliberately unhooked with an existing comment explaining that playback always writes progress alongside it, and those writes queue on transitions.

Quick104:
@coderabbitai review

Quick104:
## Re-verified on the dev server after the review fix — plus a correction

Redeployed with 7b3a9abe and re-measured. The headline numbers hold, but I need to correct something I wrote earlier in this PR.

### Settled measurements (6,375-episode series)

```
run1 mark 0.942s   unmark 8.082s
run2 mark 0.987s   unmark 6.869s
run3 mark 0.811s   unmark 8.016s
```

`episodes` index scans for a mark + unmark pair, measured via `pg_stat_user_tables`: **780** (was ~18.5M).

Also checked whether the async interest flush interferes, by varying the gap between mark and unmark:

```
unmark after   5s idle: 8.500s
unmark after  45s idle: 7.235s
unmark after 120s idle: 6.984s
```

Essentially flat — so the flush is *not* meaningfully contending after this fix.

### Correction to my earlier comment

Immediately after redeploying I measured unmark at **74–76s** and attributed it to async-flush contention, saying a 20s gap was insufficient. **That explanation was wrong.** Re-running the identical back-to-back pattern later gave 7–8s every time.

The real cause was that those runs happened minutes after a container restart, while the post-deploy library scan was still running — 1,499 `library ingest` / `scan queue: progress` log lines in that window. I was benchmarking against a server doing startup work, not measuring a regression.

Two things follow, and the second matters more than the first:

1. The post-fix figure is **~7–8s**, consistent across every measurement taken on a settled server.
2. **The same caveat applies to the "before" numbers in #646** (96s / 116s / 143s). Those were also taken shortly after a deploy. The defects are real and independently verified — the dropped capabilities are proven by a test that fails on the parent commit, and the 18.5M → 780 scan reduction is measured directly — but I would treat the *magnitude* of the unmark improvement as "roughly an order of magnitude" rather than the precise 13× I quoted. The scan-count reduction is the number I trust; wall-clock on a shared dev box with a scanner running is not.

I have updated nothing in the code as a result of this; it is purely a correction to my own reported measurements.

Quick104:
Both Codex findings addressed.

**`interest_hooks.go:336` — interest hooks lost in the batch fallback.** Same issue CodeRabbit flagged; fixed in 7b3a9abe with a regression test that fails on the parent commit. Details in the comment above.

**`interest_hooks.go:401` — advertising unsupported rollups.** Valid, and fixed in 21fe2dd8. I had flagged this exact risk in the PR description ("the wrapper now always *satisfies* the interface even when the inner store does not") but talked myself out of it on the grounds that callers handle the error. That was the wrong call: the caller handling an error is not the same as the error being free. Every jellycompat series detail and browse on a SQLite-backed deployment would enter the fast path, fail, log `series watch rollup query failed`, and only then fall back — recurring warning noise on requests that actually succeeded, which is exactly the kind of thing that trains people to ignore logs.

The pointer to `DeviceRegistry` was the useful part: this codebase already had the right pattern for a conditional capability, and I should have followed it rather than inventing an error return. Now composed in `ForUser`:

```go
registry, hasDevices := store.(userstore.DeviceRegistry)
rollup, hasRollup := store.(userstore.SeriesEpisodeRollupStore)
switch {
case hasDevices && hasRollup: // interestTrackingStoreWithDevicesAndRollup
case hasDevices:              // interestTrackingStoreWithDevices
case hasRollup:               // interestTrackingStoreWithRollup
}
```

The other three capabilities stay unconditional, and that distinction is deliberate: `WatchedBatchWriter`, `VisibleHistoryAdder`, and `HistoryVisibilityStore` all have real generic fallbacks that any store can execute, so forwarding them is always correct. The rollup has no generic fallback at the store layer — the fallback lives in the caller — which is why it alone has to be conditional. That reasoning is now a comment on the type.

Tests cover both directions:

- `TestInterestTrackingStorePreservesWatchStateCapabilities` now also asserts the SQLite-backed wrapper does **not** advertise the rollup.
- `TestInterestTrackingStoreForwardsRollupWhenSupported` (new) uses a rollup-capable store and checks the capability survives wrapping, the call reaches the backing store, and the other capabilities still compose alongside it.

```
--- PASS: TestInterestTrackingStorePreservesSettingCapabilities
--- PASS: TestInterestTrackingStorePreservesWatchStateCapabilities
--- PASS: TestInterestTrackingStoreQueuesMutationsOnBatchFallback
--- PASS: TestInterestTrackingStoreForwardsRollupWhenSupported
```

`golangci-lint run --new-from-merge-base=origin/main ./...` → 0 issues.

Quick104:
## Final verification on 21fe2dd8 — and a second correction to my own numbers

Redeployed with the rollup fix and re-measured. **Ten consecutive mark/unmark pairs** against the 6,375-episode series on a settled server:

```
   1 mark=0.993s unmark=8.475s      6 mark=1.038s unmark=7.755s
   2 mark=0.810s unmark=7.751s      7 mark=0.930s unmark=7.819s
   3 mark=1.016s unmark=9.695s      8 mark=0.935s unmark=7.994s
   4 mark=0.947s unmark=9.195s      9 mark=1.019s unmark=8.934s
   5 mark=1.044s unmark=8.210s     10 mark=0.981s unmark=8.968s
```

**mark 0.81–1.04s, unmark 7.75–9.69s.** That is the settled figure, from the largest sample I have taken.

### I got the *explanation* wrong twice; here is what actually happened

In my previous comment I said the 74–95s readings were caused by the async interest flush contending, and that a 20s gap was insufficient. Then I measured again and saw 7–8s with the same 20s gap, so I revised it to "post-deploy library scan". **That second explanation was also wrong**, and I want to correct it rather than leave a tidy-sounding story that does not hold:

- I later saw 74–95s on a server with **zero** scan-log lines in the preceding minute.
- A deliberate `docker restart` at T+45s gave **7.1s** — so "container recently started" is not the trigger either.
- Postgres was never restarted through any of this (56 days of stats uptime), so it is app-side.

What I can state with confidence:

1. The settled, reproducible numbers are the ten runs above, and they are stable across a 60s gap, no gap, and repeated iterations.
2. The slow readings only ever appeared in the minutes following a **deploy** (image rebuild + recreate), never in steady state, and I could not reproduce them on demand afterwards.
3. I do **not** have a confirmed root cause for that post-deploy window. Calling it "the scanner" was a guess that fit the first data point and failed the second.

### What this means for the numbers in #646 and in this PR

The **scan-count reduction is the measurement I trust**: `episodes` index scans for a mark+unmark pair went from ~18.5M to a few hundred, measured directly from `pg_stat_user_tables` deltas. That is a structural change, not a timing artifact, and it is corroborated by the code change itself (N queries → 1).

The **wall-clock numbers deserve a wider error bar than I originally gave them.** Both the "before" figures in #646 (96/116/143s) and my first "after" figures were taken in exactly the post-deploy window I now know to be unreliable. The honest summary is:

- mark: seconds → **~1s**, consistently
- unmark: tens of seconds → **~8s** settled, with post-deploy outliers I cannot yet explain
- `episodes` scans: **~18.5M → hundreds**

I would rather hand a reviewer that than a clean-looking "13× faster" I can no longer fully stand behind.

### State of the PR

- All four CI checks green on 21fe2dd8.
- CodeRabbit's finding: fixed (7b3a9abe), thread resolved, regression test fails on the parent commit.
- Both Codex findings: fixed (7b3a9abe, 21fe2dd8), threads now outdated.
- `golangci-lint --new-from-merge-base=origin/main`: 0 issues.
- Four tests in `internal/notifications` covering capability pass-through, the fallback queue, and both directions of the conditional rollup.

Pre-existing failures unrelated to this PR (verified identical on a clean stash): 3 in `internal/notifications` push/dispatch and 2–4 in `internal/jellycompat`, all per #599.

Issue #648, opened 2026-08-15
[bug] Safari rejects chunked progressive remux responses without byte-range support


tomislav:
### What happened

After deploying the merged fix from #617, Safari capability negotiation succeeds and Silo selects a Dolby Vision `server_remux_progressive` plan. The stream still fails immediately in Safari, followed by:

> This file's dynamic range can't be converted for this device, and it can't be played as-is.

The generated media itself is valid. A controlled test using the exact same remuxed MP4 bytes isolates the failure to the HTTP transport.

### Environment

- Safari 26.6
- macOS 26.6.1
- Apple Silicon MacBook Pro
- Built-in Liquid Retina XDR display
- Web UI
- Docker deployment running the #617 merge

### Source characteristics

Represented generically here as `video.mkv`:

- HEVC Main 10
- Dolby Vision Profile 8.1, Level 6
- Base-layer compatibility ID 1
- BT.2020/PQ
- E-AC-3 5.1
- Approximately 2160p24
- MKV requiring container normalization

### Server behavior

Planning now succeeds:

```text
outcome=plan
decision_reason=container_normalization
delivery=server_remux_progressive
play_method=remux
dynamic_range=dolby_vision
dv_profile=8
```

Safari requests the stream, receives HTTP 200, then reports `decoder_error` within approximately 200 ms.

The equivalent FFmpeg remux was captured and verified to contain:

- `dvh1`
- `hvcC`
- `dvvC` / DOVI configuration record
- DV Profile 8, Level 6
- BL compatibility ID 1
- RPU present, no enhancement layer
- HEVC Main 10 with BT.2020/PQ
- E-AC-3 audio

Both the full remux and a video-only remux play when opened as files, ruling out the DV bitstream and audio track.

### Controlled HTTP transport test

The same MP4 bytes were served to the same Safari instance in two ways:

| HTTP delivery | Result |
|---|---|
| Known `Content-Length`, `Accept-Ranges: bytes`, and correct `206 Partial Content` responses | Plays |
| Unbounded `200 OK`, `Transfer-Encoding: chunked`, no byte-range support | Fails |

The second case matches the current progressive-remux implementation in `ServeRemuxWithOptions`, which streams FFmpeg stdout as an unknown-length chunked response.

The same DV P8.1 source packaged as segmented fMP4 HLS also plays successfully in Safari.

[...]

PR #657, opened 2026-08-16
fix(playback): isolate the DV7 HDR10 base layer

### LINKED ISSUE #656: [bug] DV7 HDR10 fallback retains enhancement-layer NAL units
## What happened

On a real Docker deployment, the `server_dv7_to_hdr10` remux recipe produced an HDR10-signalled HEVC stream that still contained Dolby Vision Profile 7 enhancement-layer NAL units.

The current recipe uses `dovi_rpu=strip=1`. That removes the Dolby Vision configuration and RPU metadata, but Profile 7 enhancement-layer units are interleaved in the same video stream as NAL type 63, so selecting `0:V:0` does not remove them. This contradicts the transformation's `enhancement_layer_discarded` claim and can hand an HDR10-only client bytes it did not declare support for.

This is separate from #653: that PR changes how macOS Safari receives an HDR remux. This report is about the byte-level correctness of the DV7-to-HDR10 output for every remux delivery.

## Steps to reproduce

I reproduced this with a dual-layer Profile 7 source on Jellyfin FFmpeg 7.1.4. The media path and title are redacted; no other command output was changed.

1. Confirm that the source is HEVC Main 10, PQ/BT.2020, Dolby Vision Profile 7, with an enhancement layer.
2. Copy two seconds through the current `dovi_rpu=strip=1` recipe and append `trace_headers`. Count type-63 units emitted after `dovi_rpu`: 51.
3. Repeat with `dovi_rpu=strip=1,filter_units=remove_types=63`. Count type-63 units emitted after the complete chain: 0.
4. Probe the corrected output in memory. It remains HEVC Main 10 with BT.2020 primaries, BT.2020 non-constant-luminance matrix, and SMPTE ST 2084 transfer; it no longer contains a Dolby Vision configuration record.

Commands used:

```shell
# <profile7-source> is an explicit redaction of the private media path.
ffprobe -v error -select_streams v:0 \
  -show_entries stream=codec_name,profile,pix_fmt,color_space,color_transfer,color_primaries:stream_side_data \
  -of json <profile7-source>

ffmpeg -nostdin -hide_banner -loglevel trace -t 2 -i <profile7-source> \
  -map 0:V:0 -an -sn -dn -c:v copy \
  -bsf:v dovi_rpu=strip=1,trace_headers -f null - 2>&1 \
  | awk '/trace_headers/ && /nal_unit_type: 63/ {count++} END {print count+0}'

ffmpeg -nostdin -hide_banner -loglevel trace -t 2 -i <profile7-source> \
  -map 0:V:0 -an -sn -dn -c:v copy \
  -bsf:v dovi_rpu=strip=1,filter_units=remove_types=63,trace_headers -f null - 2>&1 \
  | awk '/trace_headers/ && /nal_unit_type: 63/ {count++} END {print count+0}'

ffmpeg -nostdin -hide_banner -loglevel error -t 2 -i <profile7-source> \
  -map 0:V:0 -an -sn -dn -c:v copy \
  -bsf:v dovi_rpu=strip=1,filter_units=remove_types=63 -f hevc - \
  | ffprobe -v error -f hevc -i - \
      -show_entries stream=codec_name,profile,pix_fmt,color_space,color_transfer,color_primaries:stream_side_data \
      -of json
```

## Expected behavior

When Silo advertises and selects `server_dv7_to_hdr10`, the output should contain a clean HDR10 base layer: no Dolby Vision configuration/RPUs and no interleaved type-63 enhancement-layer units.

The transformation should be available only when FFmpeg provides both `dovi_rpu` and `filter_units`. Because this changes the bytes promised by the recipe, its recipe version should also change so an older pooled node cannot claim compatibility.

## Silo version / commit

- Upstream baseline containing the incomplete recipe: `edd919c5`
- Real Docker deployment used for the A/B media probe: `815a2bd86e53216e81e18ce4ceaa024bfc82e21b`
- FFmpeg: `7.1.4-Jellyfin`

## Deployment

Docker

## Clients affected

- [x] Web UI
- [x] iOS/tvOS/macOS
- [x] Other/not client-specific

The invalid fallback is generated server-side, so the defect is not limited to one client. It is directly relevant to macOS Safari because Safari cannot decode Profile 7 and relies on the HDR10 fallback.

## Relevant logs

```shell
ffmpeg version 7.1.4-Jellyfin Copyright (c) 2000-2026 the FFmpeg developers

dovi_rpu
filter_units

{
    "programs": [],
    "stream_groups": [],
    "streams": [
        {
            "codec_name": "hevc",
            "profile": "Main 10",
            "pix_fmt": "yuv420p10le",
            "color_space": "bt2020nc",
            "color_transfer": "smpte2084",
            "color_primaries": "bt2020",
            "side_data_list": [
                {
                    "side_data_type": "DOVI configuration record",
                    "dv_version_major": 1,
                    "dv_version_minor": 0,
                    "dv_profile": 7,
                    "dv_level": 6,
                    "rpu_present_flag": 1,
                    "el_present_flag": 1,
                    "bl_present_flag": 1,
                    "dv_bl_signal_compatibility_id": 6,
                    "dv_md_compression": "none"
                }
            ]
        }
    ]
}

[trace_headers @ 0x60759597db80] nal_unit_type: 63(UNSPEC63), nuh_layer_id: 0, temporal_id: 0
[trace_headers @ 0x60759597db80] nal_unit_type: 63(UNSPEC63), nuh_layer_id: 0, temporal_id: 0

# type-63 units after the current dovi_rpu-only recipe, two-second sample
51

# type-63 units after dovi_rpu + filter_units, same source and interval
0

{
    "programs": [],
    "stream_groups": [],
    "streams": [
        {
            "codec_name": "hevc",
            "profile": "Main 10",
            "pix_fmt": "yuv420p10le",
            "color_space": "bt2020nc",
            "color_transfer": "smpte2084",
            "color_primaries": "bt2020"
        }
    ]
}
```

## AI involvement in this report

AI-assisted, human verified the repro

## AI tool + model

OpenAI Codex / gpt-5

## Confirmations

- [x] I reproduced this on a real deployment.
- [x] The command output above is raw copy-paste, not AI-summarized. The private media path and title are explicitly redacted; two repeated `trace_headers` lines are shown before the complete count.

blurbery:
## Problem
Fixes #656

Related: #653 (Safari transport for #648, submitted by a different contributor)

The `server_dv7_to_hdr10` transformation promised an HDR10 base layer with the Dolby Vision enhancement layer discarded, but its FFmpeg recipe only ran `dovi_rpu=strip=1`. Profile 7 enhancement-layer NAL units are interleaved in the selected HEVC stream as type 63, so stream mapping did not remove them.

That left HDR10-signalled output containing bytes the fallback client did not declare support for. This is directly relevant to macOS Safari, which cannot decode Profile 7 and relies on the server fallback, but the malformed output was server-side and affected every remux delivery.

PR #653 addresses how Safari receives HDR remuxes. This PR is deliberately separate: it makes the DV7-to-HDR10 bytes correct regardless of transport.

## Approach
- Apply `dovi_rpu=strip=1,filter_units=remove_types=63` to both progressive and HLS copy-video remuxes.
- Advertise the transformation only when FFmpeg provides both required bitstream filters, with a bounded capability probe.
- Bump `server_dv7_to_hdr10` from recipe v1 to v2 and require an exact server/name/version match.
- Carry the recipe version through progressive session state, reconstruction cards, signed stream claims, and proxy delivery. Stale or missing v1 tokens fail closed instead of silently executing different bytes.
- Regenerate the protocol fixtures and cover Profile 7, the existing compatible Profile 8 fallback, mixed-version nodes, local remux, HLS, reconstruction, and proxy paths.

I kept Safari native-HLS selection, sample-entry propagation, broader Dolby Vision routing, and Firefox behavior out of this PR because they are separate concerns.

## Testing
The raw real-media A/B probe is in #656. On the same two-second Profile 7 interval:

```text
current dovi_rpu-only recipe: 51 type-63 NAL units
corrected complete chain:       0 type-63 NAL units
```

The corrected output remained HEVC Main 10, 10-bit PQ/BT.2020 HDR and no longer exposed a Dolby Vision configuration record.

Actual local command output:

```text
$ go test ./internal/playback ./internal/playback/contract -skip 'NVENC|HWAccel' -count=1
ok  github.com/Silo-Server/silo-server/internal/playback           14.606s
ok  github.com/Silo-Server/silo-server/internal/playback/contract   0.760s

$ go test ./internal/proxy -count=1
ok  github.com/Silo-Server/silo-server/internal/proxy  0.341s

$ golangci-lint run --new-from-merge-base=upstream/main ./internal/playback/...
0 issues.

$ make verify-playback-fixtures
playback fixtures are current

$ make verify-local-paths
scripts/check-local-path-leaks.sh

$ git diff upstream/main...HEAD --check
<no output>
```

I also ran the final commit through the repository's exact [CI workflow on the fork](https://github.com/blurbery/silo-server/actions/runs/31936503495). All three jobs passed:

```text
Docs hygiene  pass
Web           pass  (lint, format, typecheck/build, tests, generated binding)
Go            pass  (build, gofmt, vet, changed-line lint, fixtures, full tests)
```

This is a server-only byte-stream change, so there is no UI screenshot.

### Follow-up review

Quick104 reviewed commit `34c4803` and found two runtime root causes plus a documentation mismatch. Commit `5cb55b9` now:

- rejects frozen Profile 7 and Profile 8 v1 recipes through the retryable capability-mismatch path before local progressive remux starts;
- treats FFmpeg capability-probe errors and timeouts as inconclusive, so they are not cached as a process-lifetime lack of support; and
- documents that both `dovi_rpu` and `filter_units` are required.

The follow-up regressions failed on `34c4803` and passed after the fix. The repository's [upstream CI run](https://github.com/Silo-Server/silo-server/actions/runs/32309181182) passed Go, Web, and docs hygiene, and the follow-up CodeRabbit review reported no actionable comments.

PR #657 is the first landing PR. PR #659 will be rebased onto the updated `main` afterward and will consolidate the DV path into its general versioned remux-filter contract instead of retaining parallel `dvrv` and `rfv` mechanisms.

### AI Disclosure
- Tool(s): OpenAI Codex
- Model(s): gpt-5
- Involvement: fully AI-generated, with human-provided playback context and approval
- Adversarial review: The initial scope review excluded Safari transport work already covered by #653. A stable-diff review traced recipe v2 through local, reconstructed, HLS, and proxy paths. CodeRabbit then identified three valid gaps: an unbounded FFmpeg capability probe, partial session updates clearing recipe versions, and a missing server-executor assertion. I fixed all three, added targeted regressions for the runtime cases, reran the full repository CI successfully, and confirmed all three review threads resolved. The repository's Codex companion script was not installed, so I used an independent read-only reviewer as the documented fallback.

## Checklist
- [x] I ran the repo verify commands: `make lint`, `cd web && pnpm run lint`, `cd web && pnpm run format:check`, and relevant `go test ./...`.

## Summary by CodeRabbit

* **Enhancements**
  * Improved Dolby Vision Profile 7 to HDR10 playback and remuxing.
  * Added support for preserving Dolby Vision playback settings across sessions, streams, and playback tokens.
  * Updated playback transformation handling to use the latest recipe version.

* **Bug Fixes**
  * Prevented playback from using outdated or incomplete Dolby Vision transformation settings.
  * Improved capability detection for HDR10 conversion and safer fallback behavior.
  * Updated FFmpeg filter requirements to correctly isolate the HDR10 base layer.

### DIFF
--- a/cmd/playbackfixtures/main.go
+++ b/cmd/playbackfixtures/main.go
@@ -335,7 +335,7 @@ func goldenCapabilityResponse() playback.CapabilityResponseV3 {
 		// fixture pins the full set so a client sees every shape it must parse.
 		Transformations: []playback.TransformationV3{
 			{Name: playback.TransformationAudioToAACV3, Executor: playback.ExecutorServerV3, RecipeVersion: "1", ValidatedClaims: []string{playback.ClaimAudioDecodeV3}},
-			{Name: playback.TransformationServerDV7HDR10V3, Executor: playback.ExecutorServerV3, RecipeVersion: "1", ValidatedClaims: playback.DV7ToHDR10ClaimsV3()},
+			{Name: playback.TransformationServerDV7HDR10V3, Executor: playback.ExecutorServerV3, RecipeVersion: playback.TransformationServerDV7HDR10RecipeVersionV3, ValidatedClaims: playback.DV7ToHDR10ClaimsV3()},
 			{Name: playback.TransformationVideoToH264V3, Executor: playback.ExecutorServerV3, RecipeVersion: playback.TransformationVideoToH264RecipeVersionV3, ValidatedClaims: []string{playback.ClaimH264DecodeV3}},
 		},
 	}
@@ -573,7 +573,7 @@ func goldenConformanceMatrix() playback.ConformanceMatrixV3 {
 	dv7File.VideoTracks[0].VideoRangeType = "DOVIWithEL"
 	dv7Request := conformanceHDRRequest()
 	dv7Request.PlaybackAttemptID = "attempt-dv7-hdr10"
-	dv7Registry := playback.NewTransformationRegistryV3([]playback.TransformationSpecV3{{Name: playback.TransformationServerDV7HDR10V3, RecipeVersion: "1", Available: true}})
+	dv7Registry := playback.NewTransformationRegistryV3([]playback.TransformationSpecV3{{Name: playback.TransformationServerDV7HDR10V3, RecipeVersion: playback.TransformationServerDV7HDR10RecipeVersionV3, Available: true}})
 	planner = append(planner, makePlannerScenario("dolby_vision_7_hdr10_fallback", "hdr_dv_matrix", dv7Request, dv7File, nil, settings, dv7Registry))
 
 	audioAdaptFile := conformanceHDRFile()
@@ -858,6 +858,6 @@ func conformanceRegistry() *playback.TransformationRegistryV3 {
 	return playback.NewTransformationRegistryV3([]playback.TransformationSpecV3{
 		{Name: playback.TransformationAudioToAACV3, RecipeVersion: "1", Available: true},
 		{Name: playback.TransformationVideoToH264V3, RecipeVersion: "1", Available: true},
-		{Name: playback.TransformationServerDV7HDR10V3, RecipeVersion: "1", Available: true},
+		{Name: playback.TransformationServerDV7HDR10V3, RecipeVersion: playback.TransformationServerDV7HDR10RecipeVersionV3, Available: true},
 	})
 }
--- a/docs/architecture/playback-protocol-v3.md
+++ b/docs/architecture/playback-protocol-v3.md
@@ -784,14 +784,14 @@ A transformation is a named, versioned media operation with claims attached.
 | --- | --- | --- | --- | --- |
 | `audio_to_aac` | `server` | `1` | — | `audio_decode` |
 | `video_to_h264` | `server` | `2` | `sdr` output | `h264_decode` |
-| `server_dv7_to_hdr10` | `server` | `1` | `hdr10` output | `dolby_vision_metadata_removed`, `hdr10_base_layer_preserved`, `enhancement_layer_discarded` |
+| `server_dv7_to_hdr10` | `server` | `2` | `hdr10` output | `dolby_vision_metadata_removed`, `hdr10_base_layer_preserved`, `enhancement_layer_discarded` |
 
 They are advertised only if the installed FFmpeg actually has the required
 capability, probed once at startup:
 
 | Transformation | Probe |
 | --- | --- |
-| `server_dv7_to_hdr10` | `ffmpeg -bsfs` contains `dovi_rpu` |
+| `server_dv7_to_hdr10` | `ffmpeg -bsfs` contains `dovi_rpu` and `filter_units` |
 | `audio_to_aac` | `ffmpeg -encoders` contains an `aac` encoder |
 | `video_to_h264` | `ffmpeg -encoders` contains any of `libx264`, `h264_qsv`, `h264_vaapi`, `h264_nvenc`, `h264_videotoolbox` |
 
--- a/docs/architecture/playback-protocol-v3.md
+++ b/docs/architecture/playback-protocol-v3.md
@@ -127,9 +127,9 @@ unknown, and both look like an absent field.
 delivery classes a client negotiates in. §4 gives the folding.
 
 `transformations` advertises only what the *installed* FFmpeg was probed for at
-startup — a server without a `dovi_rpu` bitstream filter does not list
-`server_dv7_to_hdr10`. A client must not assume a transformation exists because
-this document names it.
+startup — a server without both the `dovi_rpu` and `filter_units` bitstream
+filters does not list `server_dv7_to_hdr10`. A client must not assume a
+transformation exists because this document names it.
 
 `enabled` survives from the rollout period and is now constant `true`; the
 negative shape was deliberately removed before v1 lock because v3 is the only
--- a/docs/design/schemas/playback-v3/v3/fixtures/valid/capability_response.json
+++ b/docs/design/schemas/playback-v3/v3/fixtures/valid/capability_response.json
@@ -32,7 +32,7 @@
     {
       "name": "server_dv7_to_hdr10",
       "executor": "server",
-      "recipe_version": "1",
+      "recipe_version": "2",
       "validated_claims": [
         "dolby_vision_metadata_removed",
         "hdr10_base_layer_preserved",
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -509,6 +509,7 @@ func identityRecipeCard(s *playback.Session) playback.RecipeCard {
 	switch s.PlayMethod {
 	case playback.PlayRemux:
 		card := playback.NewRemuxRecipeCard(s.ID, s.UserID, s.ProfileID, s.MediaFileID, s.TranscodeAudio, s.AudioTrackIndex, s.RemuxDVMode)
+		card.RemuxDVRecipeVersion = s.RemuxDVRecipeVersion
 		card.TargetCodecAudio = s.TargetAudioCodec
 		card.TargetAudioChannels = s.TargetAudioChannels
 		card.TargetAudioBitrateKbps = s.TargetAudioBitrateKbps
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -878,6 +878,11 @@ func (h *PlaybackHandler) prepareIdentityTransportV3(r *http.Request, session *p
 			return preparedTransportV3{}, err
 		}
 	}
+	if !servedByProxy && planRequiresServerTransformationsV3(result.Plan) {
+		if err := validateAdvertisedTransformationsV3(result.Plan, h.transformationRegistryV3(r.Context()).Advertised()); err != nil {
+			return preparedTransportV3{}, &transportErrorV3{reason: "transcode_node_capability_unavailable", message: "No available playback executor can run the selected playback recipe.", retryable: true, cause: err}
+		}
+	}
 
 	previousNodeURL := session.TranscodeNodeURL
 	previousTransportID := remoteTransportID(session)
--- a/internal/api/handlers/playback_v3.go
+++ b/internal/api/handlers/playback_v3.go
@@ -857,6 +857,7 @@ func (h *PlaybackHandler) prepareIdentityTransportV3(r *http.Request, session *p
 	routeSession.TargetAudioChannels = result.TargetAudioChannels
 	routeSession.TargetAudioBitrateKbps = result.TargetAudioBitrateKbps
 	routeSession.RemuxDVMode = remuxDVModeForPlanV3(result.Plan)
+	routeSession.RemuxDVRecipeVersion = remuxDVRecipeVersionForPlanV3(result.Plan)
 
 	proxyNode, proxyErr := h.planIdentityProxyV3(r, session.ID, result)
 	if proxyErr != nil {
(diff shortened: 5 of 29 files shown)


Quick104:
## Review: three findings, plus a merge-order conflict with #659

Reviewed this branch at `34c4803`. I traced the new `dvrv` / `RemuxDVRecipeVersion` claim end to end (plan → session → recipe card → stream token → proxy → `ServeRemuxWithOptions`) and built/ran `./internal/playback/... ./internal/proxy/... ./internal/streamtoken/... ./internal/api/handlers/... ./internal/transcodenode/...` plus `make verify-playback-fixtures`. All green. (The two `TestBeginWebOperation*` failures in `internal/jellycompat` also fail on `main` — unrelated.)

**The claim plumbing itself is correct and symmetric** — produce (`identityRecipeCard` → `ToClaims`), transport (`prepareIdentityTransportV3` sets it before `identityStreamURLV3`), consume (`proxy.handleRemux`, `stream.go`), reconstruct (`ReconstructSession`), and snapshot/restore all carry it. The HLS side fails closed correctly because `StartTranscode` allowlists the exact filter string. `DVRPUProbe` now probing with the same full filter chain execution uses is a good catch.

The findings are all about what the *newly strict* version match does on paths that were previously name-only.

### 1. Version-mismatched Profile 8 plan silently falls through to legacy-auto remux — `internal/api/handlers/playback_v3.go:3229` (medium-high)

`remuxDVModeForPlanV3` used to match the transformation by name; it now also requires `executor==server && version=="2"` via `isCurrentServerDV7ToHDR10TransformationV3` (:3232). For `DVProfile == 7` the fall-through hits `RemuxDVRejectP7V3` (:3224) and fails loudly — fine. For **Profile 8** it reaches :3226, and since the strip plan sets `Claims.Video = VideoClaimsV3{HDR10: true}` (`plan_v3.go:393`), `Claims.Video.DolbyVision` is false, so it returns `""`. Empty mode is legacy auto in `startRemuxWithOptions` (`remux.go:259`), and `remuxDVProfile` only neutralizes profile 7 — so `buildRemuxArgs` emits **no `-bsf:v` at all**. `remuxDVRecipeVersionForPlanV3` also returns `""`, so the new 409 guard in `ServeRemuxWithOptions` never fires either.

This is reachable across an upgrade because frozen plans persist verbatim: `planstore/postgres.go` stores `current_plan` (JSON `PlanV3`, including `Transformations[].recipe_version`) for `MaxTokenTTL`, `frozenSeekReanchorResultV3` (`playback_v3.go:2174-2186`) hands `record.CurrentPlan` straight back to `prepareTransportV3`, and `ExecutableRecipeV3.ValidFor` compares only `PlanID` — so a v1 record validates fine.

Repro: Profile 8 source with `DVBLCompatID == 1`, HDR10-only client, progressive remux. Old binary plans it, freezing `server_dv7_to_hdr10@1`. Upgrade to this PR. Viewer seeks → seek-reanchor replan → v1 plan returned → `remuxDVModeForPlanV3` returns `""` → ffmpeg runs `-c copy` with no bitstream filter. DV RPUs are served while the plan advertises `dynamic_range: "hdr10"`, `claims.video.hdr10: true`, and the "Dolby Vision metadata is removed" degradation warning. On `main` this same input returned `strip_to_hdr10` and stripped correctly.

### 2. A transient probe timeout is cached permanently as "ffmpeg lacks the filters" — `internal/playback/remux.go:66` (medium)

`supportsDV7HDR10FilterChain` sets `available := err == nil && hasDV7HDR10FilterChain(out)` and unconditionally writes it to `dv7HDR10FilterSupportCache[bin]` (:73) with no re-probe. The new `context.WithTimeout(…, 3*time.Second)` (:63) makes `err != nil` reachable for reasons that say nothing about the binary.

Worth contrasting with the sibling probe, which already gets this right: `DVRPUProbe` distinguishes `dvRPUUnknown` and refuses to cache an inconclusive verdict (`dovi_rpu_probe.go:56-66, 152-158`). Also note the planner uses a *separate* `sync.Once`-memoized `ProbeTransformationRegistryV3` (`playback_v3.go:134-139`), so the two caches can disagree.

Repro: cold container on a loaded host, ffmpeg on a slow mount. The registry probe completes and advertises `server_dv7_to_hdr10@2`. Minutes later the first Profile 7 progressive remux calls `supportsDV7HDR10FilterChain`, that `exec` exceeds 3s, and `false` is cached for the process lifetime. Every subsequent start still plans `strip_to_hdr10`, and every `startRemuxWithOptions` returns `"dolby vision HDR10 remux requires the dovi_rpu and filter_units bitstream filters"` (:268) → HTTP 500 for all P7/P8 titles until restart.

Suggest mirroring `DVRPUProbe`: only cache a conclusive verdict, leave a timeout uncached.

### 3. The P7 half of the same fall-through gives an opaque 500 instead of a retryable error — `internal/api/handlers/playback_v3.go:3224` (low-medium)

With a v1-frozen P7 plan, `remuxDVModeForPlanV3` returns `RemuxDVRejectP7V3`, `startRemuxWithOptions` (`remux.go:300-303`) errors `"profile 7 remux is not eligible"`, `ServeRemuxWithOptions` writes a 500, and `StreamHandler.handleTransportStartFailure` (`stream.go:590`) only logs it. The HLS path in the same situation returns a structured retryable `transcode_node_capability_unavailable` (:780-782) that the client can act on.

**Findings 1 and 3 have one root cause and one fix.** The progressive/identity branch returns at `playback_v3.go:741`, *before* the local `validateAdvertisedTransformationsV3` check that protects the HLS branch at :779-782. Adding that check to `prepareIdentityTransportV3` closes both.

### Doc nit

`docs/architecture/playback-protocol-v3.md:129-130` still says "a server without a `dovi_rpu` bitstream filter does not list `server_dv7_to_hdr10`". The probe table at :794 was updated to `dovi_rpu` **and** `filter_units`; this prose wasn't.

### Merge-order conflict with #659

This PR and #659 independently add **two different versioning mechanisms for the same problem**: `dvrv` / `RemuxDVRecipeVersion` + `RemuxServeOptions.DVRecipeVersion` here, versus `rfv` / `RemuxFilterVersion` + `VideoBitstreamFilter` / `SourceVideoCodec` and a new `/stream/remux-v3/` route there. They conflict textually in 10 files: `remux.go`, `protocol_v3.go`, `recipecard.go`, `session.go`, `playback_v3.go`, `stream.go`, `transformations_v3.go`, `proxy/server.go`, `cmd/playbackfixtures/main.go`, and the protocol doc.

Whichever lands second needs a real rebase rather than a conflict resolution, and the pair should probably converge on **one** recipe-version claim instead of two. Worth noting that finding 1 here and #659's `ReconstructSession` finding are the same class of bug — a version field empty where a consumer requires it non-empty — so collapsing the schemes would likely eliminate both rather than needing two separate fixes.

I've deliberately left this PR's code alone pending that decision; happy to do the rebase once you and the maintainers pick a direction.

<sub>Review performed with AI assistance (Claude). Findings were verified by building and running the branch, not by reading alone; the repro scenarios above are traced through the code paths cited.</sub>

blurbery:
Thanks for the detailed review. I addressed the concrete #657 findings in `5cb55b9`:

- The local progressive/identity fallback now validates the plan’s exact server transformation before issuing a stream URL. Frozen Profile 7 and Profile 8 `server_dv7_to_hdr10@1` plans return the retryable `transcode_node_capability_unavailable` path instead of reaching legacy remux behavior; the regression covers both profiles.
- FFmpeg `-bsfs` probe errors and timeouts are now inconclusive and are not cached. The regression replaces a timed-out fake binary with a healthy one and proves the next request re-probes successfully.
- The protocol prose now requires both `dovi_rpu` and `filter_units`.

Local verification is green: the full `internal/playback` + contract suites (with the documented GPU skips), full `internal/api/handlers`, `internal/proxy`, `go vet` for the affected packages, playback fixtures, local-path hygiene, and `git diff --check`.

For the merge-order conflict, I’m treating #657 as the first landing PR. After it lands, I’ll rebase #659 onto the updated `main` and consolidate its general `RemuxFilterVersion` / `VideoBitstreamFilter` contract with the DV path so the result has one versioned remux-filter mechanism. I won’t resolve those overlapping files mechanically.

blurbery:
@Quick104 the three findings are fixed in `5cb55b9`, and Go, Web, docs hygiene, and the follow-up CodeRabbit review are all green. When you have a chance, could you please take another look at #657? The merge-order decision is #657 first, then a real #659 rebase that consolidates the versioning mechanisms. Thanks.

PR #659, opened 2026-08-16
fix(playback): normalize macOS Firefox HEVC resume boundaries

### LINKED ISSUE #658: [bug] macOS Firefox fails HEVC progressive-remux resume at an open-GOP boundary
### What happened

On macOS Firefox, resuming a copied HEVC progressive remux can fail in the platform decoder even though the same source plays from the beginning.

I reproduced this with Firefox 153.0.4 on macOS using Silo's current `server_remux_progressive` timeline behavior. The server reanchored the source at the preceding keyframe, the player retained the required `player_start_seconds` seek, and Firefox failed with `MEDIA_ERR_DECODE` before decoding a frame.

The affected source was a 2160p HEVC Main 10 Matroska file with an open GOP. It also carried Dolby Vision Profile 7, so I applied the same complete HDR10 base-layer isolation in both A/B arms. That keeps the separate DV7 correction in #657 out of the variable under test.

The only A/B difference was the initial leading-picture filter:

```text
baseline:    dovi_rpu=strip=1,filter_units=remove_types=63
filter-only: noise=drop=lt(pts\,startpts)*not(key),dovi_rpu=strip=1,filter_units=remove_types=63
```

With the filter-only stream, Firefox completed the same saved-position seek, entered `playing`, decoded 59 frames with none dropped, and reported no media error. This means the client-side workaround that skips `player_start_seconds` is unnecessary and would resume early.

### Steps to reproduce

1. Use current `main` (`edd919c5f7de7a0362f3d9d84f5145c7dd428771`) and its normal progressive timeline behavior.
2. In headed Firefox 153.0.4 on macOS, attach an HTTP/1.1 `200` chunked, non-range fMP4 generated from a non-zero input seek into an open-GOP HEVC MKV.
3. Assign `video.currentTime = 1.042`, matching the plan's saved position relative to the resolved keyframe origin, and call `play()`.
4. Observe the Apple VideoToolbox decode error and zero decoded frames.
5. Repeat with the same source, timestamp, transport, browser seek, and muxing flags, adding `noise=drop=lt(pts\,startpts)*not(key)` before the existing copy filters.
6. Observe that the seek completes and playback advances without a decoder error.

The test used an isolated local HTTP harness and the installed headed Firefox build so the macOS hardware decoder path was active. Headless Firefox was not used for the result.

### Expected behavior

A resumed HEVC copy remux should preserve the v3 timeline contract, complete the exact `player_start_seconds` seek, and play without falling back to an encoded lower-resolution route.

### Environment

- Silo commit: `edd919c5f7de7a0362f3d9d84f5145c7dd428771`
- Deployment: Docker
- Client: Web UI
- Browser: Firefox 153.0.4
- Platform: macOS
- Transport: `server_remux_progressive`, chunked fMP4, no byte ranges

### Relevant logs

Raw browser events from the unfiltered arm:

```json
{"type":"assigned-current-time","detail":1.042,"current_time":1.042,"ready_state":0,"quality":{"dropped":0,"total":0},"error":null}
{"type":"loadedmetadata","current_time":0.793,"ready_state":1,"quality":{"dropped":0,"total":0},"error":null}
{"type":"error","current_time":0.793,"ready_state":1,"quality":{"dropped":0,"total":0},"error":{"code":3,"message":"void mozilla::AppleVTDecoder::OnDecodeError(OSStatus): OnDecodeError:ffffbae2"}}
```

Raw browser events/result from the filter-only arm:

```json
{"type":"seeked","current_time":0.793,"ready_state":4,"quality":{"dropped":0,"total":0},"error":null}
{"type":"playing","current_time":0.793,"ready_state":4,"quality":{"dropped":0,"total":0},"error":null}
{"currentTime":3.259037,"paused":false,"readyState":4,"quality":{"dropped":0,"total":59},"error":null}
```

Raw harness process results:

```json
{"variant": "baseline", "returncode": 0, "stderr": ""}
{"variant": "filtered", "returncode": 0, "stderr": ""}
```

All hostnames, addresses, paths, item IDs, session IDs, and media titles are intentionally omitted. The event and process lines above are otherwise raw copy-paste output.

### Suggested scope

- Model this as a named, versioned server transformation rather than a client-supplied FFmpeg expression or an unadvertised boolean.
- Probe and advertise the required `noise` bitstream filter so local and remote executors fail closed.
- Scope it to evidenced macOS Firefox HEVC copy playback and execute it only for non-zero server seeks.
- Preserve `player_start_seconds`; do not include the old browser-side seek suppression.
- Keep zero-start playback, other codecs/platforms, route ordering, and video-transcode paths unchanged.

### AI Disclosure
- Tool(s): OpenAI Codex
- Model(s): gpt-5
- Involvement: fully AI-generated, with human-provided playback context and approval
- Adversarial review: A pre-implementation review rejected the existing fork patch because it combined route ordering, client seek suppression, and an unadvertised server filter. The isolated A/B kept exact resume behavior and changed only the server filter; it proved the filter-only arm plays while the baseline reaches the macOS decoder error. No implementation diff exists yet.

### Confirmations

- [x] I reproduced this myself on a real deployment
- [x] Logs above are raw copy-paste, not AI-summarized

### AI involvement in this report

AI-generated, human verified the original real-deployment failure and successful production playback. The isolated A/B was executed by OpenAI Codex with the human-approved server and local Firefox environment.

### AI tool + model

OpenAI Codex / gpt-5

blurbery:
## Problem
Closes #658

On macOS Firefox, resuming an HEVC `server_remux_progressive` stream at an open-GOP boundary can fail in VideoToolbox before the browser decodes a frame. Current protocol-v3 behavior correctly reanchors the copied stream at a preceding keyframe and then applies `timeline.player_start_seconds`; suppressing that browser seek avoids neither the malformed boundary nor the loss of exact resume semantics.

I reproduced this on current `main` with headed Firefox 153.0.4. With the same source, timestamp, chunked fMP4 transport, and exact browser seek, the unfiltered stream reached `MEDIA_ERR_DECODE` with zero decoded frames. Adding only `noise=drop=lt(pts\,startpts)*not(key)` produced `seeked` and `playing`, with 59 decoded frames, none dropped, and no media error. The full raw A/B output is in #658.

## Approach

I modelled the boundary normalization as the named, versioned server transformation `server_hevc_resume_leading_picture_drop@1` rather than exposing an arbitrary FFmpeg expression or adding a client workaround.

- Planning is limited to macOS Firefox, HEVC, and progressive remux. HLS, direct play, transcodes, other browsers, other operating systems, and zero-start media bytes are unchanged.
- The plan freezes the recipe from the initial start so later seeks keep the same execution contract. FFmpeg applies the filter only when the server seek is non-zero; the browser still applies `player_start_seconds` for exact resume.
- Capability discovery requires both the exact `noise` BSF and its expression-based `drop` option. The serving process repeats the bounded option probe before headers, so a replaced or downgraded FFmpeg fails before claiming a successful stream.
- The filter and recipe version survive session updates, restart reconstruction, signed stream tokens, and proxy execution. Source codec validation uses the same canonical primary-track metadata as planning.
- Recipe-bearing remuxes use the additive `/stream/remux-v3/` route. Older API or proxy binaries do not mount it, so mixed-version deployments fail closed instead of silently dropping the recipe.
- If the recipe is unavailable, the unsafe progressive candidate is skipped and the existing HLS/encoded fallback policy continues.

There is no web-client source change in this PR.

## Testing

Controlled headed-browser A/B from #658:

```text
baseline: MEDIA_ERR_DECODE, AppleVTDecoder ffffbae2, 0 decoded frames
filter-only: seeked, playing, 59 decoded, 0 dropped, no media error
```

The deployed Jellyfin FFmpeg exposes the required expression option:

```text
Bit stream filter noise
noise AVOptions:
  -amount            <string>     ...VA...B..
  -drop              <string>     ...VA...B..
  -dropamount        <int>        ...VA...B.. (from 0 to INT_MAX) (default 0)
```

Focused and relevant package verification:

```text
$ GOWORK=off go test ./internal/playback -skip '<four documented macOS NVENC probe flakes>' -count=1
ok  github.com/Silo-Server/silo-server/internal/playback  17.184s

$ GOWORK=off go test ./internal/proxy -count=1
ok  github.com/Silo-Server/silo-server/internal/proxy  0.557s

$ GOWORK=off go test ./internal/streamtoken ./cmd/playbackfixtures -count=1
?   github.com/Silo-Server/silo-server/internal/streamtoken      [no test files]
?   github.com/Silo-Server/silo-server/cmd/playbackfixtures     [no test files]

$ GOWORK=off go vet ./internal/playback ./internal/proxy ./internal/streamtoken ./cmd/playbackfixtures
(no output)

$ make verify-playback-fixtures
playback fixtures are current

$ make verify-local-paths
scripts/check-local-path-leaks.sh
```

The unfiltered full playback package run reached only the four known macOS NVENC smoke-probe failures documented by the repository; the complete remainder passes with those four skipped. Local compilation of `internal/api/handlers` was blocked by the machine's missing libvips/pkg-config dependency, so I also ran the fork's full Linux CI workflow on exact commit `e262a88f`. All Go, web, and docs jobs passed, including build, gofmt, vet, changed-lines lint, protocol fixtures, full tests, web lint/format/typecheck/build/tests, and local-path hygiene: [CI run 31941066928](https://github.com/blurbery/silo-server/actions/runs/31941066928). No `web/` files changed.

### AI Disclosure
- Tool(s): OpenAI Codex
- Model(s): gpt-5
- Involvement: AI-assisted implementation, test design, documentation, and review. I directed the scope and validated the playback behavior with the real macOS Firefox A/B described in #658.
- Adversarial review: The prescribed companion script was unavailable, so I used independent adversarial Codex passes. They found and I fixed four material gaps: exact FFmpeg option probing and live revalidation, mixed-version route fencing, canonical source-codec validation, and an unnegotiated client-quirk emission. The final passes found no remaining code-level issue. They noted that the real headed-browser A/B, rather than a committed binary fixture, proves the media behavior; an executable packet fixture would be useful follow-up coverage but was not judged a blocker.

## Risks & follow-up

Custom or older FFmpeg builds that lack expression-based `noise.drop` will no longer receive this progressive-copy recipe and will use existing fallback routing. The change is additive and requires no migration or coordinated Android/Apple client update. A small deterministic open-GOP fixture could strengthen byte-level regression coverage later; the current user-visible behavior is covered by the controlled real-browser A/B in #658.

## Checklist
- [x] I ran the repo verify commands: `make lint`, `cd web && pnpm run lint`, `cd web && pnpm run format:check`, and relevant `go test ./...`.

🤖 Developed with assistance from OpenAI Codex

## Summary by CodeRabbit

* **New Features**
  * Added improved HEVC resume playback for Firefox on macOS.
  * Resume handling now removes unsafe leading pictures after seeking while preserving playback position.
  * Added versioned remux streaming routes and capability reporting.
* **Bug Fixes**
  * Playback now validates remux recipes and rejects unsupported or outdated configurations.
  * Added fallback to HLS when progressive remuxing cannot safely support the requested playback.

### DIFF
--- a/cmd/playbackfixtures/main.go
+++ b/cmd/playbackfixtures/main.go
@@ -336,6 +336,7 @@ func goldenCapabilityResponse() playback.CapabilityResponseV3 {
 		Transformations: []playback.TransformationV3{
 			{Name: playback.TransformationAudioToAACV3, Executor: playback.ExecutorServerV3, RecipeVersion: "1", ValidatedClaims: []string{playback.ClaimAudioDecodeV3}},
 			{Name: playback.TransformationServerDV7HDR10V3, Executor: playback.ExecutorServerV3, RecipeVersion: "1", ValidatedClaims: playback.DV7ToHDR10ClaimsV3()},
+			{Name: playback.TransformationServerHEVCResumeLeadingPictureDropV3, Executor: playback.ExecutorServerV3, RecipeVersion: playback.TransformationServerHEVCResumeLeadingPictureDropVersionV3, ValidatedClaims: []string{playback.ClaimResumeLeadingPicturesRemovedV3}},
 			{Name: playback.TransformationVideoToH264V3, Executor: playback.ExecutorServerV3, RecipeVersion: playback.TransformationVideoToH264RecipeVersionV3, ValidatedClaims: []string{playback.ClaimH264DecodeV3}},
 		},
 	}
--- a/docs/architecture/playback-protocol-v3.md
+++ b/docs/architecture/playback-protocol-v3.md
@@ -785,6 +785,7 @@ A transformation is a named, versioned media operation with claims attached.
 | `audio_to_aac` | `server` | `1` | — | `audio_decode` |
 | `video_to_h264` | `server` | `2` | `sdr` output | `h264_decode` |
 | `server_dv7_to_hdr10` | `server` | `1` | `hdr10` output | `dolby_vision_metadata_removed`, `hdr10_base_layer_preserved`, `enhancement_layer_discarded` |
+| `server_hevc_resume_leading_picture_drop` | `server` | `1` | — | `resume_leading_pictures_removed` |
 
 They are advertised only if the installed FFmpeg actually has the required
 capability, probed once at startup:
@@ -792,6 +793,7 @@ capability, probed once at startup:
 | Transformation | Probe |
 | --- | --- |
 | `server_dv7_to_hdr10` | `ffmpeg -bsfs` contains `dovi_rpu` |
+| `server_hevc_resume_leading_picture_drop` | `ffmpeg -bsfs` contains the exact `noise` filter and `ffmpeg -h bsf=noise` exposes the expression-based `drop` option |
 | `audio_to_aac` | `ffmpeg -encoders` contains an `aac` encoder |
 | `video_to_h264` | `ffmpeg -encoders` contains any of `libx264`, `h264_qsv`, `h264_vaapi`, `h264_nvenc`, `h264_videotoolbox` |
 
@@ -800,10 +802,21 @@ pooled transcode nodes may still plan an HLS route using a transformation those
 nodes advertise but the local FFmpeg lacks, so the capability list is a floor,
 not a ceiling — one more reason a client must not precompute routes from it.
 
-An unavailable transformation is not silently skipped at plan time: it produces
-its own terminal reason (`dv_conversion_unsupported`,
+An unavailable conversion transformation is not silently skipped at plan time:
+it produces its own terminal reason (`dv_conversion_unsupported`,
 `audio_conversion_unsupported`, `video_conversion_unsupported`) so the client
-learns which conversion was missing rather than seeing a generic refusal.
+learns which conversion was missing rather than seeing a generic refusal. The
+HEVC resume transformation instead makes the unsafe progressive-copy candidate
+ineligible, allowing the existing HLS or encoded fallback policy to continue.
+
+`server_hevc_resume_leading_picture_drop` is frozen on macOS Firefox HEVC
+progressive-remux plans even when playback starts at zero. Its FFmpeg filter is
+executed only after a non-zero server seek, where it removes non-key open-GOP
+pictures presented before the resumed copy's first packet. The browser still
+applies `timeline.player_start_seconds`, preserving exact resume position.
+These plans use the version-fenced `/stream/remux-v3/` route on both integrated
+and proxy servers; older binaries do not mount that path and therefore cannot
+silently serve the same signed session without its frozen recipe.
 
 A client may advertise its *own* transformations in a delivery's
 `transformations[]` with `executor: "client"` — Dolby Vision profile 7 → 8.1
--- a/docs/design/schemas/playback-v3/v3/fixtures/valid/capability_response.json
+++ b/docs/design/schemas/playback-v3/v3/fixtures/valid/capability_response.json
@@ -40,6 +40,14 @@
       ]
     },
     {
+      "name": "server_hevc_resume_leading_picture_drop",
+      "executor": "server",
+      "recipe_version": "1",
+      "validated_claims": [
+        "resume_leading_pictures_removed"
+      ]
+    },
+    {
       "name": "video_to_h264",
       "executor": "server",
       "recipe_version": "2",
--- a/internal/api/handlers/playback.go
+++ b/internal/api/handlers/playback.go
@@ -489,7 +489,8 @@ func appendStreamToken(rawURL, token string) string {
 // identity stream token so a direct-play/remux session survives a restart (the
 // client re-supplies its byte position). Transcode sessions are told which URL
 // to play by their v3 plan; the URL here is an informational placeholder that
-// the plan's delivery URL supersedes.
+// the plan's delivery URL supersedes. A remux with a versioned server recipe
+// uses a route older binaries do not mount, fencing rolling deployments.
 func (h *PlaybackHandler) playbackStreamURL(s *playback.Session) string {
 	if s == nil {
 		return ""
@@ -498,13 +499,16 @@ func (h *PlaybackHandler) playbackStreamURL(s *playback.Session) string {
 		return fmt.Sprintf("/playback/transcode/%s/master.m3u8", s.ID)
 	}
 	card := identityRecipeCard(s)
-	return appendStreamToken(fmt.Sprintf("/stream/%s", s.ID), h.signSessionToken(card))
+	path := fmt.Sprintf("/stream/%s", s.ID)
+	if s.PlayMethod == playback.PlayRemux && (s.RemuxFilter != "" || s.RemuxFilterVersion != "") {
+		path = playback.RemuxRecipeStreamPathV3 + s.ID
+	}
+	return appendStreamToken(path, h.signSessionToken(card))
 }
 
-// identityRecipeCard builds the identity-only recipe for a direct-play or remux
-// session: reconstruction needs only ownership plus the audio selection, since
-// the bytes are served by HTTP Range / a re-spawned remux pipe at the
-// client-supplied position.
+// identityRecipeCard builds the durable recipe for a direct-play or remux
+// session. Direct reconstruction needs only ownership; remux reconstruction
+// also preserves its selected audio handling and versioned byte-level recipes.
 func identityRecipeCard(s *playback.Session) playback.RecipeCard {
 	switch s.PlayMethod {
 	case playback.PlayRemux:
@@ -512,6 +516,8 @@ func identityRecipeCard(s *playback.Session) playback.RecipeCard {
 		card.TargetCodecAudio = s.TargetAudioCodec
 		card.TargetAudioChannels = s.TargetAudioChannels
 		card.TargetAudioBitrateKbps = s.TargetAudioBitrateKbps
+		card.VideoBitstreamFilter = s.RemuxFilter
+		card.RemuxFilterVersion = s.RemuxFilterVersion
 		return card
 	default:
 		return playback.NewDirectRecipeCard(s.ID, s.UserID, s.ProfileID, s.MediaFileID)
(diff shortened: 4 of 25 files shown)


Quick104:
## Review: four findings, plus a merge-order conflict with #657

Reviewed every hunk against the PR head, built it, and ran `go vet` plus the `playback`, `proxy`, `api`, and `api/handlers` suites — all pass. I also validated the new FFmpeg recipe empirically against a real open-GOP HEVC file: `noise=drop=lt(pts\,startpts)*not(key)` does drop exactly the one leading picture, and the escaped comma parses correctly when chained with `dovi_rpu=strip=1`.

**Clean on inspection:** the boundary arithmetic, the `seekSeconds > 0` gate, token/claim symmetry (`vbsf` / `rfv` / `svc` produced in `identityStreamURLV3`, consumed in `proxy.handleRemux`), route registration on both the API router and the proxy, and the device-quirk matcher (`web` + `firefox/` + not `seamonkey/` + `macintosh|mac os x`, all lowercased). The `progressivePlan := plan` copy in `plan_v3.go` is also safe — the HLS branch re-applies its own subtitle decision and re-finalizes plan identity, so nothing is lost by no longer mutating `plan`.

### 1. Freezing the recipe on plain stream-copy plans breaks previously-working local playback — `internal/api/handlers/playback_v3.go:1011` (medium)

Freezing the recipe onto plans that need no transformation makes `planRequiresServerTransformationsV3` true, so `refuseLocalIdentityWorkV3` now refuses playback the API node could serve itself.

Repro: cluster with `playback.local_transcode_fallback=false`, proxy nodes still on the pre-PR binary (i.e. any rolling upgrade). A macOS Firefox user plays HEVC/AAC MP4 **from position 0** — no audio conversion, no DV strip. Before this PR `plan.Transformations` was empty, `refuseLocalIdentityWorkV3` returned nil, and a plain `-c copy` remux served from the API node. Now the plan carries `server_hevc_resume_leading_picture_drop`, `planIdentityProxySessionV3` narrows the pool to nodes advertising it (none), `ProxyNode` is nil, and the start fails with `capacity_unavailable`.

The refusal is self-contradictory in this case: the planner only formed the candidate because `firefoxResumeAvailableLocal` (`plan_v3.go:402`) proved the *local* ffmpeg carries the recipe — and at seek 0 the filter is never even added to argv, since `remux.go:188` requires `seekSeconds > 0`.

### 2. The revalidation probe is uncached and fires even when the filter won't be applied — `internal/playback/remux.go:278` (medium)

`supportsHEVCResumeLeadingPictureRecipeV3` forks `ffmpeg -h bsf=noise` on every remux start with **no per-binary cache**, unlike `supportsDoviRPUFilter` (`remux.go:57`) which caches the identical `-bsfs` probe. Once the recipe is frozen, `opts.VideoBitstreamFilter` is non-empty on every subsequent request, so each HEAD preflight, each GET, and each reconnect/seek forks an extra synchronous ffmpeg (up to a 3s timeout) before the real remux starts.

Worse than latency, though: the check fires whenever `opts.VideoBitstreamFilter != ""` **even at `seekSeconds == 0`**, where `buildRemuxArgsWithVideoBitstreamFilterV3` (`:187`) would not apply the filter at all. So on an ffmpeg build lacking `noise`, playback *from the start* fails with a 500 despite never needing the option.

### 3. `ReconstructSession` copies the filter onto transcode cards, producing a pair the validator rejects — `internal/playback/transcode_manager.go:420` (low, latent)

It copies `card.VideoBitstreamFilter` into `Session.RemuxFilter` for **every** play method. A DV7 HLS-remux/copy session stores a card via `NewRecipeCard` with `VideoBitstreamFilter = "dovi_rpu=strip=1"` and `RemuxFilterVersion = ""` (`recipecard.go:102` — the version field is only ever set by `identityRecipeCard`). After a restart the reconstructed session carries exactly the combination that makes `ServeRemuxWithOptions` return HTTP 409 "stale remux recipe" (`remux.go:414`).

Latent today: such cards reconstruct as `PlayTranscode` and `HandleStream` rejects that branch with a 400 before reaching the remux path. So this is a missing guard rather than a live bug — but it's the kind that becomes live the moment the reconstruct path grows a case.

### 4. The local path re-derives the codec from live catalog rows instead of the frozen recipe — `internal/api/handlers/stream.go:172` (low)

`sourceVideoCodec` is recomputed from the live `media_files` row rather than read from the frozen recipe, so catalog drift turns a live session into a hard 409 instead of a replan. A rescan that rewrites `media_files.video_tracks[0].codec` — or clears `video_tracks` so the stale aggregate `codec_video` becomes authoritative, which this PR's own test at `playback_v3_test.go:3825` deliberately sets up as disagreeing — makes `validateRemuxVideoBitstreamFilterV3` fail. `ServeRemuxWithOptions` then writes a plain-text 409 with no JSON error envelope and no terminal reason the client can act on, and `handleTransportStartFailure` only logs.

The proxy path doesn't have this exposure because it reads the codec from the signed token — which is the argument for the local path reading it from the frozen recipe too.

### Merge-order conflict with #657

This PR and #657 independently add **two different versioning mechanisms for the same problem**: `rfv` / `RemuxFilterVersion` + `VideoBitstreamFilter` / `SourceVideoCodec` and a new `/stream/remux-v3/` route here, versus `dvrv` / `RemuxDVRecipeVersion` + `RemuxServeOptions.DVRecipeVersion` there. They conflict textually in 10 files: `remux.go`, `protocol_v3.go`, `recipecard.go`, `session.go`, `playback_v3.go`, `stream.go`, `transformations_v3.go`, `proxy/server.go`, `cmd/playbackfixtures/main.go`, and the protocol doc.

Whichever lands second needs a real rebase rather than a conflict resolution, and the pair should probably converge on **one** recipe-version claim instead of two. Finding 3 above and #657's Profile 8 fall-through are the same class of bug — a version field empty where a consumer requires it non-empty — so collapsing the schemes would likely eliminate both rather than needing two separate fixes. Your `RemuxFilterVersion` / `VideoBitstreamFilter` shape is the more general one ("any bitstream filter"), which is worth weighing when picking a direction.

I've deliberately left this PR's code alone pending that decision; happy to do the rebase once you and the maintainers pick a direction.

<sub>Review performed with AI assistance (Claude). Findings were verified by building and running the branch and by exercising the ffmpeg recipe against real media, not by reading alone.</sub>

blurbery:
Merge-order decision: #657 will land first. I’ve pushed its review fixes in `5cb55b9`.

I’ll keep #659 separate until #657 is merged, then rebase #659 onto the updated `main`, address the four findings from this review, and consolidate the DV recipe into #659’s general `RemuxFilterVersion` / `VideoBitstreamFilter` contract. That rebase will deliberately produce one versioned remux-filter mechanism rather than preserving both `dvrv` and `rfv`, and I won’t use a mechanical conflict resolution across the overlapping files.

Issue #666, opened 2026-08-18
[bug] Progress updates alone keep dead sessions alive forever — ghost sessions hold transcode slots for 15h with zero byte flow


CoffeeKnyte:
## What happened

**Sessions that stopped playing hours ago are still counted as "watching now", and the server keeps transcoding for them.**

When someone stops watching, the server is supposed to notice and clean up — drop the
session from the active list and shut down the video conversion running for it. That
cleanup normally works. But there is a way for a session to slip past it and stay alive
forever.

The Android app sends the server a "here's where I am in the video" update roughly every
10 seconds. The server treats each of those updates as proof the person is still watching.
It never checks the thing that actually matters: **is the app still downloading any video?**

So if the app stops fetching video but keeps sending position updates, the server believes
the session is alive indefinitely. On our production server right now, two sessions have
been in this state for **just under 15 hours**. One of them downloaded a grand total of
**one** piece of video before it stopped, and has sent **5,324** position updates since.

The consequences, in order of how much they cost:

- **The server is transcoding video that nobody is receiving.** One of these sessions still
  has a live video conversion running, 15 hours after the app last asked for any video. It
  has been converting a film into a stream that goes nowhere. This occupies one of the
  server's limited hardware video-conversion slots.
- **These ghost sessions count against the limit on simultaneous conversions.** Real
  viewers can be refused a stream, or pushed onto a slower path, because slots are being
  held by sessions that stopped watching hours ago.
- **The admin "who is watching" view is wrong.** It shows people as actively watching who
  closed the app hours ago. There is currently no way for an operator to tell the ghosts
  apart from the real viewers.
- **Leftover working files pile up on disk.** 30 GB across 40 folders, against only 10
  conversions actually running.

None of this is visible as an error. Nothing crashes, nothing is logged as a failure, and
the sessions look completely normal in the admin view — they simply never end.

At the moment 4 of the 5 longest-running sessions on the server are ghosts. Only 1 is a
real viewer.

**Where it shows up:** Android app (all observed cases). Any client that reports position
independently of downloading video can trigger it.

**What triggers it:** every observed case began within ~25 seconds of a server restart. The
app reconnects, starts a session, fetches a little video, then stops fetching — but keeps
reporting position forever.

## Steps to reproduce

Commands assume the repository root is the cwd.

1. Start playback from a client that sends periodic progress updates.
2. Let the session establish (one or two segment requests).
[...]

PR #667, opened 2026-08-18
feat(streamtelemetry): measure bytes on every serving path, and merge them into one view

CoffeeKnyte:
Ten commits: four fixes to the byte paths that had to be correct before anything could be
measured, five that build the measurement, one docs consolidation.

Part of #265. Supersedes and closes #306.

## Problem

**The legacy view answers "who told us they're watching?" Telemetry answers "who is
actually receiving video?" Those turned out to be very different questions.**

The server has always tracked sessions by trusting what the app reports. An app says
"I'm at 42 minutes" every ten seconds, and the server takes that as proof someone is
watching. Nobody checked whether video was actually leaving the building.

That is also why a concurrency cap cannot tell a viewer from a ripper. It sees "one
session" either way: one person watching a film, or someone pulling the entire library at
link speed. And there is no single place to look — proxy health, admin sessions, node
sessions and playback stats each answer a different part of the question, from a different
store, with a different idea of what a "session" is.

Stream-telemetry measures the thing that costs money: bytes delivered. Putting the two
side by side for 18 hours is what exposed the gap — and it is not a rounding error:

- **Two sessions had been counted as "watching now" for 15 hours** without receiving
  video. One downloaded a single segment, then sent 5,324 position updates.
- **4 of the 5 longest-running sessions on the server were ghosts.** One was a real viewer.
- **3 of 10 running video conversions were feeding nobody**, holding hardware encode slots
  that real viewers get refused.

The operational consequence is that **the numbers the server reports today cannot be
trusted for capacity decisions**. Concurrent-stream counts, transcode-cap enforcement and
the admin "who's watching" view are all built on the legacy signal, and it over-reports.
Anyone sizing hardware or tuning the transcode cap from those numbers is working from
inflated figures, with no way to know by how much.

That specific defect is filed separately as #666 — it belongs to the old view, not to this
branch. But it is the clearest illustration of the problem: **we were reporting numbers
that could not be checked, and they were wrong.**

### Why merge now

1. **It found a production bug the legacy view is structurally incapable of finding.** Not
   a hypothetical — it was still burning an encode slot while this PR was being written.
   The parity report is what made those sessions distinguishable from ordinary timing skew
   between stores; they showed as one-sided in **179 of 185** consecutive samples.
2. **The cost is measured, and it is negligible.** View build is a **median of 3 ms, p95
   10 ms, max 52 ms** across 183 samples under real load, with a Redis footprint of 2 keys.
   This is not a bet on how it will behave in production — it *has* behaved, for 18 hours.
3. **It is clean.** 183/183 complete views, zero failures, zero stale views, zero clock
   skew, zero restarts, and exactly one mismatch in the entire run — a transient
   node-identity blip after a container recreate, understood and benign.
4. **The fix for #666 needs this to verify itself.** Any change to session liveness has to
   be checked against something, and the legacy view is the thing being corrected.
   Telemetry is the only independent measurement available.
5. **It is the foundation abuse enforcement has to stand on.** #306 set out to stop
   rippers and over-consumption, and it could not — not because the policy was wrong, but
   because there was nothing underneath it to measure against. Everything enforcement
   needs comes from here: per-session and per-user byte totals, delivery rate, viewer
   addresses, and one merged view every process publishes into. Landing this is what makes
   #306's goal buildable, and lets its thresholds be set from observed distributions
   instead of guesses.

The honest framing is that this is **not** "no regressions after 18 hours". It is "18 hours
of production data, and the instrument earned its keep by catching something".

**Four things were also quietly broken on the byte paths themselves**, found while making
them measurable:

- **Streams through the proxy had no owner.** Compat proxy redirects dropped the user and
  profile, so those streams were recorded against nobody and silently escaped the cap.
- **The proxy and audiobook listeners recorded the wrong viewer address** — the address of
  the proxy in front, not the person watching.
- **The kernel fast path for sending files was dead through the entire proxy chain**, so
  every byte took a slower route than it needed to.
- **Stream tokens had no reliable creation time**, so a session's age could not be trusted
  after a replan.

### What this does, and deliberately does not do

Every byte-serving route in every process now reports what it served, to whom, and how
fast, into one merged picture — asynchronously, off the hot path, without trusting
anything the client says. Five router families across three kinds of process publish into
Redis; a pure function merges them; an admin endpoint serves the result and diffs it
against the two projections admins read today.

**It makes no decisions.** Nothing is blocked, throttled, cut or banned. No existing admin
read has been repointed onto it. No `/api/v1` response changed, no migration, no new
Postgres write.

### The four prerequisite fixes

`826c71f6` carries an immutable session creation time in stream tokens. The JWT `iat`
could not serve — signing overwrites registered claims on every mint, and replans mint
replacement tokens from a live session. Read order is explicit, and a missing timestamp
never invalidates an otherwise valid old token.

`d5d52c4e` populates `UserID`, `ProfileID` and `MediaFileID` in the compat proxy stream
token (`buildProxyRedirectURL`), which previously omitted them entirely. Cost: the token
grows ~80–130 URL characters, and claims are signed but not encrypted.

`82825336` mounts `clientip.Middleware` on the standalone proxy and ABS routers. This
deliberately changes recorded session IPs and `RemoteAddr`-based logs to the resolved
viewer address.

`760287d6` repairs the writer chain. Because `io.Copy` finds `io.ReaderFrom` by direct
assertion and never through `Unwrap()`, every `ResponseWriter` wrapper on a media route
must forward `ReadFrom` while preserving accounting. Seven wrappers were repaired over
shared helpers in `internal/httpstream`. The sendfile finding is the notable one — measured
with `strace -f -e trace=sendfile` over an 8 MiB body, the mounted proxy direct-play router
went from **0 sendfile calls to 6**. Slice size is a correctness constraint, not a tuning
knob: the deadline is an absolute time, so slice ÷ stall window is a hard floor on
sustained client rate. The old 64 MiB slice against a 180s window implied ~3 Mbit/s and was
reaping healthy slow clients; it is now 4 MiB (~186 kbit/s).

> If you touch `CopyChunked`, re-run the strace comparison. A byte-exact body and a correct
> `Range` status prove HTTP correctness, not sendfile.

### The measurement

`a9d54b6b` adds process-local observation for native routes, with bounded retention and
release-fold. `7a49cca1` publishes snapshots to Redis and merges a global view with
publisher epoch/sequence and a `complete`/`degraded` contract. `d332db9a` and `29083144`
enrol the remaining four families.

Enrolment is **typed, not a hand-maintained list**. Every media route is declared as a
`MediaRoute` carrying family, method, pattern, class, role, session key and cap relevance;
the wrapper is derived from the declaration, and a mount-site typo panics rather than
silently un-observing a route. A per-family **manifest test** walks the mounted routers and
diffs every `(method, pattern)` against a checked-in golden — a new media route fails the
build until it is classified. 94 of 1,003 route entries are observed; the other 909 are
pinned as non-media. This is the fix for the failure mode that sank the earlier attempt,
where an unenrolled byte path was both invisible and unkillable, rediscovered four separate
times.

Two rules worth not relitigating: **viewer bytes and viewer IP belong exclusively to the
outermost viewer-facing edge** (proxy→node hops are `internal_relay`, never cap-relevant),
and **a transcode node publishes a correlation key and nothing else** — it cannot know who
is watching, so it must never fall back to generic capture and record the proxy's address
as a viewer IP.

`b6c9a1c7` adds `GET /api/v1/admin/stream-telemetry/parity`, additive and behind the same
authorization as `/admin/sessions`. P0d was specified as "compare, *then* repoint". **The
comparison shipped; the repoint deliberately did not** — the admin session payload is a
join of ~50 display fields telemetry is not canonical for, and the parity evidence now
argues against a blind swap anyway. The view is a read-driven TTL cache with single-flight
refresh, not a ticker: `BuildGlobalView` measures 347 ms at 50,000 sessions, so a ticker
would pay full rebuild cost on every server forever whether or not an admin is looking.

`f6c8b04c` reduces ten documents (4,517 lines) to two, and scopes the working document to
what shipped. The enforcement and rules designs moved to the appendix — see below.

The design is `docs/design/2026-08-17-stream-telemetry.md`; §6 carries the phase plan and
what blocks P1, and `…-appendix.md` holds the deferred enforcement design along with the
approaches already tried and rejected. Sections 3 and 5 are deliberate stubs, not gaps —
Go comments cite section numbers directly, so they are not renumbered away.

### Why monitoring alone, and enforcement deferred

**Every threshold in an enforcement rule is a guess until the traffic has been measured,
and this is the thing that measures it.** The enforcement design was written and reviewed
first, and that was the wrong order: it set numbers against a distribution nobody had seen.
This branch makes monitoring first-class; enforcement is built on top of it afterwards,
against real data. That is why #306 is being closed rather than merged.

The soak behind the claims above, in full:

| Measure | Result |
|---|---|
| `view.complete` | 183/183 |
| Build failures / stale views / clock skew | 0 / 0 / 0 |
| `build_took_ms` | median 3, p95 10, max 52 |
| Redis `silo:stelem:*` | 2 keys |
| Contradictions between publishers | 1, transient, after a container recreate |
| Container restarts caused | 0 |

Rolled family-by-family — off → `native` → `+jellycompat` — each held until quiet before
widening. Live sessions resumed mid-stream across three container recreates (QSV transcode,
`-c:v copy` direct play, HLS all confirmed).

## Risk / follow-ups

- **This is a production streaming change, not a read-only addition.** It inserts executable
  code into every live byte path and can affect which optional interfaces are visible,
  whether `io.Copy` selects `ReadFrom`, flush timing, HEAD and Range behaviour, error
  propagation, connection reuse, and ABS socket.io if `Hijacker` is not preserved. That risk
  has now been exercised rather than only reasoned about, but it is the right lens for review.
- **The soak covered `native` and `jellycompat` only.** `proxy` and `transcode_node` saw
  effectively no traffic (single-node `MODE=integrated`), and the multi-publisher merge was
  never exercised — every sample had exactly one publisher. Those rest on the per-family
  manifest tests and the two-publisher real-Redis integration test, not on production
  evidence. Worth stating plainly rather than letting "18 hours clean" imply more than it does.
- **`abs` is enrolled but unexercised** — the soak host has no audiobook traffic and never
  has. Code, manifest test and unit tests are in place; it stays off in the default family
  set until somewhere real can exercise it. Low priority.
- **Family defaults are unchanged.** `SILO_STREAM_TELEMETRY_FAMILIES` still defaults to
  `native,proxy,transcode_node`; jellycompat and ABS share the API process, so defaulting
  them on would widen instrumentation on upgrade alone. Now that native and jellycompat have
  run in production, promoting jellycompat into `defaultObservedFamilies` is a reasonable
  follow-up — deliberately not done in this PR.
- **Byte semantics are pre-compression on compressible routes** (subtitle and font),
  wire-equal on bulk routes excluded from compression. Documented at each capture site; do
  not "fix" it by moving the wrapper.
- **Legacy retirement is a separate project** with nine named consumers, gated on repeated
  parity agreement over time. A single parity report samples three independently updated
  stores and will always show one-sided differences.
- **#666 is not fixed here.** It is a defect in the legacy store that this branch's parity
  projection surfaced. Fixing it needs an independent measurement to verify against, which
  is an argument for landing this first.

## Verification

- `gofmt -l ./cmd ./internal` — clean
- `go build ./...` — OK
- `go vet ./...` — clean
- `go test ./...` — **121 packages ok**, 1 failure: `TestResolveCopySeekAnchorMatchesRealLongGOPHEVC`,
  which needs ffmpeg ≥5.x and the host has 4.4.2 (`Error splitting the argument list: Option
  not found`). Pre-existing and environmental; the test exists unchanged on `origin/main`.
- `golangci-lint run --new-from-merge-base=origin/main` — **0 issues** (matches how CI runs it)
- `make verify-local-paths` — clean
- All five per-family route manifest guards pass: `api`, `jellycompat`, `proxy`,
  `transcodenode`, `audiobooks/abs`
- `TestRedisStoreIntegration` passes against a real Redis, including the two-publisher case
- sendfile verified by `strace -f -e trace=sendfile` on a mounted proxy router, 8 MiB body
- **18-hour production soak**, 185 samples, results in the table above
- Frontend lint/format not run — no frontend changes in this branch, and pnpm is unavailable
  on this host

## AI-use disclosure

- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: AI-assisted; production soak, verification commands and their output are real
  and were executed, not synthesized
- Adversarial review: reviewing the soak data, my first conclusion was that telemetry had a
  coverage gap — it was reporting fewer sessions than the legacy store. Checking the request
  logs reversed it: those sessions had no byte flow, so telemetry was correct and the legacy
  store was wrong (#666). I also went back and narrowed the soak claims after checking what
  the run actually exercised rather than what it was configured for — `proxy`,
  `transcode_node`, `abs` and the multi-publisher merge saw no real traffic, so the Risk
  section states the coverage boundary rather than resting on the headline numbers.


Quick104 (commented):
Automated code review (Claude Code `/code-review`, multi-agent find → adversarial-verify pass). 23 findings posted inline: 8 confirmed correctness issues, 8 plausible correctness issues, 7 cleanup/convention notes. 10 additional candidates were refuted during verification and not posted.

Highest-priority items: the ABS login rate-limiter bypass (`cmd/silo/main.go:2795`), the jellycompat session-key mismatch (`internal/jellycompat/streamtelemetry.go:45`), the sticky `truncated` flag (`registry.go:265`), and the single-variable distributed-mode disable (`config.go:183`).

cmd/silo/main.go
```diff
@@ -2701,8 +2791,11 @@ func main() {
 	var absSrv *http.Server
 	if (mode == "integrated" || mode == "api") && deps.ABSHandler != nil && cfg.AudiobookshelfCompat.Listen != "" {
 		absRouter := chi.NewRouter()
+		if ipResolver != nil {
+			absRouter.Use(clientip.Middleware(ipResolver))
```
Quick104, line 2816: **[CONFIRMED] correctness: ABS clientip middleware defeats login rate limiter**

New `absRouter.Use(clientip.Middleware(ipResolver))` overwrites `r.RemoteAddr` with the X-Forwarded-For/X-Real-IP-derived address whenever the ABS listener's TCP peer is in `clientip.DefaultTrustedProxies` (RFC1918 + loopback), silently defeating the ABS login rate limiter's deliberate RemoteAddr-only keying (`internal/audiobooks/abs/login_ratelimit.go:97`: "spoofable forwarding headers are deliberately ignored").

**Failure scenario:** ABS listener fronted by any reverse proxy or Docker's 172.17.x bridge → attacker POSTs `/login` with rotating `X-Forwarded-For: 203.0.113.<n>` → each request gets a fresh `loginLimitBurst` bucket, credential-stuffing limit is bypassed and `LoginLimiter.buckets` grows one entry per spoofed IP until the 10-minute GC. Affects every deployment (the `Use()` is unconditional). `requestClientIP` already read `clientip.FromContext` first, so only the limiter regresses.

internal/jellycompat/streamtelemetry.go
```diff
@@ -0,0 +1,71 @@
...
+		Subject:     streamtelemetry.UserSubject(session.StreamAppUserID),
+		ProfileID:   session.ProfileID,
+		MediaFileID: mediaFileID,
+		// The compat token is a session token, not a signed stream token whose
+		// iat this path verifies. Recording "verified" would be a lie.
+		TokenIssuedAtSource: streamtelemetry.TokenIssuedAtSourceNone,
+		StartedAtSource:     streamtelemetry.StartedAtSourceFirstSeen,
+	}
+	if play != nil {
+		attachment.SessionID = play.ID
```
Quick104: **[CONFIRMED] correctness: jellycompat telemetry keyed on play.ID, not upstream id**

jellycompat sets `attachment.SessionID = play.ID` (compat play-session id) while the proxy (`claims.SessionID = playSession.UpstreamSessionID` from `buildProxyRedirectURL`), nodesessions and `playback_sessions_sync` all key on `playback.Session.ID`, and `BuildGlobalView`/`CompareLiveSessions` merge and match strictly by SessionID string; nothing maps `play.ID` to the upstream id.

**Failure scenario:** Jellyfin client `GET /Videos/{id}/stream` redirected to a proxy → compat edge publishes session `play.ID` (~0 bytes, just the 307) and the proxy publishes a second session under the upstream id carrying all the bytes → one viewing shows as two sessions in the merged view, and every compat session appears as `telemetry_only` with a `legacy_only` twin in both parity sources. Native uses `session.ID` and is unaffected.

internal/streamtelemetry/registry.go
```diff
@@ -0,0 +1,458 @@
...
+		}
+		s.mu.Unlock()
+	}
+	if obs.reserved {
+		r.observationReservations.Add(-1)
+	}
+}
+
+func (r *Registry) drop(reason string) {
+	r.truncated.Store(true)
```
Quick104: **[CONFIRMED] correctness: Registry.truncated flag is sticky for process lifetime**

`drop()` sets `r.truncated` and nothing ever clears it (the only write in the package), so one transient capacity drop marks every later `SnapshotAt` Truncated for the process lifetime; `BuildGlobalView` then marks the publisher `PublisherDegraded`, adds `publisher_truncated` and pins `Complete=false` until restart. Dropped counters are documented as monotonic; Truncated is not, and Complete's stated purpose (current blindness) requires it to be recoverable.

**Failure scenario:** One burst past MaxObservations/MaxTransfers (or one attach with empty SessionID on a playback route) at 09:00 → at 12:00 with an idle fleet `GET /api/v1/admin/stream-telemetry/parity` still reports `complete=false`, `incomplete_reasons=[publisher_truncated]`; a later real truncation is indistinguishable.

internal/streamtelemetry/config.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+	if value := os.Getenv(keyPrefixEnv); value != "" {
+		if strings.TrimSpace(value) == "" || strings.IndexFunc(value, unicode.IsSpace) >= 0 {
+			distributedInvalid = append(distributedInvalid, keyPrefixEnv)
+		} else {
+			cfg.KeyPrefix = value
+		}
+	}
+	if cfg.SweepInterval > time.Duration(1<<63-1)/3 || cfg.Freshness < 3*cfg.SweepInterval {
```
Quick104: **[CONFIRMED] correctness: Env cross-checks use defaults; one var disables distributed**

The cross-checks (`cfg.Freshness < 3*cfg.SweepInterval`, `cfg.MembershipTTL <= cfg.Freshness`) compare an env-supplied value against the DEFAULT of the other knob, so setting a single variable disables distributed mode and logs an error blaming a variable the operator never set. `config_test` only pins the both-variables-set cases.

**Failure scenario:** Operator sets only `SILO_STREAM_TELEMETRY_SWEEP_INTERVAL=2s` (default Freshness 5s < 6s) or only `FRESHNESS=60s` (default MembershipTTL 60s <= 60s) → `distributedInvalid` gets freshnessEnv/membershipTTLEnv, `cfg.Distributed=false`, main.go keeps `NewLocalStore` → every process runs local-only and the merged view/parity sees only its own publisher yet reports `Complete=true`.

internal/streamtelemetry/registry.go
```diff
@@ -0,0 +1,458 @@
...
+	return &r.shards[h.Sum64()%shardCount]
+}
+
+func (r *Registry) SetRealtimeConnection(sessionID string, connected bool) {
+	if r == nil || !r.cfg.Enabled || sessionID == "" {
+		return
+	}
+	shard := r.shard(sessionID)
+	shard.RLock()
+	s := shard.sessions[sessionID]
```
Quick104: **[CONFIRMED] correctness: SetRealtimeConnection dropped if socket opens before attach**

`SetRealtimeConnection` is a no-op when `shard.sessions[sessionID]` is nil and the value is never stored or re-applied (no pending map, `newLogicalSession` has no realtime input), so in the normal client ordering (`POST /playback/start` → open realtime websocket → first `GET /stream/{id}`) `RealtimeConnectionAlive` stays false for the whole session.

**Failure scenario:** Web client opens the socket as soon as sessionId exists (`usePlaybackRealtime.ts:86-96`) before any enrolled media route is hit; `setRealtimeConnectionState(true)` finds no logical session and drops it; `attach()` later creates the session with `realtimeAlive=false` and `setRealtimeConnectionState` early-returns when manager state already matches, so views report `RealtimeConnectionAlive=false` for a live control socket. `session_ws_test.go` only passes because it seeds an attach before dialing.

internal/api/handlers/downloads.go
```diff
@@ -534,7 +537,10 @@ func (h *DownloadHandler) handleDirectDownload(w http.ResponseWriter, r *http.Re
 			return
 		}
 	}
-	if err := h.svc.ServeDirect(r.Context(), w, r, userID, fileID, r.URL.Query().Get("format"), filter); err != nil {
+	serveCtx := downloads.WithServeAuthorized(r.Context(), func(target downloads.FileTarget) {
+		attachTransfer(r.Context(), userID, apimw.GetProfileID(r.Context()), target.MediaFileID)
+	})
+	if err := h.svc.ServeDirect(serveCtx, w, r, userID, fileID, r.URL.Query().Get("format"), filter); err != nil {
```
Quick104: **[CONFIRMED] correctness: direct-download lacks rolling deadline; 120s WriteTimeout**

`handleDirectDownload` passes the raw ResponseWriter to `ServeDirect` (→ `http.ServeContent`) with no `httpstream.NewRollingDeadlineWriter`, unlike the sibling `/downloads/{id}/file` at line 476, so the API server's absolute 120 s WriteTimeout (`cmd/silo/main.go` ~2635) truncates large direct downloads. Pre-existing on this line, but the PR touched this call and added the route to `skipNativeMediaCompression`, so it now streams the whole body in one unbounded sendfile ReadFrom.

**Failure scenario:** `GET /api/v1/direct-download?file_id=N` for a multi-GB original over a ~100 Mbit/s or slower link exceeds 120 s → connection killed mid-body, truncated download; `/api/v1/downloads/{id}/file` serving the same bytes survives.

internal/api/handlers/downloads.go
```diff
@@ -549,7 +555,11 @@ func (h *DownloadHandler) redirectDirectDownload(ctx context.Context, w http.Res
 	if err != nil {
 		return false, err
 	}
-	return h.redirectToProxy(w, r, secret, target, userID, "")
+	handled, err := h.redirectToProxy(w, r, secret, target, userID, "")
+	if handled {
+		attachTransfer(ctx, userID, "", target.MediaFileID)
```
Quick104: **[CONFIRMED] correctness: redirectDirectDownload attaches with empty profile id**

`redirectDirectDownload` hardcodes an empty profile id in both `redirectToProxy(..., userID, "")` and the new `attachTransfer(ctx, userID, "", ...)`, although the same handler's local branch uses `apimw.GetProfileID(r.Context())` two lines away and `redirectManagedDownload` passes the real profile.

**Failure scenario:** A household member hits `GET /api/v1/direct-download-proxy?file_id=N` → the telemetry transfer, the proxy's `attachTransfer` (from the token's empty ProfileID claim) and the `nodesessions.SessionInfo` all record ProfileID `""`, while the same download via `/api/v1/direct-download` records the real profile; per-profile attribution undercounts exactly the proxy-served traffic.

internal/httpstream/rolling_deadline.go
```diff
@@ -32,9 +32,10 @@ const (
 	// SetWriteDeadline per step rather than one per 32 KB chunk.
 	bumpStep = 15 * time.Second
 
-	// readFromChunk bounds each ReadFrom slice so the deadline keeps rolling
-	// during zero-copy (sendfile) transfers of large files.
-	readFromChunk int64 = 64 << 20
+	// readFromChunk bounds each ReadFrom slice so the deadline keeps rolling.
+	// At the default 180s window, 4 MiB permits steady clients down to roughly
+	// 186 kbit/s without expiring mid-slice.
```
Quick104: **[CONFIRMED] correctness: Documented 186 kbit/s floor is really ~203 kbit/s**

The stated floor "4 MiB permits steady clients down to roughly 186 kbit/s" (also in `docs/design/2026-07-09-streaming-write-deadline.md` and the stream-telemetry design) ignores that `bump()` is throttled by `bumpStep=15s` and now only fires after a slice completes, so a slice can start with only ~165 s of window left; the guaranteed floor is ~203 kbit/s. `readfrom_deadline_test.go` constructs the writer with `step=0` so the production throttle is never exercised.

**Failure scenario:** Slice N ends 14 s after the last bump → `bump()` returns early → slice N+1 starts against lastBump+180 s; a client sustaining the documented 186–200 kbit/s needs >165 s for the next 4 MiB and is reaped as `stalled_reap` despite the doc's guarantee. Either fix the number or bump before each slice as the old loop did.

internal/streamtelemetry/registry.go
```diff
@@ -0,0 +1,458 @@
...
+	if attachment.TokenIssuedAtSource == "" {
+		attachment.TokenIssuedAtSource = TokenIssuedAtSourceNone
+	}
+	if obs.route.Class == ClassTransfer {
+		if !reserve(&r.transferReservations, r.cfg.MaxTransfers) {
+			obs.countingOnly = true
+			r.drop("transfer capacity exhausted")
+			return
+		}
+		t := &transfer{id: obs.id, subject: attachment.Subject, profileID: attachment.ProfileID,
```
Quick104: **[PLAUSIBLE] correctness: One transfer record per Range request exhausts MaxTransfers**

Transfer-class attaches create one `transfer` record per HTTP request keyed by `obs.id` with no merge by (subject, mediaFileID), retained for Retention (5 min) after their single observation ends, against MaxTransfers (10 000) — and ranged byte routes (ABS `/api/items/{id}/file/{ino}` via ServeDirectPlay Range, `/downloads/{id}/file`, direct-download, ebook read, jellycompat `/Items/{id}/Download` incl. HEAD) are all ClassTransfer.

**Failure scenario:** A few ABS clients streaming audiobooks with many small Range GETs sustain ~34 transfer requests/s → 10 000 live records within 5 min → `drop("transfer capacity exhausted")`, further transfers countingOnly with bytes in DroppedBytes plus the sticky Truncated flag; `SnapshotAt` renders/clones up to 10k TransferViews every 1 s sweep (Redis publish itself is delta-based). Telemetry-only impact.

internal/proxy/egress.go
```diff
@@ -71,6 +73,16 @@ func (w *meteredResponseWriter) Write(b []byte) (int, error) {
 	return n, err
 }
 
+func (w *meteredResponseWriter) ReadFrom(src io.Reader) (int64, error) {
```
Quick104, line 90: **[PLAUSIBLE] correctness: Proxy egress meter now credited once per 4 MiB slice**

`meteredResponseWriter` previously hid `io.ReaderFrom` on purpose so every byte flowed through Write → `egressMeter.Add` per ~32 KiB; the new ReadFrom credits the meter once per completed 4 MiB slice, so slow viewers' bytes are invisible to the node's egress rate for the duration of a slice and then land in a single 1-second bucket.

**Failure scenario:** A 200–500 kbit/s direct-play viewer on a proxy node takes 60–170 s per 4 MiB slice; `RateKbps()` (60 s window) reads that stream as 0 for most samples, `/api/v1/status` EgressKbps under-reports committed egress and `nodepool.Planner.effectiveEgressKbps` can over-admit sessions onto a saturated proxy (partly mitigated by the planner's short-lived reserved-bandwidth bridge).

internal/streamtelemetry/store_redis.go
```diff
@@ -0,0 +1,392 @@
...
+	full = s.needFullResync || s.publishCount == 0 || (s.cfg.FullResyncEvery > 0 && s.publishCount%uint64(s.cfg.FullResyncEvery) == 0)
+	sets = make(map[string][]byte, len(fields))
+	if full {
+		for field, value := range fields {
+			sets[field] = value
+		}
+		return sets, nil, fields, true, nil
+	}
+	for field, value := range fields {
+		if old, ok := s.published[field]; !ok || old != digest128(value) {
```
Quick104, line 96: **[PLAUSIBLE] correctness: Delta publish trusts local digest map; key loss unnoticed**

Publish's delta path only rewrites fields whose in-process `published` digest changed and never checks that `silo:stelem:snap:<pub>` still exists (no EXISTS/HLEN/WATCH), so a non-erroring key loss leaves unchanged sessions unwritten until the next full resync while `meta.SessionCount` still counts them.

**Failure scenario:** maxmemory eviction on the shared Redis, an out-of-band DEL/FLUSHDB, a failover to a replica missing the key, or a >MembershipTTL publish gap letting PExpire lapse → hash holds only changed fields, `decodeSnapshotHash` sees `decodedSessions != meta.SessionCount` → `count_mismatch`/Degraded and under-reported live sessions for up to FullResyncEvery publishes (~60 s). Bounded and acknowledged in §8 of the design, but no eviction-policy requirement is documented for this keyspace.

internal/streamtelemetry/global.go
```diff
@@ -0,0 +1,648 @@
...
+			status.DecodeErrors = saturatingInt(status.DecodeErrors, problem.DecodeErrors)
+			if status.Reason == "" {
+				status.Reason = problem.Reason
+			}
+			if problem.Reason == publisherReasonOversized || problem.Reason == publisherReasonMetaMissing || problem.Reason == publisherReasonIdentityMismatch {
+				unusable = true
+			}
+		}
+		capturedAge := at.Sub(snapshot.CapturedAt)
+		if capturedAge < -params.Freshness {
```
Quick104, line 248: **[PLAUSIBLE] correctness: Clock skew detection one-directional; behind clock = stale**

Both skew checks (`heartbeatAge < -Freshness`, `capturedAge < -Freshness`) only fire for a publisher whose clock runs ahead; one running behind is silently classified `PublisherStale` and its sessions dropped from the merged view, with `ClockSkewSuspected` left false. Sequence/Epoch are recorded but never used to distinguish an advancing-but-behind publisher from a stalled one.

**Failure scenario:** A transcode node's clock drifts 8 s behind (NTP not converged after boot) → capturedAge 8 s > Freshness 5 s → marked stale, added to MissingPublishers, `Complete=false`, all its sessions vanish from the global view; past MembershipTTL of lag the roster floor drops it entirely so it is not even listed as missing.

internal/httpstream/rolling_deadline.go
```diff
@@ -131,28 +132,19 @@ func (s *RollingDeadlineWriter) Write(p []byte) (int, error) {
...
 		s.bump()
-		return io.Copy(writerOnly{s}, r)
+		return io.Copy(WriterOnly(s), r)
 	}
-	var total int64
-	for {
+	if s.statusCode == 0 {
+		s.statusCode = http.StatusOK
+	}
+	return CopyChunked(rf, r, readFromChunk, func(n int64, err error) {
```
Quick104: **[PLAUSIBLE] correctness: ReadFrom no longer bumps deadline before first slice**

The old loop called `s.bump()` before each `rf.ReadFrom`; the new code passes bump into CopyChunked's post-slice callback, so the FIRST slice runs against the window set at construction/WriteHeader. Latent today: `bump()` only suppresses bumps within 15 s, so any long gap before WriteHeader still refreshes, and every current caller reaches its first ReadFrom within seconds of the last bump.

**Failure scenario:** A future handler that calls WriteHeader and then blocks >180 s before its first ReadFrom (e.g. waiting on artifact readiness) would have its very first write fail with `os.ErrDeadlineExceeded` and `Outcome()` report `stalled_reap` for a stream that never stalled. A one-line pre-slice bump restores the old guard.

internal/streamtelemetry/writer.go
```diff
@@ -0,0 +1,119 @@
...
+	n, err := w.w.Write(p)
+	if w.bodyEligible {
+		w.observation.AddBytes(int64(n))
+	}
+	w.observation.recordWriteError(err)
+	return n, err
+}
+
+func (w *observedWriter) ReadFrom(reader io.Reader) (int64, error) {
+	if w.observation.cut.Load() {
```
Quick104, line 96: **[PLAUSIBLE] correctness: observedWriter.ReadFrom samples cut once; HTTP/2 differs**

`observedWriter.ReadFrom` checks `observation.cut` once at entry and the CopyChunked record callback never consults it, whereas `Write` checks per ~32 KiB; on HTTP/1.1 sendfile a cut cannot interrupt an in-flight transfer while HTTP/2 (no ReaderFrom on the h2 writer) falls back to io.Copy and is cut within 32 KiB. Latent: nothing calls `cut.Store` today and doc.go declares the package observational.

**Failure scenario:** When enforcement lands, an operator cutting a 20 GB direct-play session over HTTP/1.1 keeps sending until the whole file drains, while the same cut over HTTP/2 stops within 32 KiB — a kill switch whose behavior depends on protocol. Both halves should land together.

internal/proxy/server.go
```diff
@@ -415,15 +443,21 @@ func (s *Server) touchTranscodeSession(r *http.Request, claims *streamtoken.Clai
 // sessionInfo builds the node-session tracker record for a verified token,
 // copying the numeric ownership keys the node-session tracker needs.
 func sessionInfo(tr *nodesessions.Tracker, claims *streamtoken.Claims, kind string) nodesessions.SessionInfo {
+	startedAt, source := claims.StartedAt()
```
Quick104, line 487: **[PLAUSIBLE] correctness: Legacy tokens without ostn fall back to iat as started_at**

`sessionInfo` replaced `StartedAt: time.Now()` with `claims.StartedAt()`, which falls back to the JWT `iat` when the new `ostn` claim is absent, and stream tokens carry a 24 h TTL; `iat` is issue time (Sign rewrites RegisteredClaims on every mint), not playback start.

**Failure scenario:** During a rolling deploy, or when a client replays a cached 307 proxy URL / stored `?st=` token minted by an un-upgraded API/compat node with no `ostn`, the `silo:sessions:*` record and admin live-session list report `started_at` up to 24 h before playback began, and `comparePostgres` flags a started_at mismatch beyond the 1 s tolerance. Labeled via `started_at_source=issued_at` and bounded to legacy tokens; `compareNodeSessions` moves with it and does not flag.

internal/streamtelemetry/session.go
```diff
@@ -0,0 +1,190 @@
...
+		s.playMethod = a.PlayMethod
+		s.playMethods.add(a.PlayMethod)
+	}
+	if rank := startedAtRank(a.StartedAtSource); !a.StartedAt.IsZero() && rank > startedAtRank(s.startedAtSource) {
+		old := s.startedAt.Format(time.RFC3339Nano)
+		s.startedAt = a.StartedAt
+		s.startedAtSource = a.StartedAtSource
+		s.startedDegraded = a.StartedAtSource == StartedAtSourceIssuedAt || a.StartedAtSource == StartedAtSourceFirstSeen
+		if len(s.identityConflicts) < max {
+			s.identityConflicts = append(s.identityConflicts, IdentityConflict{
```
Quick104: **[PLAUSIBLE] correctness: started_at_replaced appended but flag not set**

The started-at branch of `recordConflicts` appends `IdentityConflict{Field:"started_at_replaced"}` (and may set `identityOverflowed`) without setting `s.hasIdentityConflict`, unlike the subject/profile loop, so the exported `HasIdentityConflict` flag can be false while `IdentityConflicts`/`LocalIdentityConflicts` is non-empty. No test pins the pure-upgrade case.

**Failure scenario:** Session first attached with `StartedAtSourceFirstSeen` (proxy, no ostn) then with `StartedAtSourceClaim` → a benign authority upgrade consumes the per-session conflict budget and can set `IdentityConflictsOverflowed`, yet consumers filtering on `HasIdentityConflict` never see it while consumers reading the list do. Either the entry does not belong in identityConflicts or the flag should be set.

internal/streamtelemetry/writer.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	n, err := w.w.Write(p)
+	if w.bodyEligible {
+		w.observation.AddBytes(int64(n))
+	}
+	w.observation.recordWriteError(err)
+	return n, err
+}
+
+func (w *observedWriter) ReadFrom(reader io.Reader) (int64, error) {
```
Quick104, line 95: **[CONFIRMED] simplification: Nine identical ResponseWriter ReadFrom wrappers**

Nine ResponseWriter wrappers (`activitylog.statusWriter`, `api/middleware.statusWriter`, `api/middleware.requestStatusWriter`, `abs.statusRecorder`, jellycompat's `compatImageProxyTagResponseWriter`/`loggingResponseWriter`/`debugResponseWriter`, `proxy.meteredResponseWriter`, `streamtelemetry.observedWriter`) hand-roll the same 'mark status written, then ReaderFromOf / io.Copy(WriterOnly) fallback / CopyChunked' ReadFrom tail around httpstream's primitives.

**Failure scenario:** A future fix to the ReadFrom forwarding logic has to be re-applied in nine files, and a missed site silently reverts to the io.Copy fallback and loses zero-copy sendfile plus deadline/byte accounting; a `httpstream.ForwardReadFrom(dst, markWritten func(), src, chunk, record)` helper would collapse all nine to a one-line call.

internal/api/media_routes.go
```diff
@@ -0,0 +1,75 @@
...
+	nativeRoute(http.MethodGet, "/api/v1/downloads/{id}/file", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodHead, "/api/v1/downloads/{id}/file", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodGet, "/api/v1/downloads/{id}/file-proxy", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodHead, "/api/v1/downloads/{id}/file-proxy", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodGet, "/api/v1/downloads/{id}/subtitles/{ref}", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodGet, "/api/v1/direct-download", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodHead, "/api/v1/direct-download", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodGet, "/api/v1/direct-download-proxy", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodHead, "/api/v1/direct-download-proxy", streamtelemetry.ClassTransfer, false),
+	nativeRoute(http.MethodGet, "/api/v1/ebooks/{content_id}/files/{file_id}/read", streamtelemetry.ClassTransfer, false),
```
Quick104, line 28: **[CONFIRMED] reuse: media_routes scaffolding + viewerIP copied across 5 pkgs**

The route-declaration/capture/lookup/observe skeleton (`xRoute` builder, `declareXMediaRoutes`, identical O(n) scan-then-panic `xMediaRoute` lookup, `observeX` wrapper) is copy-pasted near-verbatim across api, jellycompat, proxy, abs and transcodenode `media_routes.go`, and the viewer-IP fallback chain (`clientip.FromContext` → `net.SplitHostPort(RemoteAddr)` → `RemoteAddr`) is duplicated in api/jellycompat/proxy even though `internal/streamtelemetry/route.go:120` already has an identical unexported `viewerIP(r)`.

**Failure scenario:** A bugfix to the viewer-IP fallback (e.g. IPv6 handling) must be applied identically in four places or families silently diverge in the merged admin view; adding a sixth family or changing the lookup contract means touching five files. Export `viewerIP` and add a generic lookup/observe helper over the existing central `declarations` map.

internal/streamtelemetry/config.go
```diff
@@ -0,0 +1,261 @@
...
+		case FamilyNative, FamilyJellycompat, FamilyProxy, FamilyABS, FamilyTranscodeNode:
+			families[family] = true
+		default:
+			return nil, false
+		}
+	}
+	return families, true
+}
+
+func envEnabled(value string) bool {
```
Quick104: **[CONFIRMED] reuse: envEnabled() duplicates existing truthy-env parsers**

`envEnabled()` re-implements the 'truthy env var' parser (1/true/yes/on) that already exists independently in `internal/mail/mail.go`, `internal/api/ebook_conversion.go`, `internal/api/handlers/{subtitle_search,admin,admin_diagnostics}.go`, `internal/scanner/probe.go`, `internal/notifications/settings.go` and the sibling `internal/telemetry/config.go`.

**Failure scenario:** Now the ~9th ad hoc copy, each with slightly different accepted spellings (`enabled` is accepted in ebook_conversion.go but not here); a shared `envutil.Bool(name)` helper adopted here would stop the pattern spreading and give one place to fix inconsistencies.

internal/streamtelemetry/codec.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func timeFromUnixNano(value int64) time.Time {
+	if value == 0 {
+		return time.Time{}
+	}
+	return time.Unix(0, value)
+}
+
+func checkVersion(data []byte) error {
```
Quick104: **[CONFIRMED] efficiency: checkVersion double-parses every session/transfer record**

`decodeSession`/`decodeTransfer`/`decodeMeta` each call `checkVersion(data)`, which `json.Unmarshal`s the payload into a throwaway `{V int}` struct, then immediately unmarshal the same bytes again into `wireSession`/`wireTransfer`/`publisherMeta`, which already carry `V`.

**Failure scenario:** These decoders run per record from `RedisStore.LoadAll` feeding `BuildGlobalView`, which viewcache.go notes measured ~347 ms at the 50 000-session cap; every record is JSON-parsed twice per rebuild. Decoding once and checking `w.V != codecVersion` afterward halves the parse work with no behavior change.

cmd/silo/main.go
```diff
@@ -133,6 +137,76 @@ func resolveNodeIdentity() string {
...
+	})
+}
+
+// newStreamTelemetryRegistry builds the telemetry registry for this process,
+// preferring the Redis-backed store in distributed mode. It never falls back to
+// LocalStore on a failed ping: cache.NewRedisClient builds a lazy client that
+// never dials, and a Redis restart mid-deploy must not strand a publisher
+// local-only for the life of the process.
+func newStreamTelemetryRegistry(ctx context.Context, nodeID string, redisClient *redis.Client) *streamtelemetry.Registry {
+	streamTelemetryConfig := streamtelemetry.ConfigFromEnv(nodeID)
```
Quick104, line 185: **[CONFIRMED] efficiency: ConfigFromEnv parsed twice at startup**

`newStreamTelemetryRegistry` (main.go:899) and `newStreamTelemetryViewCache` (main.go:901) are called back-to-back and each independently calls `streamtelemetry.ConfigFromEnv(nodeID)`, re-parsing and re-validating every `SILO_STREAM_TELEMETRY_*` variable just to read `cfg.ViewTTL`.

**Failure scenario:** An invalid variable (e.g. bad view TTL) is logged twice via slog.Error/Warn, and the two calls could disagree if env changes between them; pass the already-built Config (or its ViewTTL) from `newStreamTelemetryRegistry` into `newStreamTelemetryViewCache`.

internal/api/handlers/playback.go
```diff
@@ -1075,22 +1108,14 @@ func (h *PlaybackHandler) HandleStartPlayback(w http.ResponseWriter, r *http.Req
 	h.handleStartPlaybackV3(w, r, body)
 }
 
+// PlaybackClientInfoFromRequest captures and normalizes playback client headers
+// at the HTTP request boundary.
+func PlaybackClientInfoFromRequest(r *http.Request) playback.ClientInfo {
+	return playback.ClientInfoFromRequest(r)
+}
+
```
Quick104: **[CONFIRMED] simplification: Three names for playback.ClientInfoFromRequest**

`playbackClientInfoFromRequest` (line 1117) wraps `PlaybackClientInfoFromRequest` (line 1113), which wraps `playback.ClientInfoFromRequest` — three names, one body; the unexported wrapper has 2 callers in playback_v3.go, the exported one 1 in api/media_routes.go, and proxy/media_routes.go calls the playback package directly.

**Failure scenario:** A future editor has to know which of three identical-behavior symbols to call; delete `playbackClientInfoFromRequest` and call `handlers.PlaybackClientInfoFromRequest` (or `playback.ClientInfoFromRequest`) everywhere — low-risk collapse.

internal/api/router.go
```diff
@@ -2836,6 +2825,16 @@ func NewRouter(deps Dependencies) chi.Router {
 							}
 
 							r.Get("/sessions", adminHandler.HandleListSessions)
+							// P0d parity projection: the merged telemetry view beside
+							// both legacy live-session projections and their diff. It
+							// compares only — the repoint is the separate retirement
+							// change, which this endpoint exists to give evidence for.
+							r.Get("/stream-telemetry/parity", (&handlers.StreamTelemetryParityHandler{
```
Quick104, line 2862: **[PLAUSIBLE] conventions: New admin parity endpoint has no docs/*-api.md entry**

`GET /api/v1/admin/stream-telemetry/parity` has no `docs/*-api.md` entry and no `docs/feature-changelog.md` entry, which CLAUDE.md requires for client-visible API changes; it is thoroughly described in `docs/design/2026-08-17-stream-telemetry.md` and is a comparison-only diagnostic, and the ~20 sibling admin routes in the same block also lack `docs/*-api.md` coverage, so the precedent is weak.

**Failure scenario:** Anyone consuming the v1 admin API from the docs tree cannot discover the endpoint or its response shape; the literal CLAUDE.md rule was not formally satisfied or ruled out in the PR.

Quick104:
## Review fixes — 10 commits, `f6c8b04c..88451eeb`

Every one of the 23 inline findings was checked against the code before anything was
changed. All 8 CONFIRMED correctness issues held. Three of the PLAUSIBLE ones resolve to
"documented, not changed" for reasons given below rather than being quietly dropped.

### Correctness

| Finding | Commit | What changed |
|---|---|---|
| ABS clientip defeats the login rate limiter | `97b9833` | `clientip.Middleware` now preserves the pre-overwrite transport peer in the request context, and the limiter keys on that. Unmounting the middleware would have given back the wrong viewer IP, which is what it was mounted for. |
| jellycompat keyed on `play.ID` | `ef565d8` | Compat attaches only under `UpstreamSessionID`. A play session doesn't learn that id until `ensureUpstreamPlayback`/`ensureTranscodeManifest` runs, so the pre-side-effect attach is a no-op on a session's first request and the handler attaches again the moment the id exists — still before any byte is written. A provisional key was rejected deliberately: it recreates the same ghost session, and a session whose id didn't exist a moment ago can't have a pending cut against it. |
| Sticky `truncated` | `9bd8c31` | Decays over `Freshness`, the same horizon `BuildGlobalView` uses to decide a publisher is current. The monotonic `Dropped*` counters keep the permanent record. |
| `SetRealtimeConnection` dropped before attach | `9bd8c31` | State for an unknown session is held in a bounded per-shard map, applied when an attach creates the session, and pruned by the sweep. |
| One env var disables distributed mode | `9bd8c31` | Repair-then-validate against resolved values: a knob left at its default moves to satisfy the invariant, and only a pair the operator actually pinned to inconsistent values is an error — naming only the variables they set. |
| Documented 186 kbit/s floor is really ~203 | `6ec97c1` | The `bumpStep` throttle was written for the 32 KB `Write` path; a ReadFrom slice is already bounded at 4 MiB, so throttling around one bought nothing and cost the documented floor. Slices now bump unconditionally, including before the first — which also closes the separate "no pre-first-slice bump" finding. |
| Proxy egress credited once per 4 MiB | `5a9e6e7` | Metered slices are 256 KiB: a credit every 4-10 s for a 200-500 kbit/s viewer, well inside the 60 s rate window, and still 8x more per sendfile call than the `Write` path it replaced. |
| direct-download: no rolling deadline; empty profile id | `ff19ab6` | Both fixed. |
| Ranged requests exhaust `MaxTransfers` | `fc46ad8` | A transfer is now one subject pouring one file over one route, not one HTTP request. `RequestCount` finally means what it says. |
| Delta publish trusts the local digest map | `fc46ad8` | An `HLEN` inside the same transaction catches a reconstructed-from-delta key and forces the next publish full — self-heals in one sweep instead of up to `FullResyncEvery`. |
| `started_at_replaced` appended without the flag | `fc46ad8` | A pure authority upgrade that confirms the recorded instant now records nothing at all (it isn't a conflict and shouldn't have consumed the budget); a replacement that moves the value sets both the flag and the list. |

### Documented rather than changed

- **Clock skew is one-directional.** Correct, and not fixable from one sample: the roster
  score *is* the publisher's own `CapturedAt`, so heartbeat and snapshot drift together
  and there is no independent clock to compare against. A publisher running behind is
  genuinely indistinguishable from one that stalled. `PublisherStatus` already exports
  `Epoch` and `Sequence` so two successive parity reads resolve it; that's now stated at
  the check.
- **`observedWriter.ReadFrom` samples `cut` once.** Latent — nothing calls `cut.Store`.
  Left deliberately rather than half-fixed: a per-slice check would still act at 4 MiB
  granularity, so h1 and h2 would still disagree, just less visibly. Written up as a
  requirement on the enforcement change that first introduces a caller, including a test
  that a cut behaves identically over both protocols.
- **Legacy tokens falling back to `iat`.** Verified already labeled — `StartedAtSource`
  flows through to `nodesessions.SessionInfo` as `started_at_source` and sets
  `startedDegraded`. No change.

### Cleanups

`813509d6` collapses all nine hand-rolled `ReadFrom` tails onto one
`httpstream.ForwardReadFrom`. `f25ef9b5` exports `streamtelemetry.ViewerIP` and drops the
three inline copies, adds `internal/envutil` and adopts it in both telemetry packages,
decodes each wire record once instead of twice, stops parsing `ConfigFromEnv` twice at
startup, and collapses the three names for `playback.ClientInfoFromRequest`.
`88451eeb` adds `docs/admin-api.md` with the full parity response shape, plus a changelog
entry.

The one finding only **half** addressed: the media_routes scaffolding duplicated across
five packages. The viewer-IP half is deduplicated; the generic lookup/observe helper is
not, because a five-package refactor of the enrolment skeleton is disproportionate on a
PR this size. Worth a follow-up before a sixth family is added.

### Three calls that go past what the findings asked

1. `envutil.Truthy` accepts `enabled`, which the old `envEnabled` did not — so
   `SILO_STREAM_TELEMETRY_ENABLED=enabled` and `SILO_OTEL_ENABLED=enabled` now turn on
   where they were silently off. Unifying on the superset was the point of the finding,
   but it is a live behavior change to two flags.
2. The config repair auto-*lowers* `SweepInterval` when only `FRESHNESS` is pinned low.
   That means more Redis writes, not just different blame attribution.
3. `drop("pending realtime capacity exhausted")` increments `droppedObservations` for
   something that is not an observation.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal` — clean
- `go test ./...` — the only failures are `TestBeginWebOperationRecoversDeadProcessLock`
  and `TestBeginWebOperationRejectsLiveProcessLock`, confirmed to fail identically on
  unmodified `f6c8b04c` in a separate worktree. `TestResolveCopySeekAnchorMatchesRealLongGOPHEVC`
  passes here (this host has a newer ffmpeg than the one in the PR description).
- `golangci-lint run --new-from-merge-base=origin/main` — 0 issues
- `make verify-local-paths` — clean
- The two new rolling-deadline tests were run against the base branch and **fail** there,
  so they reproduce the defect rather than merely passing.

### Still outstanding

The branch conflicts with `main`. One conflict, and it is not a code one: #675 deleted
`docs/design/2026-07-09-streaming-write-deadline.md` in the docs prune while this branch
modified it. The clean resolution is to accept the deletion and move the "Writer-chain
conformance" section this branch added into `docs/architecture/` — the same
distill-then-delete pattern #675 itself used — which removes the conflict at its root
without merging `main` into the PR. Not done here; it is a call for the author.

### AI-use disclosure

- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: AI-assisted. Every finding was read against the code before being acted
  on; the verification commands above and their output are real and were executed. The
  two new deadline tests were run against the unmodified base branch to confirm they fail
  there, and the two `web_component_test.go` failures were reproduced on the base branch
  before being called pre-existing.
- Adversarial review: my first plan for the jellycompat session-key fix was to add a
  session-rekey path to the registry so a provisional attach could be promoted once the
  upstream id appeared. Working through what that costs — reversing session reservations,
  route activity and conflict-budget state under two locks on the hot path of a branch
  whose merge argument is an 18-hour soak — it was the wrong trade for a first-request
  artifact. Making the attach placement conditional on whether the id is already known
  closes the finding completely with no registry changes at all.

cmd/silo/main.go
Quick104: **Fixed** in `97b9833`.

Confirmed as described. Unmounting the middleware was not an option — it is there so the
proxy and ABS listeners record the *viewer's* address rather than the proxy's, which was
one of the four prerequisite fixes. So `clientip.Middleware` now preserves the
pre-overwrite transport peer in the request context, and `clientIP` in
`login_ratelimit.go` reads that, falling back to `RemoteAddr` when the middleware never
ran. The deliberate "spoofable forwarding headers are ignored" rule is restored exactly.

`login_ratelimit_test.go` adds three cases, including one that rotates a forged
`X-Forwarded-For` behind a fixed peer and asserts the burst is never exceeded.

Worth knowing for other callers: anything that must key on an address a client cannot
forge has to use `clientip.PeerFromContext`, not `RemoteAddr`. That is now stated on the
middleware.

internal/jellycompat/streamtelemetry.go
Quick104: **Fixed** in `ef565d8`.

Confirmed — this would have doubled every compat viewing in the merged view and made all
of them read `telemetry_only` in parity.

Compat now attaches only under `UpstreamSessionID`, the key the proxy, `nodesessions` and
`playback_sessions_sync` all use. The wrinkle is that a play session does not *have* that
id until `ensureUpstreamPlayback`/`ensureTranscodeManifest` runs, so the pre-side-effect
attach is now a no-op on a session's first request and each handler attaches again the
moment the id exists — still before any byte is written, which is what §4.2 requires.

I first tried adding a session-rekey path to the registry so a provisional attach could be
promoted. Reversing session reservations, route activity and conflict-budget state under
two locks on the hot path was the wrong trade for a first-request artifact, and a
provisional key recreates exactly the ghost session this fixes. A brand-new session also
cannot have a pending cut against it, since its id did not exist a moment ago.

`CanonicalSessionKey` is updated `compat_play_session` → `upstream_playback_session`, and
the mounted-router test now asserts the id is the upstream one *and* is not any compat
play-session id.

internal/streamtelemetry/registry.go
Quick104: **Fixed** in `9bd8c31`.

Confirmed, including the "a later real truncation is indistinguishable" part, which is the
worse half.

`Truncated` now decays over `Freshness` — the same horizon `BuildGlobalView` already uses
to decide whether a publisher is current, so the two agree on what "now" means. The
monotonic `Dropped*` counters keep the permanent record, so nothing is lost; only the
"currently blind" claim expires, which is what `Complete` documents it to mean.

`TestTruncatedDecaysAfterFreshness` pins both halves: set at the drop and inside the
window, clear after it, with the dropped count still non-zero.

internal/streamtelemetry/config.go
Quick104: **Fixed** in `9bd8c31`.

Confirmed, and the failure mode was worse than a bad error message: every process silently
runs local-only while still reporting `Complete=true`, so the merged view looks healthy
while seeing one publisher.

Restructured to repair-then-validate. The config now tracks which variables were actually
set; a knob left at its default moves to satisfy the invariant (set `SWEEP_INTERVAL=2s`
alone and freshness becomes 6s), and only a pair the operator explicitly pinned to
inconsistent values is an error — naming just the variables they set. If the *defaults*
ever violate their own invariant, both names are reported, since that is a code bug rather
than an operator one.

Four single-variable cases added to `config_test.go`, one per direction of both
cross-checks — the gap you identified.

internal/streamtelemetry/registry.go
Quick104: **Fixed** in `9bd8c31`.

Confirmed, and the client ordering you cite is the normal one, so this was affecting every
web session rather than a rare interleaving.

State for a session that does not exist yet is now held in a per-shard `pendingRealtime`
map and applied when an attach creates the session. Bounded against the session budget
(`MaxSessions/shardCount + 1`) so it cannot grow without limit, and pruned by the sweep on
the same `Retention` horizon as an idle session, so a socket that opens and closes without
any media request does not leak an entry.

Three tests: state set before attach is applied, the capacity bound refuses when full, and
the sweep reclaims an orphan. Your point about `session_ws_test.go` only passing because it
seeds an attach first is exactly right — the new test does it in the real order.

internal/api/handlers/downloads.go
Quick104: **Fixed** in `ff19ab6`.

Confirmed. Pre-existing on that line, but as you note the PR made it materially worse by
adding the route to `skipNativeMediaCompression`, so the whole body now rides one
unbounded sendfile against a 120s absolute deadline.

`handleDirectDownload` now wraps in `httpstream.NewRollingDeadlineWriter` before
`ServeDirect`, matching `/downloads/{id}/file` two functions above.

internal/api/handlers/downloads.go
Quick104: **Fixed** in `ff19ab6`.

Confirmed — and the asymmetry you point at is the tell: the same handler's local branch
reads the real profile two lines away, and `redirectManagedDownload` passes it correctly.

`redirectDirectDownload` now reads `apimw.GetProfileID(ctx)` and passes it to both
`redirectToProxy` and `attachTransfer`, so the telemetry transfer, the token claim the
proxy re-attaches from, and the `nodesessions.SessionInfo` all carry the profile.

internal/streamtelemetry/store_redis.go
Quick104: **Fixed** in `fc46ad8`.

Confirmed. Acknowledged in §8 as bounded, but "bounded" assumed the key still exists — the
delta path never checks, so a non-erroring loss is invisible until the next full resync.

`Publish` now issues an `HLEN` inside the same transaction as the delta writes, so it sees
the post-write field count. A mismatch against what was just published means the key was
reconstructed from the delta alone, and the next publish is forced full. One extra
pipelined command, and it self-heals in one sweep instead of up to `FullResyncEvery` (~60s).

That covers all four causes you list — eviction, out-of-band `DEL`, replica failover, and a
lapsed `PExpire` — without needing an eviction-policy requirement documented for the
keyspace, since the store now detects the loss rather than assuming it cannot happen.

internal/streamtelemetry/session.go
Quick104: **Fixed** in `fc46ad8`.

Confirmed, and working through it the entry turned out not to belong in `identityConflicts`
at all for the case you describe.

Two changes. A started-at update that only *upgrades authority* while confirming the same
instant — the common proxy-then-claim path — now records nothing: it is not a conflict, and
consuming the per-session budget for it could set `IdentityConflictsOverflowed` for a
benign event. An update that genuinely *moves* the value means two sources disagree about
when playback began, so it sets `hasIdentityConflict` alongside the entry.

Either way the flag and the list can no longer disagree, which was your core point.
`TestStartedAtAuthorityUpgradeAndConflictAgree` pins both branches, including the pure
upgrade case that had no coverage.

internal/httpstream/rolling_deadline.go
Quick104: **Fixed** in `6ec97c1`.

Confirmed, and the arithmetic holds: `bump()` returning early means a slice can begin with
only `window - step` left, which is a ~203 kbit/s floor rather than the 186 kbit/s the
constant and both design documents promise.

Took your second option — bump before each slice, as the old loop did. The throttle exists
so a fast stream does not issue one `SetWriteDeadline` per 32 KB `Write`; a `ReadFrom`
slice is already bounded at 4 MiB, so throttling around one saves at most a syscall per
4 MiB and costs the documented guarantee. `forceBump` now runs before the first slice and
after each one, and the reasoning is recorded on the function so nobody reintroduces the
throttle as an optimization.

Your note that `readfrom_deadline_test.go` uses `step=0` and so never exercises the
production throttle was the important part — that is why this survived. Two tests added
that do, one with `step` longer than the whole transfer. Both fail on `f6c8b04c`.

internal/httpstream/rolling_deadline.go
Quick104: **Fixed** in `6ec97c1`, together with the sibling finding above.

You called this latent, and it was, but it stopped being hypothetical once the throttle was
removed from the slice path: the fix for the floor and the fix for the first slice are the
same one-line change. `ReadFrom` now calls `forceBump()` before entering `CopyChunked` as
well as after each slice, restoring the pre-`CopyChunked` guard exactly.

`TestReadFromBumpsBeforeTheFirstSlice` builds the case you describe — headers committed,
then a wait longer than the window, then the first write — and fails on `f6c8b04c`.

internal/proxy/egress.go
Quick104: **Fixed** in `5a9e6e7`.

Confirmed, and the consequence you trace to `nodepool.Planner.effectiveEgressKbps` is the
part that matters: under-reported egress means over-admitting onto a saturated proxy.

Kept `ReadFrom` — hiding it again would forfeit the sendfile repair that is one of this
branch's four prerequisite fixes — and instead sized the metered slice for the meter's own
resolution. `internal/proxy.meterChunk` is 256 KiB, so a 200-500 kbit/s viewer credits
every 4-10s, well inside the 60s window, while still handing the kernel 8x more per
sendfile call than the 32 KB `Write` path this replaced.

Slice size here is a rate-fidelity constraint rather than a tuning knob, which is now
recorded on the constant next to the same warning about `readFromChunk`.
`TestMeteredResponseWriterReadFromCreditsIncrementally` pins the granularity.

internal/streamtelemetry/registry.go
Quick104: **Fixed** in `fc46ad8`.

Marked PLAUSIBLE, but it holds — and the `RequestCount` detail makes it plainer than the
capacity math alone: a field that exists to count requests per transfer was structurally
pinned at 1, because each request got its own record.

A transfer is now one *pour* — one subject moving one media file over one route — keyed on
`(subject, profileID, mediaFileID, method, pattern)` with a map of in-flight observations,
rather than one record per `obs.id`. Overlapping Range GETs fold into it, `RequestCount`
becomes meaningful, and the ABS scenario you describe no longer exhausts `MaxTransfers`.
Bounded per-transfer by `MaxObservationsPerSession`, mirroring sessions.

Two tests: 25 ranged requests fold to one record with `RequestCount=25` and one
reservation, and a different file or subject still separates.

internal/streamtelemetry/global.go
Quick104: **Documented, not changed** — `fc46ad8`.

The asymmetry is real and I am not disputing it, but I do not think it is fixable from a
single sample, so I would rather state the limit than add a check that appears to handle it.

The roster score *is* the publisher's own `CapturedAt` (`store_redis.go:151`), so a
publisher running behind has its heartbeat and its snapshot drift together — there is no
independent clock to compare against, and it is genuinely indistinguishable from one that
stopped publishing. Classifying it stale is the safe direction: the alternative is serving
minute-old sessions as current.

Your observation that `Sequence`/`Epoch` are recorded but unused is the actual answer.
`BuildGlobalView` is a pure function of one sample and cannot use them, but they are
exported on `PublisherStatus` precisely so two successive reads of the parity endpoint
distinguish "behind but advancing" from "stalled". That is now written at the check rather
than left implicit.

Genuinely fixing it needs a skew estimate from a shared clock — a Redis `TIME` sample per
publish, most likely — which is a design change, not a patch to this branch.

internal/streamtelemetry/writer.go
Quick104: **Documented, not changed** — `fc46ad8`.

Agreed on the analysis, including that it is latent today: nothing calls `cut.Store`, and
`doc.go` declares the package observational.

I deliberately did not half-fix it. Adding a per-slice `cut` check here would still only act
at 4 MiB granularity, so h1 and h2 would still disagree — just less visibly, which is worse
than a documented gap. Your last line is the right disposition: both halves land together.

So it is now a stated requirement on the enforcement change that introduces the first
`cut.Store` caller — it owns making the granularity uniform, and must land with a test that
a cut behaves identically over h1 and h2. Recorded on `observedWriter.ReadFrom` itself
rather than only in the design, so whoever writes that caller reads it.

internal/proxy/server.go
Quick104: **No change** — checked, and I believe this is already handled.

The `iat` fallback is real, but it is labeled rather than silent, which I think answers the
concern. `sessionInfo` records `StartedAtSource` (`server.go:457`), it reaches
`nodesessions.SessionInfo` as the `started_at_source` JSON field
(`tracker.go:36`), and on the telemetry side `StartedAtSourceIssuedAt` sets
`startedDegraded` (`session.go:186`), so a consumer can tell an `iat`-derived start time
from a claim-derived one without inferring it.

You already note it is bounded to legacy tokens and that `compareNodeSessions` moves with
it. Given that, a rolling deploy showing a `started_at` mismatch against Postgres for
un-upgraded tokens is the labeled, expected behavior of the `ostn` migration rather than a
defect — and the label is what makes it diagnosable.

Happy to revisit if you would rather the proxy refuse the `iat` fallback outright and record
first-seen instead, but that trades a labeled-imprecise value for a definitely-wrong one.

internal/streamtelemetry/writer.go
Quick104: **Fixed** in `813509d6`.

Adopted your suggested shape almost exactly: `httpstream.ForwardReadFrom(inner, self, src,
chunk, record)`. All nine sites collapse to a one-line call, each keeping its own chunk
size and record callback, so behavior is unchanged.

The failure mode you describe is the reason it is worth doing — a missed site silently
reverts to the `io.Copy` fallback and loses both sendfile and that wrapper's accounting,
with nothing failing. `grep ReaderFromOf` outside `internal/httpstream` now returns
nothing, so the pattern has one implementation rather than nine.

internal/api/media_routes.go
Quick104: **Half fixed** in `f25ef9b5` — and I want to be explicit about which half.

The viewer-IP duplication is gone. `streamtelemetry.ViewerIP` is exported and the three
inline copies in `api`/`jellycompat`/`proxy` are deleted, so the IPv6-handling scenario you
describe now has one place to fix. That was the part with a live correctness cost:
divergence there means families reporting different viewer addresses into the same merged
view.

The `xRoute`/`declareXMediaRoutes`/lookup/`observeX` skeleton is **not** deduplicated. On a
121-file PR that is already a production streaming change, restructuring the enrolment
contract across five packages seemed like the wrong risk to take on at this point — the
manifest tests would catch a mistake, but it widens the diff in exactly the area the branch
asks reviewers to trust.

I think it should land before a sixth family is added, which is when the copy-paste starts
costing something. Happy to open an issue under #265 for it, or to do it here if you would
rather it not ship duplicated.

internal/streamtelemetry/config.go
Quick104: **Fixed** in `f25ef9b5`.

Took the shared-helper suggestion: new `internal/envutil` with `Truthy(value)` and
`Bool(name)`, adopted here and in the sibling `internal/telemetry`, whose local `truthy`
is deleted.

One behavior change worth flagging rather than burying, since you noted the accepted
spellings differ between copies: `envutil.Truthy` accepts the **superset**, including
`enabled` from `ebook_conversion.go`. So `SILO_STREAM_TELEMETRY_ENABLED=enabled` and
`SILO_OTEL_ENABLED=enabled` now turn on where they were previously silently off. That is
the point of unifying, but it is live behavior on two flags and someone should agree with
it deliberately.

The other seven copies are untouched — migrating them is unrelated to this branch. The
package doc says new flags belong here and the rest should migrate as the code around them
is touched, so it stops spreading.

internal/streamtelemetry/codec.go
Quick104: **Fixed** in `f25ef9b5`.

Exactly as suggested: `checkVersion` now takes the already-decoded `V` field instead of the
raw bytes, and each decoder unmarshals once then checks. No behavior change — an
unsupported version still returns `errUnsupportedCodecVersion` before any validation runs.

Your point about where it lands is the reason it is worth doing: these run per record from
`LoadAll` feeding `BuildGlobalView`, which is the ~347 ms path at the 50 000-session cap,
and nearly all of that is decode.

cmd/silo/main.go
Quick104: **Fixed** in `f25ef9b5`.

`Registry.ViewTTL()` exposes the resolved value, and `newStreamTelemetryViewCache` reads it
off the registry that already parsed the config instead of calling `ConfigFromEnv` a second
time — so it no longer takes `nodeID` at all.

Both consequences you name are gone: an invalid variable is logged once, and the two calls
cannot disagree.

internal/api/handlers/playback.go
Quick104: **Fixed** in `f25ef9b5`.

Deleted `playbackClientInfoFromRequest` *and* `PlaybackClientInfoFromRequest`, rather than
keeping one of the two. Every caller now goes straight to `playback.ClientInfoFromRequest`,
which is what `proxy/media_routes.go` was already doing — so there is one name instead of
three, and it is the one that owns the behavior.

Touched `playback_v3.go` (2 calls), `api/media_routes.go` (1) and the two assertions in
`playback_sessions_test.go`.

internal/api/router.go
Quick104: **Fixed** in `88451eeb`.

You are right that the precedent is weak — the ~20 sibling admin routes are undocumented
too — but "the rule was not formally satisfied or ruled out" is fair, and it was easier to
satisfy than to argue.

Added `docs/admin-api.md`, scoped honestly to what it actually covers rather than
pretending to document the whole admin surface, with the full response shape for the parity
endpoint: `view` completeness fields, per-source `report` fields, and the caveats needed to
read one correctly — including that a single report samples three independently updated
stores, so one-sided differences are expected and are not on their own evidence of a
defect. Plus a `docs/feature-changelog.md` entry.

The clock-skew row cross-references the limitation from the `global.go` thread, so an
operator reading `clock_skew_suspected: false` knows what it does not cover.

Quick104:
## Conflict resolved, CI green for the first time since 18 August

`d116d492` — `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, all three CI jobs pass
(Go, Web, Docs hygiene).

Worth stating plainly, because the PR description's verification section predates it: **CI
had never passed on this branch.** The only prior run was 18 August on `f6c8b04c`, and it
failed the Web job on
`useIntroSkipPrompt.test.ts(49,11): error TS2322` — a file this branch does not contain. It
arrived from main via #660, CI builds the *merge* commit, and main fixed it in `d285d0c2`
about thirteen hours later. Because the branch was `dirty` from then on, GitHub could not
compute a merge ref and `pull_request` CI had nothing to check out, so no run was ever
attempted again. Merging main both unblocks the button and picks up that fix.

### The three resolutions

**`docs/design/2026-07-09-streaming-write-deadline.md` — accepted main's deletion.**
#675 pruned the shipped design artifacts after distilling their durable content into
`docs/architecture/`. `a1ba4d64` does that half for this document: the invariants a future
change must respect — the rolling-deadline contract, why slice size is a correctness
constraint rather than a knob, the two rules every `ResponseWriter` wrapper on a media
route must follow, the one-limiter sendfile trap, why chi's compressor is bypassed rather
than repaired, and how conformance is actually verified — now live in
`docs/architecture/streaming-write-deadline.md`. The one-shot half (the 2026-07-09
debugging session, the per-file application table, the rollout plan, the silo-apple
follow-up list) goes with the deletion. The stream-telemetry design's `Related:`
cross-reference is repointed; no dangling references remain.

It also picks up the two rules this review turned up: the bump throttle belongs to `Write`
and never to a `ReadFrom` slice, and the proxy egress meter has the same shape of
constraint at a different value.

**`docs/feature-changelog.md`** — ordering only, both sides kept.

**`internal/api/testdata/media_routes.txt`** — main added `GET /api/v1/api-keys/scopes`
(#649's scoped API keys), and the manifest guard refused to accept it unclassified. That is
the guard doing exactly what it was built for, on a route from outside the branch. Checked
against `HandleListAPIKeyScopes` — it writes a small JSON scope catalog and serves no media
bytes — and pinned as `non-media`.

### Verification on the merged tree

Every CI step run locally before pushing, including the frontend ones the PR description
lists as not run (pnpm is available on this host):

- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal` — clean
- `go test ./...` — passes except `TestBeginWebOperationRecoversDeadProcessLock` and
  `TestBeginWebOperationRejectsLiveProcessLock`, which I ran against `origin/main` in a
  clean worktree and which **fail there identically**. macOS-only; the Linux CI Go job
  passes. `TestResolveCopySeekAnchorMatchesRealLongGOPHEVC` passes here — this host has a
  newer ffmpeg than the one in the description.
- `golangci-lint run --new-from-merge-base=origin/main` — 0 issues
- `pnpm run lint` — 0 errors (154 pre-existing warnings), `pnpm run format:check` — clean,
  `pnpm run build` — typecheck passes
- `make test-web` — 283 files, 2023 tests, all pass
- `make verify-settings-bindings-web`, `make verify-local-paths` — clean

### Review threads

All 23 have per-finding replies; 22 are resolved. The one left open is the media_routes
scaffolding duplication, because it is the only finding **half** addressed and the reply
asks you a question rather than reporting a fix: the viewer-IP duplication is gone
(`streamtelemetry.ViewerIP` exported, three copies deleted), but the
`xRoute`/`declareX`/lookup/`observeX` skeleton is still copy-pasted across five packages. I
judged a five-package restructure of the enrolment contract to be the wrong risk to add to
this diff, but it should land before a sixth family exists. Your call whether that is an
issue under #265 or work for this PR — I left the thread open so it is not lost either way.

Three judgment calls from the earlier fixes are still worth a deliberate ack rather than
silent acceptance: `envutil.Truthy` now accepts `enabled`, so
`SILO_STREAM_TELEMETRY_ENABLED=enabled` and `SILO_OTEL_ENABLED=enabled` turn on where they
were previously off; the config repair auto-*lowers* `SweepInterval` when only `FRESHNESS`
is pinned low, which means more Redis writes; and `drop("pending realtime capacity
exhausted")` increments `droppedObservations` for something that is not an observation.

### What is left

Nothing mechanical. Branch protection requires no approvals and no status checks, so this
is mergeable now; what remains is a human review of the fixes, and a decision on the
scaffolding follow-up.

### AI-use disclosure

- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: AI-assisted. Every command above was executed and its output read; the two
  pre-existing Go failures and the CI-blocking frontend failure were each reproduced
  against `origin/main` in a clean worktree before being characterized as not this
  branch's.
- Adversarial review: I initially reported the conflict as "just simple docs" and was about
  to resolve it as a one-file deletion. Checking the actual CI history first is what
  surfaced that the branch had never had a passing run at all, and that the merge was
  load-bearing for reasons unrelated to the conflict — the manifest guard then caught a
  second, real issue in the same merge. Treating the conflict as trivial would have shipped
  an unclassified media route past a guard this PR exists to establish.

PR #668, opened 2026-08-18
fix(metadata): stabilize image cache queue

blurbery:
## Problem

No linked issue; this was reproduced and verified on a real Docker deployment before the upstream submission.

**Cache Metadata Images** had no reliable durable progress model. While it was running, the task could look stuck at 0%, and the available processing counters described only the current execution rather than the persistent queue. Admins could not tell how much artwork was actually complete or whether the backlog was converging.

The queue also treated `failed` as reusable work. Once `next_attempt_at` elapsed, unchanged failed rows could be moved back to `queued` by both enqueue conflicts and catalog discovery. Provider jobs whose resolver returned an empty URL therefore consumed the normal retry sequence, and rows that had already exhausted their attempts could return in later scans with the same source.

In practice this caused:

- a task that appeared stuck or never-ending even while useful work was being completed;
- repeated provider resolution/download attempts for images already known to be unusable;
- avoidable database updates, catalog discovery, provider traffic, cache activity, and CPU load;
- misleading use of “failed” for both a failed attempt and a genuinely terminal job;
- no stable tombstone preventing invalid/missing provider artwork from re-entering the queue.

## Approach

- Read exact queue-wide counts from `metadata_image_cache_jobs` and report cumulative run stats plus durable completed/total counts after each claim or discovery batch.
- Keep an active execution below 100% while final discovery can still add work, then let the task report the authoritative execution completion at 100%.
- Show indeterminate “Running” UI at 0%, preserve fractional percentages, and display the progress message alongside the number in all admin task views.
- Mark an empty resolver result terminal on its first failed attempt. Leave terminal failures parked for an unchanged source while preserving source-change recovery.
- Distinguish **failed attempts** in the current execution from **terminal failures** in the persistent queue.
- Reduce cache concurrency from 12 workers to 2. The 1,000-job claim size and 10-minute runtime cap remain unchanged.

There is no schema migration and this change deletes no media, artwork, library, or content records. Existing terminal rows are deliberately retained as deduplication tombstones.

## Real-deployment validation

`@blurbery` reproduced the failure and validated the corresponding fix on a real Docker deployment with a large persistent image queue. The task converged without deleting media, artwork, library, or content records; invalid/missing image failures remained as terminal deduplication records instead of returning as an endless active backlog.

The operational cleanup and deployment procedure are not part of this PR. Production identifiers and queue metrics are intentionally omitted. This PR reapplies the code fix as one clean commit on current upstream `main` and includes the corrections from the adversarial review below. It does not change the code's default trigger interval.

## Testing

Passed locally:

```text
pnpm exec vitest run src/lib/taskProgress.test.ts src/components/ServerActivity.test.tsx
Test Files  2 passed (2)
Tests       3 passed (3)

pnpm exec vitest run <repository exclusions from make test-web>
Test Files  280 passed (280)
Tests       1961 passed (1961)

pnpm run format:check

pnpm run lint
0 errors (the repository emitted 151 warnings)

make verify-local-paths
passed

git diff --check upstream/main...HEAD
passed
```

The fork's manual run of the upstream CI workflow also completed on this exact commit:

```text
Go
passed: build, gofmt, vet, changed-line golangci-lint, settings bindings,
playback fixtures, and the full make test-go suite

Docs hygiene
passed

Web
passed: install, lint, and format check
stopped: Typecheck and build at the upstream-main failure described below
```

CI run: https://github.com/blurbery/silo-server/actions/runs/32124185565

Local limitations, reported rather than hidden:

- The targeted local Go test command reached Silo's native `bimg` dependency, but this Mac does not have `libvips`; the Linux CI Go job above supplies the authoritative native result and passed completely.
- `pnpm run build` stops at `src/player/hooks/useIntroSkipPrompt.test.ts:49` with TS2322. That file is unchanged by this branch and byte-identical to `upstream/main`. Current upstream-main CI fails at the same Web step: https://github.com/Silo-Server/silo-server/actions/runs/32050073543
- No screenshot is attached because the production task had completed, and a production cache run was not triggered solely to manufacture a screenshot. The running/percentage presentation is covered by the component tests above.

## Risks and follow-up

- A terminal row no longer retries automatically for the same source. This is intentional; changing the source requeues it, and regression coverage protects that recovery path.
- The queue snapshot is one aggregate query per 1,000-job claim/discovery batch. Real-deployment validation did not leave a persistent active backlog, but maintainers may want to add telemetry if substantially larger queues expose count-query pressure.
- Two workers trade peak throughput for safer provider, database, and cache load. The execution remains bounded by the existing 10-minute runtime.

### AI Disclosure
- Model(s): GPT-5 (the exact backend model ID was not exposed to this task)
- Involvement: AI-assisted. `@blurbery` identified and reproduced the bug, designed and implemented much of the fix, chose the safety limits, approved and performed the operational cleanup and deployment, and verified the live result. Codex assisted with isolating the upstream patch, refining code and tests, adversarial review, and PR preparation.
- Adversarial review: The post-commit review found that the original fork branch carried unrelated history, so the change was rebuilt as one commit on current upstream `main`. It also found ambiguous failure wording, incorrect singular grammar, an empty-URL terminal record that overstated its attempt count, and missing coverage for source-change recovery; all four were corrected. No material unresolved finding remains. The repository's companion-review script was not installed, so the documented review rubric was applied directly to the complete `upstream/main...HEAD` diff.

## Checklist

- [x] I ran the repo verify commands locally and through the repository's Linux CI workflow: frontend lint/format and tests passed, and the Go build/gofmt/vet/changed-line lint/full test suite passed. The only failing gate reproduces the current upstream-main Web build failure documented above.

## Summary by CodeRabbit

* **New Features**
  * Improved image-cache job recovery with automatic retries, cooldowns, and permanent parking for stable failures.
  * Re-admits eligible failed jobs after their retry period or when source artwork changes.
  * Added backlog-based image-cache progress reporting with startup, ongoing, and completion updates.
  * Enhanced task progress displays with bounded percentages, fractional values, and indeterminate “Running” states.

* **Bug Fixes**
  * Prevented progress from incorrectly reaching 100% before work is complete.
  * Improved handling of expired jobs and failed image-cache requests.


Quick104:
## Pushed review fixes to this branch (`3c8c1c95`)

I reviewed this branch, found six defects, and pushed fixes rather than leaving them as comments. Happy to drop the commit if you'd rather take them yourself.

**What changed**

1. **Empty resolved URL was terminal on attempt 1** (`image_cache_job_repo.go:136`). `classifyImageCacheFailure` short-circuited on `imageCacheEmptyResolvedURLError`, but `PluginImageResolver` returns `""` for transient conditions too — no source registered for the scheme while `reloadPluginImageResolvers` swaps sources, a plugin disabled or upgrading, a swallowed gRPC error from `resolvePluginBatch`, a nil presigner. Disabling TMDB for two minutes would tombstone every poster the sweep touched. Now takes the ordinary retryable path.

2. **Attempt-exhausted rows were unrecoverable** (`:326`, `:832`). With the `status='failed'` branch gone from both the upsert and discovery, a row that merely burned its 8 attempts during an S3 outage had no way back — a later refresh writes the *same* `source_path`, so the source-changed branch never fires, and `requeueSucceeded` only revives `succeeded`. `next_attempt_at` now carries the distinction: `isStableProviderImageFailure` (403/404/410/418, local-file errors) parks permanently and stays a dedup tombstone, everything else parks for a cooldown and is re-admitted after it. Same for lease-expiry exhaustion in `recoverExpiredRunning` — repeated expiry means workers kept dying, which says nothing about the artwork.

3. **Progress described the durable queue, not the run** (`cache_metadata_images.go:130`). `Completed/Total` recomputed per batch while discovery added up to 1000 rows and retention deleted up to 1000, so it oscillated ~50%↔99.9%, and a steady-state server opened at ~99.9% before doing any work. Now measured against the run.

4. **`GetQueueProgress` was an unindexed five-way aggregate on the hot path** (`:88`). Replaced with `GetBacklog` — two scalar counts served by the existing partial status indexes, sampled once per run instead of once per batch.

5. **`toFixed(1)` rendered "100.0%" on a running task** (`web/src/lib/taskProgress.ts:9`). Rounds before comparing now.

6. **The regression test grepped source instead of asserting behavior** (`image_cache_job_repo_test.go:78`). It rejected the literal `j.status = 'failed'` anywhere in the file, which would have blocked fix 2 and would fail on any future admin retry query. Replaced with behavioral tests plus a DB-backed test for the re-admission gate.

**Worth your attention**

- **A first deploy will look busy.** Every existing `failed` row has `next_attempt_at` in the past, so the first discovery sweep re-admits all of them. Bounded by `claimLimit` and the runtime budget, so not a herd, but expect one unusually active run.
- **Genuinely-missing artwork now costs 8 resolver attempts per cooldown instead of 1.** That is the price of not being able to distinguish "provider has no image" from "resolver unavailable" at this layer. The real fix is making that distinction explicit across `catalog.ResolvedImageURL` and the plugin gRPC contract — worth a follow-up issue, and I've left a comment pointing at it rather than pretending it's solved.
- **The SQL half of fix 2 is only exercised by a `SILO_TEST_DATABASE_URL`-gated test**, which CI does not set. I read all seven `DO UPDATE` CASE arms, the upsert `WHERE`, and the discovery filter by hand and they're consistent, but one live run against a database before merge would be worth it.

`go build ./internal/...`, `go vet`, and the `metadata` + `taskmanager` suites are clean; web lint, `format:check`, and the affected vitest files pass. `go build ./...` fails on the missing `web/dist` embed, which is environmental and unrelated.

Note this branch and #676 both rewrite `image_cache_job_repo.go`. They merge textually clean but pull in opposite directions — this one decides what stays parked, #676 adds a per-target revival path. Worth landing them in a deliberate order.

<sub>Review and fixes produced with AI assistance (Claude), verified by building and running the branch. See [docs/ai-contributions.md](docs/ai-contributions.md).</sub>

PR #669, opened 2026-08-18
fix(metadata): back off failed person refreshes

blurbery:
### Summary

- persist a dedicated `metadata_refresh_attempted_at` timestamp before person provider lookups
- select scheduled candidates by the newer of their metadata update and last provider attempt
- add a concurrent partial expression index for the scheduled query
- enqueue fresh incomplete people when their detail page is viewed
- recover safely if an interrupted concurrent build leaves the new index invalid

### Root cause

Incomplete people were always immediately eligible, while a failed provider lookup did not change any durable state. The worker therefore selected the same unresolved records every 10-minute cycle. The detail handler also returned early for incomplete people, so it skipped the useful on-demand refresh path.

### Impact

Provider misses and partial responses now receive the same 90-day schedule as successful lookups without changing `updated_at` semantics. The migration is additive, builds its index concurrently, and includes the repository's standard invalid-index recovery guard. Fresh incomplete people remain eligible on demand through the person detail route.

### Verification

- `make migrate-validate`
- Go CI: build, gofmt, vet, changed-line lint, generated bindings/fixtures, and full `make test-go` passed: https://github.com/blurbery/silo-server/actions/runs/32135940645
- multi-architecture production-base image build passed: https://github.com/blurbery/silo-server/actions/runs/32136842786
- deployment validation: migration applied, the index was valid and selected by the exact candidate query, and the worker remained healthy across a scheduled interval
- upstream `main` already has an unrelated Web typecheck failure in `useIntroSkipPrompt.test.ts`; its Go job passes at the base commit: https://github.com/Silo-Server/silo-server/actions/runs/32050073543

### Adversarial review

- verified failed lookups are marked before provider I/O, including timeout/error paths after the person is loaded
- verified successful updates retain their normal `updated_at` behavior
- verified `staleAfter <= 0` retains the prior incomplete-person selection semantics
- verified interrupted concurrent index creation cannot strand an invalid index that blocks retry
- verified the patch applies cleanly both to upstream `main` and to an independently built deployment branch

### AI Disclosure
- Tool(s): Codex
- Model(s): GPT-5
- Involvement: AI-assisted; the user identified the problem, proposed the fix, and actively guided diagnosis, implementation decisions, and live validation
- Adversarial review: Reviewed retry/error paths, query/index equivalence, migration interruption recovery, deployment-base preservation, and live post-deploy behavior; added the invalid-index recovery guard found during that review.

## Summary by CodeRabbit

- **New Features**
  - Improved person metadata refresh scheduling for incomplete and stale records.
  - Added retry backoff tracking to prevent repeated provider lookups too soon.
  - Refresh attempts are now recorded, including unsuccessful provider lookups.

- **Bug Fixes**
  - Corrected selection of people eligible for metadata refresh.
  - Refresh processing now continues when recording an attempt encounters an error.
  - Preserved person update timestamps while tracking refresh attempts.

### DIFF
--- a/internal/api/handlers/people.go
+++ b/internal/api/handlers/people.go
@@ -368,11 +368,7 @@ func (h *PeopleHandler) enqueuePersonRefreshIfDue(person models.Person) {
 		return
 	}
 
-	if personMetadataIncomplete(person) {
-		return
-	}
-
-	if person.UpdatedAt.Before(time.Now().Add(-personMetadataStaleAfter)) {
+	if personMetadataIncomplete(person) || person.UpdatedAt.Before(time.Now().Add(-personMetadataStaleAfter)) {
 		h.refreshQueue.Enqueue(person.ID)
 	}
 }
--- a/internal/api/handlers/people.go
+++ b/internal/api/handlers/people.go
@@ -39,8 +39,6 @@ var personRefreshRate = ratelimit.Rate{
 	Burst:             10,
 }
 
-const personMetadataStaleAfter = 90 * 24 * time.Hour
-
 // PeopleHandler serves person-related API endpoints.
 type PeopleHandler struct {
 	personRepo      peopleRepository
@@ -363,22 +361,18 @@ func (h *PeopleHandler) toResponse(ctx context.Context, p models.Person) personR
 	return resp
 }
 
+// enqueuePersonRefreshIfDue queues an on-demand provider lookup for a person
+// whose detail page was just viewed. It shares catalog.PersonRefreshDue with
+// the background sweep, so a person who was looked up recently is not sent to
+// the provider again on every page view.
 func (h *PeopleHandler) enqueuePersonRefreshIfDue(person models.Person) {
-	if h.refreshQueue == nil || !personHasRefreshableProviderID(person) {
+	if h.refreshQueue == nil {
 		return
 	}
 
-	if personMetadataIncomplete(person) || person.UpdatedAt.Before(time.Now().Add(-personMetadataStaleAfter)) {
+	if catalog.PersonRefreshDue(person, time.Now()) {
 		h.refreshQueue.Enqueue(person.ID)
 	}
-}
-
-func personHasRefreshableProviderID(person models.Person) bool {
-	return person.TmdbID != "" || person.ImdbID != "" || person.TvdbID != ""
-}
-
-func personMetadataIncomplete(person models.Person) bool {
-	return person.Bio == "" || person.PhotoPath == "" || person.PhotoPath == "-" || person.BirthDate == nil
 }
 
 func parseOptionalPersonDate(raw string) (*time.Time, error) {
--- a/internal/api/handlers/people_refresh_test.go
+++ b/internal/api/handlers/people_refresh_test.go
@@ -1,0 +1,87 @@
+package handlers
+
+import (
+	"testing"
+	"time"
+
+	"github.com/Silo-Server/silo-server/internal/models"
+)
+
+type recordingPersonRefreshQueue struct {
+	ids []int64
+}
+
+func (q *recordingPersonRefreshQueue) Enqueue(id int64) {
+	q.ids = append(q.ids, id)
+}
+
+func TestEnqueuePersonRefreshIfDue(t *testing.T) {
+	now := time.Now()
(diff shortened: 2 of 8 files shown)


Quick104:
## Pushed review fixes to this branch (`81f80c74`)

I reviewed this branch, found five defects, and pushed fixes rather than leaving them as comments. Happy to drop the commit if you'd rather take them yourself.

**The mechanism here is right — the policy on top of it wasn't.** The new column, the expression index, and putting `MarkRefreshAttempt` before the provider call are all correct, and I verified the migration by applying it to a scratch Postgres 18: goose conventions match, `GREATEST`/`COALESCE` are immutable so the expression index builds, the nullable column needs no backfill, and `EXPLAIN` on a 200k-row table picks `idx_people_metadata_refresh_due` with no sort node. I did not touch it.

**What changed**

1. **The sweep stopped enriching new people entirely** (`person_repo.go:1015`). Dropping the incompleteness predicate left selection on `GREATEST(updated_at, COALESCE(attempted_at, updated_at)) < now()-90d`. But people are created by `BatchFindOrCreate` from credits, and `plugin_provider.go:661-668` sets only name, provider ids, and photo path — never bio, birth date, birthplace, homepage. So every person is born incomplete with `updated_at = now()` and can never satisfy that. Nothing enqueues at scan time either; the only sources are the detail page, the manual endpoint, and this sweep. Scan a new library and 4,000 people get no bio for 90 days, then arrive all at once. The predicate is back, gated on the new column rather than `updated_at`, so a never-looked-up person is eligible immediately and a person the providers can't complete is retried weekly instead of every sweep.

2. **The on-demand path had no backoff and no rate limit** (`people.go:371`). `models.Person` had no attempt timestamp, and `MarkRefreshAttempt` deliberately doesn't touch `updated_at`, so `enqueuePersonRefreshIfDue` had nothing to read and enqueued unconditionally. With `photo_path = '-'` — the persisted "provider has no photo" sentinel — `personMetadataIncomplete` was true *forever*, so 200 page views meant 200 TMDB round trips. Meanwhile the sibling `POST /people/{id}/refresh` is rate-limited at 10/min precisely because it causes provider I/O. The column is surfaced on the model, `'-'` now counts as an answer rather than a gap, and one shared definition of "due" backs both the SQL sweep and the Go handler so they can't drift.

3. **`MarkRefreshAttempt` could veto the refresh** (`person_refresh.go:112`). Any non-`ErrNoRows` error returned early, so a pool hiccup aborted a lookup that would have succeeded *and* left the person immediately re-eligible — the exact hot loop this branch fixes, restored for the duration. It logs and continues now. The write stays before provider I/O; that ordering is what makes the backoff durable across a crash.

4. **The `staleAfter <= 0` branch was unreachable** (`person_repo.go:991`). `main.go:1519` wires a fixed 90 days with no env or settings override and `NewPersonRefreshWorker` only clamps negatives, so nothing could reach the safety net the PR body cites. Removed.

5. **`people_refresh_test.go:28` pinned the bug.** The `"incomplete person with provider id"` case asserted `want: true` for a person with `UpdatedAt: now`, i.e. it protected the unconditional enqueue and would have blocked fix 2. Updated to distinguish "incomplete and not recently attempted" from "incomplete and attempted 30 seconds ago".

**Worth your attention**

- **Sort order deprioritizes exactly the people fix 1 is for.** `ORDER BY GREATEST(updated_at, COALESCE(attempted, updated_at)) ASC` puts freshly-ingested never-attempted people *last* among due candidates, since their `updated_at` is now. Only observable when due rows exceed the 100/tick batch, and eventual coverage is unaffected. I left it deliberately: ordering by `metadata_refresh_attempted_at` instead would match the new gate more naturally, but it would strand the expression index this PR adds. Your call whether that trade is worth revisiting.
- **The steady-state sweep has no bound on the ORDER BY expression**, so when nothing is due it walks the whole partial index every 10 minutes rather than range-scanning to a cutoff. Not a regression against `main`, which had no index at all.

`go build`, `go vet`, `gofmt`, and the `catalog` / `metadata` / `api/handlers` / `worker` suites all pass. `golangci-lint` isn't installed here, so CI covers that.

<sub>Review and fixes produced with AI assistance (Claude), verified by building and running the branch and by applying the migration to a real Postgres. See [docs/ai-contributions.md](docs/ai-contributions.md).</sub>

PR #670, opened 2026-08-18
fix(playback): let original_http carry a client-selected audio track (client_audio_track_selection_v1)

Quick104:
## Problem

For an Apple TV client that advertises Dolby Vision, the V3 planner returned `server_remux_progressive` / `container_normalization` for DV Profile 5 and Profile 8 HEVC MKVs, while an HDR10 HEVC MKV with the same request got `original_http` / `validated_original_playback`. Reproduced on shared dev and production. Downstream on tvOS: the progressive remux is a live fMP4 pipe without byte-range support and AVPlayer fails it every time (`-11850` / `-12939`), then the HLS remux strips DV to HDR10.

**Root cause is not Dolby Vision.** The profile's audio preference (`en`) resolves to a **non-default** audio track on the foreign-release DV titles, and the direct branch requires the container default (`audioSelectionUsesContainerDefaultV3`, `internal/playback/plan_v3.go:341`). The HDR10 title had no English track, so the preference fell back to the default and passed. Evidence: `playback_v3_attempts` for the captured probe attempts shows `selected_tracks.audio.index` = 1 / 1 / 0 ↔ remux / remux / original.

## Change

Capability-scoped rather than a blanket removal (a browser cannot select an audio track and must keep getting the remux):

- `protocol_v3.go`: new per-delivery feature `client_audio_track_selection_v1` — advertised by a delivery whose executor selects the plan's audio track itself.
- `plan_v3.go`: the direct branch uses `originalAudioSelectionRealizableV3` = container default **or** the `original_http` delivery advertises the feature.
- `docs/architecture/playback-protocol-v3.md` §4: contract paragraph.
- Tests: `TestPlanPlaybackV3AppleOriginalHTTPHonorsClientAudioTrackSelection` (real dev rows for DV P5 / DV P8 / HDR10 + the captured tvOS request, sanitized fixtures under `internal/playback/testdata/apple_tvos_dv_*.json`) and `TestPlanPlaybackV3OriginalHTTPKeepsRemuxWithoutAudioSelectionFeature`; the pre-existing `TestPlanPlaybackV3NonDefaultAudioSelectionCannotUseOriginalHTTP` still passes untouched.

Inert until a client advertises the flag. silo-apple advertises it in `player/architecture-remediation` (and routes non-default audio to its loopback executor); Android honours `selected_tracks.audio.index` too and can add the same line.

## Validation

- `go test ./internal/playback/... ./internal/api/...` green (branch merged with current `main`).
- Deployed to shared dev and replayed the captured request: DV P5 / DV P8 → `original_http` / `validated_original_playback` (audio idx 1) with the feature; unchanged legacy remux without it; HDR10 unchanged. 9/9 scenarios matched.

## Follow-ups (not in this PR)

- Two same-shaped sibling gates still require the container default: `plan_v3.go:297` (Profile 7 client-transform direct branch) and `:573` (audio-only route).
- `server_remux_progressive` is never playable by AVPlayer on tvOS/iOS (live fMP4 pipe, no Range); the Apple client now advertises `progressive` as `supported_on_device: false`, so the planner falls to HLS remux. No server-side gate added on purpose (platform-neutral contract).
- Data note: 3 P8 rows on dev lack `dv_bl_compat_id` and conservatively lose the P8→HDR10 strip; 23 rows carry compat id 6 (possibly over-conservative to exclude).

## Summary by CodeRabbit

- **New Features**
  - Original HTTP playback can now honor explicitly selected non-default audio tracks when supported by the client.
  - Audio-track selections remain honored after playback replanning.
  - Unsupported audio-track selections continue using remuxed playback.

- **Bug Fixes**
  - Improved direct playback behavior for Apple TV Dolby Vision and HDR10 content while preserving selected audio tracks.

- **Documentation**
  - Documented the client audio-track selection capability for source-preserving playback.

### DIFF
--- a/docs/architecture/playback-protocol-v3.md
+++ b/docs/architecture/playback-protocol-v3.md
@@ -442,6 +442,16 @@ because "the user turned HLS off" and "this device has no HLS player" call for
 different degradation warnings and different diagnostics. A class the client
 omits entirely is unavailable — the server will not guess.
 
+One `features` string is load-bearing on `original_http`:
+`client_audio_track_selection_v1` declares that this delivery's executor selects
+`selected_tracks.audio` itself out of the container it is handed. Without it, a
+source-preserving route is only planned when the selection *is* the container's
+default track — the bytes are untouched, so a client that ignores the selection
+would silently play the default, usually in the wrong language, and the server
+routes it through a remux that maps the chosen stream explicitly instead. A
+client advertising the string must honour the selection on every original
+delivery, including after a `track_change` replan.
+
 `stream.header_refresh` tells the client what to do when the stream URL's auth
 expires: `none` means the URL is stable for the session, `session` means
 re-request headers from `header_refresh_url` rather than restarting playback.
--- a/internal/playback/plan_v3.go
+++ b/internal/playback/plan_v3.go
@@ -338,7 +338,7 @@ func PlanPlaybackV3(input PlannerInputV3) PlannerResultV3 {
 	}
 
 	if source.DVProfile != 7 && deliveryAvailableV3(input.Request, DeliveryClassOriginalHTTPV3) && containerOK && videoOK && rangeOK && audioOK && quality.PreservesSource &&
-		audioSelectionUsesContainerDefaultV3(file, input.AudioTrackIndex) && !subtitle.RequiresBurn {
+		originalAudioSelectionRealizableV3(input.Request, file, input.AudioTrackIndex) && !subtitle.RequiresBurn {
 		plan := base
 		plan.Delivery = DeliveryOriginalHTTPV3
 		plan.Stream = StreamV3{Protocol: StreamHTTPProgressiveV3, Container: source.Container, MIMEType: MimeFromExtension(file.FilePath), Headers: map[string]string{}, HeaderRefresh: HeaderRefreshNoneV3}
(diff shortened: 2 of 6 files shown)


blurbery:
I implemented the two remaining `original_http` audio-selection gates called out in the review threads, based directly on this PR's current head (`f9c6561`).

Commit: https://github.com/blurbery/silo-server/commit/451bd1635bab6286e97b133dd6a280c08185d747

```bash
git cherry-pick 451bd1635bab6286e97b133dd6a280c08185d747
```

What changed:
- Profile 7 client-transform delivery now uses `originalAudioSelectionRealizableV3` for both DV7→DV8.1 and retry DV7→HDR10 plans.
- Audio-only `original_http` now uses the same delivery-scoped capability gate.
- Added regression coverage for a selected non-default audio track on both route families, including the Profile 7 retry.
- Legacy clients without `client_audio_track_selection_v1` keep the existing remux fallback.

Before the implementation, both new tests fell to `server_remux_progressive:container_normalization`; after it, they pass with `original_http` and preserve selected audio index 1. The server change is platform-neutral and applies to tvOS, iOS, and macOS clients advertising this shared Apple capability.

Validation:
- Focused regression tests: pass.
- `go build ./internal/... ./cmd/...`: pass.
- `go vet ./...`: pass.
- `go test ./internal/api/... -count=1`: pass.
- `make verify-local-paths`: pass.
- `golangci-lint v2.12.2 --new-from-merge-base=upstream/pr-670 ./...`: `0 issues`.
- Playback packages pass with the four documented macOS NVENC/GPU-probe tests excluded; those four were independently `signal: killed` in this environment.

### AI Disclosure
- Tool(s): Codex desktop, Codex CLI
- Model(s): gpt-5, gpt-5.6-sol
- Involvement: AI-assisted
- Adversarial review: the committed diff was reviewed independently with `gpt-5.6-sol`; it found no actionable issues and confirmed the delivery-scoped gate, legacy remux fallback, Profile 7 retry, and audio-only coverage.

Quick104:
Superseded by #737, which landed the same capability with a different contract shape: `client_selected_audio_track_v1` as a `validated_claims` entry on the `original_http` delivery (plus the companion `client_managed_dynamic_range_v1`), matching what the Apple client ships since Silo-Server/silo-apple#185. Keeping this open would leave two divergent tokens for the same behavior. The root-cause analysis in this PR's description remains the reference for why the container-default gate existed.

PR #673, opened 2026-08-18
feat(settings): add player.apple.control_plane rollout switch (contract rev 8)

Quick104:
Adds the remote kill-switch for the Apple client's new playback control plane — decision **P11** of the silo-apple player architecture remediation ([silo-apple PR #172](https://github.com/Silo-Server/silo-apple/pull/172)). Precedent for a paired server change: #670.

## What this is

Settings contract **revision 7 → 8**, one new key:

| | |
|---|---|
| Key | `player.apple.control_plane` |
| Type | enum `off` \| `native_direct` \| `loopback` \| `all` |
| Default | `off` (the control plane ships dark) |
| Scope | `profile_device` (matches every other remote `player.*` key) |
| Platforms | ios, tvos, macos |

Semantics — this is the client/server agreement being recorded: `off` keeps every route on the legacy player path; `native_direct` enables the new control plane only for direct-play routes; `loopback` only for loopback routes; `all` enables it everywhere. Route-family values (not a percentage) because the two families fail differently — a regressing direct route can go back to legacy without also reverting loopback. Device scope is what makes rollback targeted: one misbehaving Apple TV returns to `off` without moving the fleet, and no TestFlight release is ever needed to roll back.

The key is **hidden from the web device-settings screen** (it is an operations flag, and its `player` category would otherwise file it under Picture next to HDR) but stays flippable from the admin device-overrides view, which iterates every device-scoped key independently of the hidden set.

## ⚠️ Naming decision — merging ratifies this

`manifest.schema.json` says canonical key names *do not encode a platform*; `player.apple.control_plane` does. The exception is deliberate (rationale in the key's `notes`): what rolls out is one client's playback stack, no other platform can honour the value, and retiring it removes the key rather than generalizing it. The alternative would be `player.control_plane` with the `platforms` field alone carrying the restriction. The `player.apple.*` name is what the remediation program docs have used throughout — but this PR is the sign-off point, so if you prefer the canonical name, say so and it's a one-line change before merge.

## Changes

- `contracts/settings/v1/manifest.json` — the key, revision bump
- `contracts/settings/v1/conformance.json` — 2 new cases: `apple_control_plane_defaults_to_off`, `apple_control_plane_enrolls_one_device`
- `internal/settingskeys/keys.go`, `web/src/lib/settingsContract.ts`, `web/src/lib/settingsConformance.json` — regenerated via `make settings-bindings`, not hand-edited
- `web/src/lib/deviceSettingGroups.ts` + test — key added to `HIDDEN_KEYS` with rationale comment

## Validation

- `go test` over `settingscontract`, `settingsresolve` (both new conformance cases PASS by name), `settingskeys`, `settingsmigrate`, and the settings/device/profile API handler tests — all ok
- `make verify-settings-bindings-all` — bindings current
- `go build ./...` — clean (after the standard `make embed-stub` for a fresh worktree)
- Web: `vitest` full run 1997/2003 — the 6 failures (`SeasonContent.test.tsx`, `ServerStorageStep.test.tsx`, `ResizeObserver is not defined`) reproduce identically at base `80945139`, pre-existing jsdom polyfill gap, unrelated
- Kotlin/Swift binding steps skip when sibling repos aren't checked out beside the worktree (by design); **after this merges**, silo-apple regenerates `SettingKeys.generated.swift` on its side — never before, so clients can't drift ahead of the contract

### DIFF
--- a/contracts/settings/v1/conformance.json
+++ b/contracts/settings/v1/conformance.json
@@ -1,6 +1,6 @@
 {
   "fixture_version": 1,
-  "manifest_revision": 7,
+  "manifest_revision": 8,
   "description": "Cross-platform conformance cases for settings resolution. Every case runs against the shipped manifest in this directory: definitions are referenced by key, never restated, so an expectation can only be satisfied by resolving the real contract. Each platform's resolver (Go in internal/settingsresolve, TypeScript in web/src/lib/settingsResolve.ts, Kotlin and Swift in the client repos) runs every case through a hand-written runner; a runner must fail on any fixture field it does not know, because schema drift in the fixture itself is drift. A case's constraint_bindings attach a constraint to a copy of a real definition so constraint semantics stay testable even while no shipped definition carries that constraint kind. In expected entries, constrained:true requires stored_value and constraint_kind to be present, and stored_value may be null to mean the authored value was JSON null.",
   "cases": [
     {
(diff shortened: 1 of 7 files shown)


Quick104:
Closing unmerged per owner decision (P11 = no remote key): Stage 2 of the Apple player remediation will ship as a hard cutover rather than behind a remote kill-switch, so the settings contract stays at revision 7. Rationale and decision recorded in silo-apple docs/cleanup/HANDOFF.md §7. If a staged rollout is ever wanted after all, this branch implements the key end-to-end and can be reopened.

PR #676, opened 2026-08-19
fix(metadata): select and cache refresh artwork correctly

blurbery:
## Problem

Related to #211; this PR does not close it.

Manual Quick and Complete Refresh can finish metadata while the selected item's artwork remains queued behind the global image-cache backlog. Poster selection also has no way to distinguish a genuinely text-bearing poster from a language-tagged but textless poster, so an English library can keep the wrong artwork.

## Approach

- Add an optional, backward-compatible `includes_text` image signal and use four explicit poster tiers: preferred-language text, English text, other text, then textless. Local sidecars remain authoritative; backgrounds remain language-neutral; logos retain language-aware fallbacks.
- Give interactive refreshes a target-scoped cache path that retries failed/backed-off jobs, waits for a concurrent worker, reports cache failures, and never drains unrelated artwork. A concurrent index keeps those target operations independent of global backlog size.
- Wire both Quick and Complete Refresh through that path, including the remapped content ID after Complete Refresh.
- Replace the Web UI's loading toast in place with the final success or failure toast.

### Authorship and live verification

All of this work is mine. I tested it end-to-end in my live Silo installation and confirmed that it works.

Codex assisted with upstream cleanup, tests, adversarial review, and PR preparation as disclosed below.

## Testing

Focused Go tests:

```text
$ go test ./internal/metadata ./internal/adminjob
ok  github.com/Silo-Server/silo-server/internal/metadata  (cached)
ok  github.com/Silo-Server/silo-server/internal/adminjob  (cached)

$ go test -race ./internal/metadata ./internal/adminjob
ok  github.com/Silo-Server/silo-server/internal/metadata  2.006s
ok  github.com/Silo-Server/silo-server/internal/adminjob  2.199s
```

CI-equivalent changed-line lint:

```text
0 issues.
```

Changed Web tests and formatting:

```text
$ pnpm exec vitest run src/hooks/queries/items.test.ts
Test Files  1 passed (1)
Tests       8 passed (8)

$ pnpm exec prettier --check src/hooks/queries/items.ts src/hooks/queries/items.test.ts
Checking formatting...
```

Additional successful checks:

- `go vet ./internal/metadata ./internal/adminjob ./cmd/silo`
- `go build ./internal/... ./cmd/...`
- `make migrate-validate`
- `make verify-local-paths`
- `pnpm exec eslint src/hooks/queries/items.ts src/hooks/queries/items.test.ts`
- Full Web suite: 278 files and 1,960 tests passed
- Full Web lint: 0 errors (151 existing warnings)
- Full Web format check and production build passed

The legacy unscoped `make lint` was run and reported the documented upstream baseline of 300 findings; the PR/CI changed-line gate above is clean. A full local `make test-go` also reached unrelated macOS-sensitive failures in `internal/jellycompat` process-lock identity tests and fake-NVENC playback probes; all changed packages passed, and a transient `internal/transcodenode` timing failure passed in isolation.

A new screenshot/recording is not attached because capturing the transient toast would require triggering another metadata-changing live refresh. The behavior has already been live-tested; visual evidence can be added if maintainers require it.

## Adversarial review

The diff review found two material issues and both were fixed before publishing:

1. A background worker could release a failed target job after the interactive loop checked it, causing a false success. The loop now requeues target failures/backoff on every pass and has a regression test for the concurrent-worker handoff.
2. The first optional-target SQL predicate could prevent efficient index use on the global path. Targeted and global claims now use fixed query shapes, backed by a dedicated target/status/due index.

The final pass also checked local artwork precedence, provider ordering within tiers, background/logo fallbacks, Complete Refresh ID remapping, error propagation, and per-mutation toast identity. No material findings remain.

## Risks & follow-up

- Metadata providers must emit the canonical optional `includes_text` field to distinguish language-tagged textless posters. Providers that do not emit it retain the current language-based fallback.
- The index is created concurrently and the migration handles an invalid prior index safely.
- No REST API contract, Android client, Apple client, deployment configuration, or database data is changed.
- Coordinated TVDB and TMDB provider PRs will carry the provider-side signal.

### AI Disclosure
- Tool(s): Codex
- Model(s): gpt-5.6
- Involvement: AI-assisted
- Adversarial review: Found and fixed a concurrent-worker false-success race and a planner-hostile optional SQL predicate; re-reviewed the final diff with no material findings remaining.

## Checklist

- [x] I ran the repository verification commands and documented both passing gates and known upstream/environmental failures above.

## Summary by CodeRabbit

- **New Features**
  - Metadata refreshes now cache refreshed artwork automatically and report artwork-processing progress.
  - Artwork selection better prioritizes local images, text-bearing posters, preferred languages, and image ratings.
  - Image metadata can indicate whether artwork contains text.
  - Artwork processing retries pending work and reports failures clearly.

- **Bug Fixes**
  - Improved handling of artwork refresh jobs, including targeted retries and recovery of interrupted processing.
  - Refresh notifications now update the existing loading message with success or error results.

- **Documentation**
  - Added a changelog entry describing the updated metadata refresh and artwork behavior.

### DIFF
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -1593,6 +1593,9 @@ func main() {
 				deps.EventBus,
 				deps.RealtimeHub,
 			)
+			if metadataImageCacheProcessor != nil {
+				itemRefreshExecutor.SetArtworkCacher(metadataImageCacheProcessor)
+			}
 		}
 	}
 
--- a/docs/feature-changelog.md
+++ b/docs/feature-changelog.md
@@ -6,8 +6,9 @@
 Manual Quick and Complete Refresh now finish the selected item's artwork before reporting success instead of leaving it behind the global image-cache backlog.
 - Chooses a text-bearing poster in the library's metadata language, then English, another language, and finally textless artwork.
 - Honors an optional `includes_text` signal from metadata plugins so a language-tagged but textless poster cannot outrank one that actually carries a title.
-- Claims only the refreshed item's cache jobs, retries failed or delayed jobs, and waits for an existing worker without draining unrelated artwork.
-- Replaces the web app's refresh spinner with the final success or failure message.
+- Claims only the refreshed item's cache jobs — including its seasons, episodes, and localizations — retries delayed ones once, and waits briefly on a worker that already holds a job without draining unrelated artwork.
+- Keeps the refresh itself successful when artwork cannot be cached in time: the item, its new artwork paths, and the page it lives on still update, and the leftover artwork is reported as a warning and finishes on the background queue.
+- Replaces the web app's refresh spinner with the final success, warning, or failure message.
 
 ## 2026-08-16
 
--- a/docs/feature-changelog.md
+++ b/docs/feature-changelog.md
@@ -1,5 +1,14 @@
 # Feature Changelog
 
+## 2026-08-19
+
+### Make metadata refresh finish with the right artwork
+Manual Quick and Complete Refresh now finish the selected item's artwork before reporting success instead of leaving it behind the global image-cache backlog.
+- Chooses a text-bearing poster in the library's metadata language, then English, another language, and finally textless artwork.
+- Honors an optional `includes_text` signal from metadata plugins so a language-tagged but textless poster cannot outrank one that actually carries a title.
+- Claims only the refreshed item's cache jobs, retries failed or delayed jobs, and waits for an existing worker without draining unrelated artwork.
+- Replaces the web app's refresh spinner with the final success or failure message.
+
 ## 2026-08-16
 
 ### Let viewers turn the intro prompt off
--- a/internal/adminjob/item_refresh.go
+++ b/internal/adminjob/item_refresh.go
@@ -416,6 +416,10 @@ type ItemRefreshIngester interface {
 	IngestSubtree(ctx context.Context, folder *models.MediaFolder, subtreePath string) (*libraryingest.Result, error)
 }
 
+type ItemRefreshArtworkCacher interface {
+	CacheTargetArtwork(ctx context.Context, targetContentID string) error
+}
+
 type ItemRefreshExecutor struct {
 	folderRepo      itemRefreshFolderRepo
 	fileRepo        itemRefreshFileRepo
@@ -430,8 +434,9 @@ type ItemRefreshExecutor struct {
 		RefreshItemForLibrary(ctx context.Context, contentID string, folderID int) error
 		RefreshTargetForLibrary(ctx context.Context, targetType, contentID string, folderID int) error
 	}
-	eventBus    cache.EventBus
-	realtimeHub *notifications.Hub
+	artworkCacher ItemRefreshArtworkCacher
+	eventBus      cache.EventBus
+	realtimeHub   *notifications.Hub
 }
 
 func NewItemRefreshExecutor(
@@ -466,6 +471,10 @@ func NewItemRefreshExecutor(
 	}
 }
 
+func (e *ItemRefreshExecutor) SetArtworkCacher(cacher ItemRefreshArtworkCacher) {
+	e.artworkCacher = cacher
+}
+
 func (e *ItemRefreshExecutor) Execute(ctx context.Context, req ItemRefreshRequest, progress func(current, total int, message string)) (*ItemRefreshResult, error) {
 	if e.folderRepo == nil || e.ingester == nil || e.refresher == nil {
 		return nil, fmt.Errorf("resolve scan scope: item refresh executor is not fully configured")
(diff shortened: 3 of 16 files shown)


blurbery:
@coderabbitai review

Quick104:
## Pushed review fixes to this branch (`b1ca4fb3`)

I reviewed this branch, found seven defects, and pushed fixes rather than leaving them as comments. Happy to drop the commit if you'd rather take them yourself.

I did **not** touch the migration (verified correct — single file, `NO TRANSACTION` on both the concurrent create and the drop, `StatementBegin`-wrapped `DO` block, and the column order serves all four new predicates) or the poster tier ordering, which is a product call — see the bottom.

**What changed**

1. **An artwork failure abandoned a committed refresh** (`item_refresh.go:557`). `CacheTargetArtwork`'s error returned directly, jumping over `publish(EventScanComplete/EventMetadataUpdated)`, both catalog publishes, and the result payload. Since web `onSuccess` owns every `invalidateQueries` *and* the `onReplaced(detailContentID)` navigation, a Complete Refresh that remapped `old-episode-id` → `new-episode-id` and then hit one 404 poster left the browser on an id that no longer resolved, with every cache entry stale. The failure is now a warning on the result, surfaced as one in the UI; the refresh completes.

2. **The target-scoped wait never reached a series' children** (`image_cache_processor.go:254`) — the case the changelog claims to fix. Season and episode jobs are enqueued under *their own* content ids (`service.go:4400,4419`) with `series_id` pointing at the series, so filtering `target_content_id` alone meant a series refresh claimed the two series-level rows, found nothing, and reported success while 60 episode stills sat behind the global backlog. Scope now matches either column; the existing `(series_id, status)` index serves it. Person jobs have an empty `series_id` and stay out of scope.

3. **The wait was unbounded and hot-polled** (`:288-301`). No deadline beyond the 15-minute lease, and four statements per 100ms pass. A worker killed mid-download parked an interactive refresh on a spinner for the full lease. Now backs off 100ms→2s, gives up after 30s without progress, and returns `ErrTargetArtworkPending` so the queue finishes the rest.

4. **`retryTargetNow` cleared backoff on unrelated jobs** (`image_cache_job_repo.go:343`). It reset `attempt_count = 0`, `next_attempt_at = NOW()`, and `last_error = ''` for every row sharing the target, on *every* pass. It now runs once up front and preserves `attempt_count`/`last_error` — a manual refresh buys one immediate retry, it doesn't refund a budget a background worker already spent.

5. **Backdrop selection lost its terminal tier** (`service.go:6819`). Narrowed to `languageNeutral` with no catch-all, while poster and logo kept theirs. An item whose backdrops are all language-tagged — common for animation and title-card art — kept an empty backdrop permanently, and no `media_item_localizations` row could ever pick a language-specific one. Language-neutral still wins; a tagged backdrop now beats none.

6. **Step count lied when caching was off** (`item_refresh.go:491`). The cacher is wired whenever `S3Public != nil`, independently of `metadata.cache_images`, so every refresh advertised 4 steps and published "Caching refreshed artwork" at 4/4 while `CacheTargetArtwork` returned immediately. Now gated on whether caching would actually run.

7. **Failure progress went backwards** (`runner.go:756`). `containsPhase` is a plain `strings.Contains` and matched none of the phases for the new message, so a job that had just published 4/4 finalized as 0/3. Totals now track what the executor published.

**Worth your attention**

- **Fix 2 makes a series refresh do materially more synchronous work.** Waiting for every season poster and episode still is the point, but a 200-episode series now spends minutes on "Caching refreshed artwork" (16 per claim, concurrency 3) where it used to return immediately and lie. The 30s idle timeout bounds *stalls*, not total work, since it resets on progress. That's the intended trade but it deserves a conscious sign-off, since it's the most visible behavior change here.
- **The fallback backdrop tier isn't language-preference-ordered.** Within `hasLanguage` it's first-provider-then-max-rating, so an English library with only `de` and `fr` backdrops picks arbitrarily. `logoFilters` orders by preferred-then-English. Not a regression — before this it got nothing — but an asymmetry worth closing.
- **The new SQL has no test that executes it.** I checked all four statements by hand and against every enqueue site, but the `(target_content_id = $n OR series_id = $n)` predicate has never run against a database here.
- **Unchanged, needs your call:** the new poster tier 3 (`posterHasText`) sits *above* `posterTextless`, so an English library with no English poster now deterministically prefers a Japanese-titled poster over a clean textless one, and `local_image_source_test.go` enshrines it. The old behavior was inconsistent rather than correct, so this reads as deliberate — but it's a taste decision that will show up in real libraries, and it shouldn't be settled by a test name.

`go build ./internal/...`, `go vet`, `gofmt`, `go test -race` on `adminjob` + `metadata`, `tsc --noEmit`, prettier, eslint, and the affected vitest files all pass. Two `jellycompat` failures are pre-existing on `main` (macOS PID reuse) and untouched by this diff.

Note this branch and #668 both rewrite `image_cache_job_repo.go` in opposite directions — #668 decides what stays parked, this adds a per-target revival path. They merge textually clean, so the order they land in matters.

<sub>Review and fixes produced with AI assistance (Claude), verified by building and running the branch. See [docs/ai-contributions.md](docs/ai-contributions.md).</sub>

Issue #348, opened 2026-07-08
Cached artwork breaks permanently when the public S3 provider changes


Quick104:
## Problem

When an admin changes the public S3 provider (endpoint, bucket, or key prefix), every cached image breaks permanently unless they migrate the bucket contents themselves:

- The DB stores bucket-relative keys (`tmdb/movies/550/poster/original.webp`) in `poster_path` / `backdrop_path` / etc., and serving resolves them against the *current* S3 config — so after a provider change every URL points into a bucket where nothing exists.
- The cached path in the row is the image cache pipeline's durable dedup marker: `EnqueueExistingProviderArtwork` only enqueues rows whose destination is still a provider URL or empty, so already-cached rows are never re-enqueued.
- Clients fetch images directly from S3 via presigned/read-endpoint URLs, so the server never even observes the 404s. The breakage is silent and recovery requires hand-written SQL.

The same silent-breakage applies to chapter thumbnails, collection artwork, library posters, branding assets, and audiobook/ebook covers — everything in the public bucket.

## Proposal

Automatic verify-and-reconcile, no admin decisions required. Design spec: `docs/superpowers/specs/2026-07-08-s3-artwork-reconcile-design.md`.

1. **Storage identity fingerprint** (`endpoint|bucket|key_prefix` of the public write config) persisted in `server_settings`; compared at startup.
2. **`reconcile_artwork_cache` task** (startup trigger, gated by the fingerprint; manually runnable): probes a sample of cached objects, then either bulk-resets (nothing migrated) or per-row verifies (migrated / partial). Missing provider-sourced artwork is reset to its `*_source_path` so the existing image cache queue rebuilds it; non-redownloadable surfaces are cleared so their owning pipeline (chapter backfill, book enrichment, collection collage) refills them, with unrecoverable uploads reported.
3. **Settings UI callout** warning that uploaded artwork can't be re-downloaded.

Users never see broken images during the transition: reset rows serve the provider's original URL via the existing absolute-URL pass-through, and thumbhashes are kept.

PR #684, opened 2026-08-20
fix(metadata): make artwork recovery explicit

blurbery:
## Problem

Part of #348. Follow-up to #349.

The storage-change reconciler introduced in #349 can run automatically at startup. When a newly configured public bucket is empty, incomplete, or misrouted, successful `HeadObject` 404 responses can satisfy the 95% bulk-reset threshold. Silo then rewrites artwork pointers before an administrator has had a chance to migrate the existing objects.

The existing scheduled **Cache Metadata Images** task also mixes two very different operations: draining durable jobs created by scans and metadata changes, and periodically discovering provider artwork across the entire catalog. Because its discovery timestamp is process-local, the first run after every restart can begin another full-library sweep even when the durable queue has already been completed.

Interrupted reconcile runs also persisted progress only in memory, so a manual retry restarted the full catalog walk. Person-photo and library-poster pagination cast numeric IDs to text, forcing repeated scans and sorts on large catalogs.

## Approach

- Fail closed for every scheduled reconciliation. The startup trigger may detect and log a changed storage identity, but it never launches the mutating sweep.
- Require an explicit administrator-run **Reconcile Artwork Cache** task after the existing artwork objects have been copied or migrated.
- Make scheduled/startup **Cache Metadata Images** drain only durable jobs already queued by scans, refreshes, and metadata changes.
- Move full-catalog discovery into a separate manual-only **Backfill Metadata Images** task. It has no default triggers and its scheduler preflight always fails closed; only an explicit administrator run can start it.
- Enforce manual-only behavior in the task API and UI, so a backfill schedule cannot be configured accidentally.
- Claim only the two artwork jobs that can start immediately while retaining 1,000-row discovery pages, avoiding lease expiry on a large unstarted batch.
- Give every execution a unique lease owner and serialize drain/backfill runs through a cancellable gate, preventing stale or overlapping workers from finalizing each other's jobs.
- Make a manual backfill run until discovery is complete or explicitly cancelled, and fail closed if image caching is disabled before or during the run.
- Warn in Storage Settings that reconcile resets missing cache records and that backfill is a separate manual action.
- Persist a versioned verify-mode checkpoint after each fully completed batch, scoped to the exact baseline-to-target storage transition.
- Preserve interrupted checkpoints for an explicit manual retry; same-identity manual recovery still starts a fresh sweep.
- Keep pagination and guarded updates on native primary-key types, casting only selected cursor values for serialization.
- Stop at the first unsafe batch instead of advancing a checkpoint past storage errors.

Normal incremental caching is preserved: newly added or changed artwork still creates durable jobs, and the scheduled cache task drains those jobs. Manual reconcile and backfill retain the existing recovery behavior, but neither can be initiated by normal scheduling.

This PR does not silently certify a new bucket or copy/delete S3 objects. The person-photo enrichment feedback loop is addressed separately in #685.

## Testing

```text
GOWORK=off go test ./internal/metadata ./internal/taskmanager/tasks ./internal/api/handlers ./internal/config ./cmd/silo -count=1
GOWORK=off go test -race ./internal/metadata ./internal/taskmanager/tasks -count=1
GOWORK=off go vet ./internal/metadata ./internal/taskmanager/... ./internal/api/handlers ./internal/config ./cmd/silo
pnpm exec vitest run src/pages/admin-settings/StorageSettings.test.tsx
pnpm exec tsc -b
pnpm exec eslint src/pages/admin-settings/StorageSettings.tsx src/pages/admin-settings/StorageSettings.test.tsx
pnpm exec prettier --check src/pages/admin-settings/StorageSettings.tsx src/pages/admin-settings/StorageSettings.test.tsx
pnpm run build
git diff --check
```

The PostgreSQL-backed person-enrichment integration case remains separate follow-up work and skips when `SILO_TEST_DATABASE_URL` is not configured.

### AI Disclosure

- Tool(s): OpenAI Codex
- Model(s): GPT-5 (Codex)
- Involvement: Developed collaboratively by @blurbery and Codex. @blurbery reproduced the behavior on a real deployment, supplied runtime and storage observations, and directed the desired safety behavior and upstream contribution path. Codex assisted with code analysis, implementation, tests, and verification.
- Adversarial review: Independent read-only review found both the automatic-startup reconciliation risk and the restart-triggered catalog-discovery path. The revised design blocks scheduled reconciliation mutation and separates safe durable-queue draining from explicit full-catalog backfill.

## Checklist

- [x] Scheduled storage-change reconciliation fails closed.
- [x] Scheduled metadata caching drains only durable queued work.
- [x] Full-catalog discovery is a separate manual-only task.
- [x] Manual-only scheduling is enforced by both API and UI.
- [x] Queue leases, overlap, cancellation, and mid-run disable behavior are covered.
- [x] Manual recovery remains available and checkpointed.
- [x] Admin UI explains the migration-first and explicit-backfill requirements.
- [x] Relevant Go and Web tests pass locally.
- [x] AI assistance is disclosed.

## Summary by CodeRabbit

* **New Features**
  * Added resumable artwork cache reconciliation that preserves progress across interruptions.
  * Added a manual metadata-image backfill task for existing catalog content.
  * Task details now identify tasks that can only be run manually.

* **Changes**
  * Queued image processing is separate from full-catalog backfills for more predictable operation.
  * Public storage changes now clearly require manual cache reconciliation and image backfill.

* **Bug Fixes**
  * Prevented scheduled execution of manual-only tasks.
  * Protected system-managed reconciliation settings from administrative edits.

### DIFF
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -2140,10 +2140,10 @@ func main() {
 		}
 		if deps.S3Public != nil {
 			identity := tasks.ArtworkStorageIdentity(cfg.S3.Public.Endpoint, cfg.S3.Public.Bucket, cfg.S3.Public.KeyPrefix)
-			// Seed the fingerprint on first boot so an unchanged storage
-			// identity never triggers a sweep. On the boot after a provider
-			// change the stored (old) identity survives this call and the
-			// startup trigger runs the reconcile.
+			// Seed the fingerprint on first boot. After a provider change the
+			// stored (old) identity survives this call, so the startup preflight
+			// can warn without mutating artwork; an administrator must migrate
+			// objects and run the reconcile task explicitly.
 			if _, err := settingsRepo.SetIfAbsent(appCtx, tasks.ArtworkStorageIdentityKey, identity); err != nil {
 				slog.Warn("artwork reconcile: seeding storage identity failed", "error", err)
 			}
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -2137,6 +2137,7 @@ func main() {
 		}
 		if metadataImageCacheProcessor != nil {
 			taskMgr.Register(tasks.NewCacheMetadataImagesTask(metadataImageCacheProcessor))
+			taskMgr.Register(tasks.NewBackfillMetadataImagesTask(metadataImageCacheProcessor))
 		}
 		if deps.S3Public != nil {
 			identity := tasks.ArtworkStorageIdentity(cfg.S3.Public.Endpoint, cfg.S3.Public.Bucket, cfg.S3.Public.KeyPrefix)
--- a/internal/api/handlers/admin.go
+++ b/internal/api/handlers/admin.go
@@ -1259,6 +1259,21 @@ func (h *AdminHandler) HandleUpdateItemMetadata(w http.ResponseWriter, r *http.R
 // encryption can never drift apart. See catalog.SensitiveSettingKeys.
 var sensitiveSettingKeys = catalog.SensitiveSettingKeys
 
+// machineManagedSettingKeys contains durable internal state that shares the
+// server_settings store but is not part of the administrator settings API.
+var machineManagedSettingKeys = map[string]bool{
+	config.ArtworkStorageReconcileCheckpointKey: true,
+}
+
+func redactAdminSettings(values map[string]string) {
+	for key := range sensitiveSettingKeys {
+		delete(values, key)
+	}
+	for key := range machineManagedSettingKeys {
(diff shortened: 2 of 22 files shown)


blurbery:
@coderabbitai review\n\nPlease review the latest head (939848e), including the queue-only scheduler split, manual-only backfill enforcement, lease ownership/claim sizing, cancellable run serialization, and cache-disable behavior.

blurbery:
@coderabbitai review\n\nPlease review the latest head (3da9c5f). This supersedes the prior request after the changed-line linter required a comments-only spelling correction.

PR #685, opened 2026-08-20
fix(catalog): preserve cached person artwork during enrichment

blurbery:
## Problem

Per-item credit enrichment treats provider person photos as authoritative and overwrites an existing cached `people.photo_path`. The image-cache worker then restores the immutable S3 key, but a later item scan overwrites it again. Discovery is allowed to re-queue succeeded jobs, so the same actor image can be downloaded repeatedly.

Replacing the cached path also queues the displaced immutable revision for artwork garbage collection. If the pointer is not restored before the grace period expires, valid cached profile variants can be deleted.

## Fix

- Make single and batch person enrichment follow their fill-empty contract for photo path, source path, and thumbhash.
- Continue allowing a real photo to replace the explicit `-` no-photo sentinel.
- Leave replacement of an existing person image to the full person-refresh flow, which already preserves cached artwork while recording/enqueuing a changed source.
- Add a real change predicate so no-op credit enrichment does not advance `updated_at` and postpone scheduled full-person refreshes.
- Exercise the production batch SQL in the normal unit suite and add an optional PostgreSQL integration regression test.

## Testing

```text
GOWORK=off go test ./internal/catalog -run 'TestPerson(PhotoEnrichmentSQLGuardsExistingArtwork|CreditEnrichmentPreservesCachedArtwork)' -count=1
GOWORK=off go test ./internal/catalog -count=1
```

The PostgreSQL-backed case skips when `SILO_TEST_DATABASE_URL` is not configured; the SQL-shape regression runs in every normal unit suite.

### AI Disclosure

- Tool(s): OpenAI Codex
- Model(s): GPT-5 (Codex)
- Involvement: Developed collaboratively by @blurbery and Codex. @blurbery reproduced the repeated actor caching on a real deployment and established the expected behavior. Codex assisted with source tracing, implementation, tests, and verification.
- Adversarial review: Independent review caught the need to preserve real-photo replacement of the `-` sentinel and to avoid no-op `updated_at` changes; both are covered in the final patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved person data enrichment to fill missing details without overwriting existing artwork or cached photo paths.
  * Correctly replaces no-photo placeholders and uncached photo links when valid artwork becomes available.
  * Prevented unnecessary timestamp changes and cleanup activity when enrichment makes no changes.

* **Tests**
  * Added coverage for artwork preservation, placeholder replacement, coordinated photo updates, and no-op enrichment behavior.

### DIFF
--- a/internal/catalog/person_repo.go
+++ b/internal/catalog/person_repo.go
@@ -86,8 +86,52 @@ func (r *PersonRepository) FindOrCreate(ctx context.Context, p models.Person) (i
 }
 
 // enrichExisting updates empty fields on an existing person with non-empty values from p.
+func personPhotoFillPredicate(existing, incoming string) string {
+	return fmt.Sprintf(
+		"((COALESCE(%s, '') = '' AND %s <> '') OR (%s = '-' AND %s NOT IN ('', '-')))",
+		existing, incoming, existing, incoming,
+	)
+}
+
+func batchPersonEnrichmentQuery() string {
+	photoPathFill := personPhotoFillPredicate("people.photo_path", "t.photo_path")
+	photoSourceFill := personPhotoFillPredicate("people.photo_source_path", "t.photo_source_path")
+	photoThumbFill := personPhotoFillPredicate("people.photo_thumbhash", "t.photo_thumbhash")
+	return fmt.Sprintf(`
+		UPDATE people SET
+			tmdb_id = CASE WHEN COALESCE(people.tmdb_id, '') = '' AND t.tmdb_id <> '' THEN t.tmdb_id ELSE people.tmdb_id END,
+			imdb_id = CASE WHEN COALESCE(people.imdb_id, '') = '' AND t.imdb_id <> '' THEN t.imdb_id ELSE people.imdb_id END,
+			tvdb_id = CASE WHEN COALESCE(people.tvdb_id, '') = '' AND t.tvdb_id <> '' THEN t.tvdb_id ELSE people.tvdb_id END,
+			plex_guid = CASE WHEN COALESCE(people.plex_guid, '') = '' AND t.plex_guid <> '' THEN t.plex_guid ELSE people.plex_guid END,
+			photo_path = CASE WHEN %[1]s THEN t.photo_path ELSE people.photo_path END,
+			photo_source_path = CASE WHEN %[2]s THEN t.photo_source_path ELSE people.photo_source_path END,
+			photo_thumbhash = CASE WHEN %[3]s THEN t.photo_thumbhash ELSE people.photo_thumbhash END,
+			bio = CASE WHEN COALESCE(people.bio, '') = '' AND t.bio <> '' THEN t.bio ELSE people.bio END,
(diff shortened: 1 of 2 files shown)


Quick104:
Pushed a follow-up commit (e0fa219) addressing the review findings.

**Photo columns now move as a unit.** `photo_path`, `photo_source_path`, and `photo_thumbhash` were each gated on their own value. A credit carries a photo URL but never a source path, so the sentinel-replacement arm rewrote `photo_path` and left the previous source URL behind. `photo_source_path` is what `UpdatePhotoIfSourceMatches` keys the cache handshake on and what `EnqueueExistingProviderArtwork` downloads from, so the finished job landed the *old* image on the row, under the old image's thumbhash. All three now move together under one decision taken on `photo_path`.

**The guard protects cached keys, not any populated value.** Deferring every replacement to the full person refresh stranded people with no `tmdb_id`/`imdb_id`/`tvdb_id` — `FindRefreshCandidates` skips them, so nothing would ever revisit a dead URL. An empty column, the `-` sentinel, and an uncached provider URL are replaceable again; a cached artwork key is not. "Not a cached key" is the same `LIKE '%://%'` test the artwork GC trigger (`queue_displaced_artwork_revision`) and the image cache sweep use, so displacing a URL still queues nothing for deletion. Replacement also requires a genuinely different path, so re-scanning an unchanged credit stays a no-op — otherwise it would clear the source path a pending cache job is keyed on.

**Tests.** The DB test's helpers took the subtest's `*testing.T` instead of closing over the parent, so a failure now reports its message instead of `test executed panic(nil) or runtime.Goexit`. The SQL-shape test matches whole generated clauses rather than loose fragments, so a mis-wired column fails it. New cases cover the stale-source binding, uncached-URL replacement, and the unchanged-credit no-op.

To be clear about coverage: the behavioral test still needs `SILO_TEST_DATABASE_URL`, and CI has no Postgres service, so it skips there. The description's claim that the shape regression covers the behavior isn't accurate — it pins the SQL, not the outcome. Verified locally against `pgvector/pgvector:pg18` with all SQL migrations applied:

- `SILO_TEST_DATABASE_URL=... go test ./internal/catalog/ -run 'TestPersonPhotoEnrichmentSQLGuardsCachedArtwork|TestPersonCreditEnrichmentPhotoRules' -v` — all 6 subtests pass.
- Mutation check: reverting the predicate to the old overwrite-if-real rule fails `cached key survives single/batch find or create` and `unchanged url is a no-op`.
- `go test ./internal/catalog/` (no DB) and `go build ./...` pass. `TestEpisodeSearchPostgresAndDocumentSource` fails with a DB configured, but it fails identically on this PR's head without these changes — unrelated.
- `golangci-lint run --new-from-merge-base=origin/main ./internal/catalog/...` — 0 issues.

Also built the batch enrichment SQL once via `sync.OnceValue` instead of rebuilding ~1.5 KB per batch.

One thing left alone deliberately: credits still never populate `photo_source_path`, so a person whose photo arrives only from item credits is not picked up by the image cache sweep until a full refresh derives the source. That matches the insert path, which has always stored an empty source for new people, and changing it would newly enqueue downloads for the whole credit population — worth its own PR if wanted.

Issue #629, opened 2026-08-13
feat(access): add video and audio transcoding controls to access groups


tomislav:
## Summary

Access groups can currently define `max_transcodes`, but they cannot control whether members may perform video or audio-only transcoding.

The equivalent user editor supports:

- Disabling video transcoding with `transcode_allowed`
- Independently permitting audio-only transcoding with `audio_transcode_allowed`

Adding these controls to access groups would allow administrators to configure the policy once for every member instead of editing users individually.

## Current behavior

Confirmed by auditing current `main` at commit `ee9356a`:

- `users` stores `transcode_allowed` and `audio_transcode_allowed`.
- `access_groups` stores `max_transcodes`, but neither permission flag.
- The access-group CRUD API does not expose either flag.
- `ApplyGroupPolicy` derives both effective permission flags solely from the user.
- The access-group editor renders a numeric “Max transcodes” field without the video-disable/audio-transcodes controls available in the user editor.

This is a code-audit finding, not a runtime bug report. No deployment reproduction or logs are claimed.

## Proposed behavior

Add `transcode_allowed` and `audio_transcode_allowed` to access groups.

Because group composition is restriction-only, effective permissions should be:

```text
effective transcode_allowed =
  user.transcode_allowed && group.transcode_allowed

effective audio_transcode_allowed =
  user.audio_transcode_allowed && group.audio_transcode_allowed
```

Existing access groups should default both fields to `true` so the migration preserves current behavior.

The access-group editor should offer the same interaction currently available under a user’s “Max Transcodes” setting:

- Enable or disable video transcoding.
- When video transcoding is disabled, independently allow or disallow audio transcodes.
- Disable the numeric limit input when video transcoding is disabled.

## Scope

- Additive Goose migration for both access-group columns
- Access-group store models, queries, and CRUD inputs
- Additive `/api/v1/admin/access-groups` request and response fields
- Restriction-only effective-policy composition
- Access-group web API types and editor controls
- Store, handler, policy-composition, and frontend tests

## Compatibility

This can remain additive within `/api/v1`. Defaulting both new fields to `true` should preserve the behavior of existing groups and their members.

No Android or Apple client changes appear necessary because this is an admin web surface and the effective playback policy is enforced server-side.

### AI Disclosure

- Tool(s): Codex
- Model(s): gpt-5.6
- Involvement: AI-assisted
[...]

PR #691, opened 2026-08-20
Scoped admin API keys, and inherit/override user policy

Quick104:
Part of #629 (access-group transcode controls). Client follow-up: Silo-Server/silo-apple#178, Silo-Server/silo-android#240.

Two related changes to the admin API's authorization and to how per-user policy is resolved.

## Review fixes (commit f8574f3e)

Changes made in response to the review pass and the CodeRabbit/Codex comments:

- **Scoped-key escalation closed.** An `admin:users` key could create an admin account (or reset an admin's password) and log in for an unscoped JWT. Scoped actors may no longer assign the `admin` role or change credentials on an existing admin; ordinary-account provisioning still works. `POST /api/v1/api-keys` now accepts `scopes` (it silently dropped them), and `GET /api/v1/api-keys/scopes` is the capability endpoint. New `docs/api-keys-api.md`.
- **`download_transcode_allowed` migration.** It was the one column whose old default was `false`, so every legacy row became an explicit deny and the group toggle was inert. `false` now maps to inherit, `NoGroupPolicy` matches the old default (`false`), and Down restores `false` to match the column default. Legacy negative caps (previously "defer to group") map to inherit instead of becoming an unlimited override. Validated up/down/up on a throwaway Postgres.
- **Fail closed.** Home sections and the item access filter now return 500 on a policy-lookup error like `/libraries` did (the path is unreachable with the viewer-access middleware, but it no longer fails open). The legacy metadata-curation gate resolves inherited group libraries. `EffectivePolicyForUser` skips the DB for ungrouped users, which removes the extra query on login and `/auth/me`.
- **Web admin.** Inherit hints follow the *selected* group on both tabs and on the create form (they were stale on the Limits tab and absent on create); Override seeds the real inherited value and never `0`; a cleared limit box cannot be saved as `0`/`NaN`; the effective panel shows the group-intersected permissions and always shows the audio-transcode row; the Access Groups copy describes inherit/override.
- **Cleanups.** One `optionalField[T]` decoder (the two bespoke ones and `AccessGroupIDSet` are gone), generic `clonePtr`, shared cap validator, table-driven `UserRepository.Update`, one policy field table in the web form, orphaned `UserTranscodeLimitField` and unused `OverrideSources` removed. The requests service dropped its dead group-only fallback.

Declined: keeping the admin user fields non-nullable with overrides in a separate object (CodeRabbit) — the API is pre-lock and this PR already coordinates the break with client issues above; the "nil user from `GetByID`" concern does not apply (`scanUser` maps `ErrNoRows` to `ErrNotFound`). Not done: de-duplicating the two policy resolutions on media-request creation would require reshaping the entitlement resolver — left as-is.

## Scoped API keys, and admin user API hygiene

`sa_` API keys can now carry scopes. A key with no scopes behaves exactly as before — full access as its owning user — so nothing existing changes. A key with scopes is an allowlist credential: the auth middleware admits it only to the routes its scopes name. Scopes narrow, never grant; the owning user's role checks still apply afterwards.

Two scopes exist: `admin:users` (user lifecycle plus reading a user's profiles) and `admin:access-groups:read`. Impersonation is explicitly outside both — an API key cannot mint a session for another user under any scope.

The admin user endpoints also got a pass for consistency: explicit column lists instead of `SELECT *`, uniform validation of policy input, and the same tri-state decoding on every field rather than three different conventions.

## Per-user policy: inherit or override

User policy stops being "the strictest of the user's value and the group's value wins" and becomes inherit/override.

Every user policy field is now nullable — `max_streams`, `max_transcodes`, `max_playback_quality`, `transcode_allowed`, `audio_transcode_allowed`, `download_allowed`, `download_transcode_allowed`, `library_ids`, plus a new `requests_allowed`. `NULL` means inherit the access group's value. A value set on the account is authoritative **in either direction**: an admin can grant downloads to one member of a no-downloads group, or cap one member of an unlimited group. Resolution is `user.field ?? group.field ?? no-group default`.

Consequences worth calling out:

- `0` on a stream or transcode cap now means an explicit "unlimited" override. It used to mean "defer to the group", which is what `NULL` means now.
- Access groups gain `transcode_allowed` and `audio_transcode_allowed`, so every account-level field has a group value to inherit from.
- `GET` on an admin user returns the stored overrides — null where inherited — alongside an `effective_policy` block carrying the resolved values, so a caller never has to reimplement the merge. Effective policy fails closed if the group lookup fails rather than reporting an unrestricted account.
- `PUT` accepts an explicit `null` on any policy field to clear an override back to inherit, and omitting a field leaves it untouched. An empty `library_ids` array round-trips as a real "no libraries" override rather than collapsing to inherit.
- Login and `/auth/me` report the resolved `download_allowed`.
- The web admin gains per-field Inherit/Override controls that show the effective value beside each inherited field, and the access-group editor gains the two transcode gates.

## Migration

`20260819213116_user_policy_inherit_override` makes the columns nullable and maps existing rows so behavior is preserved wherever an account was already deferring to its group: `0`, `''`, and `true` become inherit, while explicit restrictions — `false`, positive caps, a named quality, a library list — are kept as overrides. Pre-existing `NULL`s stay inherit rather than being read as a restriction.

One behavior change is deliberate and unavoidable: an account whose stored cap is **above** its group's cap now wins instead of being clamped. That is the point of the model. On the deployment this was rehearsed against, that affected two accounts out of 132.

The down migration is lossy in the usual way — inherit and "zero means defer" are not distinguishable once the column is non-nullable again — and says so in the file.

## Verification

- `go build ./...`, `go vet`, and unit plus database-backed tests across `access`, `auth`, `api`, `api/handlers`, `api/middleware`, `downloads`, and `models`.
- Web: `tsc`, vitest, eslint, prettier.
- Applied to a copy of a real deployment (132 users) and exercised end to end: create-inherits, an override beating a more restrictive group, an explicit null clearing an override, an absent field leaving the value untouched, a group edit propagating to inheriting accounts, `/auth/me` and download capability resolving correctly, and scoped-key allow/deny enforced per route.
- `internal/jellycompat`'s `TestBeginWebOperation*` failures reproduce on `main` at the same commit and are unrelated to this branch.

## Summary by CodeRabbit

* **New Features**
  * Added access-group inheritance for user permissions, libraries, playback quality, downloads, requests, and transcoding.
  * Added separate video and audio transcoding controls for access groups and individual users.
  * Admin screens now show effective policies and distinguish inherited settings from personal overrides.
  * Added scoped API keys with route restrictions and scope discovery.
  * Invitation, authentication, library, catalog, download, and request access now respect effective group policies.
* **Documentation**
  * Added API-key documentation and updated architecture guidance and feature changelog for policy inheritance and access controls.


Quick104:
Addressed the automated review comments in f8574f3e:

| Comment | Outcome |
|---|---|
| Codex: expose scopes via a capability route | Done — `GET /api/v1/api-keys/scopes` |
| Codex: legacy metadata gate ignores inherited libraries | Done — gate now resolves via `EffectivePolicyForUser`, fails closed |
| Codex / CodeRabbit: Limits tab uses stale `effective_policy` | Done — hints derive from the selected group on both tabs and the create form |
| Codex: negative legacy caps become unlimited | Done — migration maps all caps `<= 0` to inherit |
| Codex: Access Groups page copy | Done — rewritten for inherit/override |
| CodeRabbit: `download_transcode_allowed = false` becomes an explicit deny | Done — maps to inherit; `NoGroupPolicy` and Down fixed to match |
| CodeRabbit: fail closed in items/sections | Done |
| CodeRabbit: NaN / empty numeric input | Done — raw string state, empty box blocks save |
| CodeRabbit: effective panel uses `user.permissions` | Done — uses `effective.permissions` |
| CodeRabbit: audio-transcode row hidden | Done — always rendered |
| CodeRabbit: DB test should skip on missing columns | Done — gate extended to the new columns |
| CodeRabbit: "Access composition unchanged" heading | Done — "Invitation binding unchanged" |
| CodeRabbit: keep admin user fields non-nullable | Declined — pre-lock API; break coordinated via silo-apple#178 / silo-android#240 |
| CodeRabbit: nil user from `GetByID` | Not applicable — `scanUser` returns `ErrNotFound`, never `(nil, nil)` |

Also fixed from a separate review pass: a scoped `admin:users` key could mint an admin and log in unscoped (now refused), and `POST /api/v1/api-keys` silently dropped `scopes`.

PR #712, opened 2026-08-21
feat: add playback.proxy_policy setting to control proxy streaming routing

jasjeetsuri:
## Summary

Add a new `playback.proxy_policy` admin setting that controls when available proxy nodes serve media streams to clients.

## Motivation

Currently, when proxy nodes are available and healthy, **all** streams — direct play, remux, and transcode — are unconditionally routed through them. This is ideal when the proxy has more egress bandwidth than the API server, but it creates unnecessary overhead for operators whose API server has sufficient bandwidth for direct play and only needs to offload computationally expensive video transcoding to remote hardware (e.g. Intel QSV on a homelab server).

A common distributed deployment pattern is:
- **VPS** (ARM, limited CPU) running Silo in integrated mode — handles API, direct play, audio remux
- **Homelab** (Intel, dedicated GPU) running proxy + transcode workers — handles only video re-encoding

Without this setting, the VPS sends all streaming traffic through the homelab proxy node, even for direct play where the VPS could serve the file bytes directly with zero transformation overhead. This doubles the network hops and adds latency for no benefit.

### New setting: `playback.proxy_policy`

An enum with three values:

| Value | Direct Play | Audio Remux (HLS) | Video Transcode |
|---|---|---|---|
| `always` (default) | Proxy node | Proxy + Transcode node | Proxy + Transcode node |
| `transcode_only` | API server | API server (local ffmpeg) | Proxy + Transcode node |
| `never` | API server | API server (local ffmpeg) | API server relays from Transcode node |

- **`always`** — preserves existing behavior; no change for current deployments.
- **`transcode_only`** — proxy nodes are only selected for sessions requiring video transcoding (`DeliveryTranscodeHLSV3`). Direct play, progressive remux, and HLS audio adaptation (`DeliveryRemuxHLSV3`) stay on the API server.
- **`never`** — proxy nodes are not used for any streaming. Transcode nodes still perform video encoding work, but the API server relays segments to clients. HLS audio remux runs locally.

### Files changed

- **`internal/nodepool/planner.go`** — `ProxyPolicy()` setting reader, following the same pattern as `LocalTranscodeFallbackAllowed()`
- **`internal/config/admin_settings.go`** — default value (`"always"`) and `normalizeAdminEnum` validation
- **`internal/api/handlers/playback_v3.go`** — gates proxy selection in `planIdentityProxyV3()` (direct play / progressive remux path) and clears proxy/transcode node selection in the HLS path based on policy
- **`web/.../PlaybackSettings.tsx`** — admin UI dropdown in the Transcoding section

### Design decisions

- The setting is read at request time (not cached), consistent with how `local_transcode_fallback` works — changes take effect immediately without restart.
- Under `transcode_only` and `never`, HLS audio remux (`DeliveryRemuxHLSV3`) runs locally on the API server rather than being dispatched to a remote transcode node. Audio remux is lightweight (stream-copy video + convert audio) and does not benefit from GPU offload.
- Default is `always` so existing deployments are unaffected.
- The setting is orthogonal to `local_transcode_fallback` — they control different aspects of the routing decision.

## Testing

Tested on a live deployment with:
- Oracle Cloud ARM VPS (integrated mode)
- Homelab K3s cluster with Intel QSV transcode + proxy workers via Pangolin reverse tunnel

Screenshot showing `never` mode with three concurrent streams:
- **Video transcode** (480p H.264) → dispatched to Homelab Transcode node ✅
- **Direct play** (4K HEVC TrueHD) → served by Local server (VPS) ✅
- **Audio remux** (DTS→AAC, video copy) → served by Local server (VPS) ✅

All three policy modes verified:
- `always`: all streams via proxy (existing behavior) ✅
- `transcode_only`: direct play and audio remux on VPS, video transcode via homelab ✅
- `never`: all streaming from VPS, only video encoding on homelab ✅

## Summary by CodeRabbit

- **New Features**
  - Added proxy usage controls: **Always**, **Transcode only**, or **Never**.
  - Added hardware and software HDR-to-SDR tone-mapping settings for transcoding.
  - Playback now supports additional media-authentication options and flexible proxy delivery.
  - Improved local fallback and progressive remux handling when proxy delivery is unavailable.

- **Bug Fixes**
  - Improved cleanup when playback transports are replaced or sessions are replanned.
  - Disabled or burned-in subtitle artifacts are now removed appropriately.
  - Improved tone-mapping capability detection and retry handling.
  - Added validation for playback features and improved copy-safety checks.


Quick104:
Thanks for putting this together and testing it on a real distributed setup. The underlying use case is one I want Silo to support: operators should be able to control where playback work runs and where client-facing egress comes from.

After reviewing the change, I don’t think a standalone `playback.proxy_policy` is the right abstraction to merge. Execution and egress need to be separate routing decisions, applied consistently across native playback, Jellyfin-compatible clients, retries, and replans. In the current shape those paths can disagree—for example, Jellycompat does not honor the setting, and a tone-map fallback can select a proxy under `never`.

I’m going to close this rather than ask you to keep iterating on an implementation shape we’re likely to replace. I do want to carry this deployment scenario and its routing requirements into the broader node-routing policy work.

Thank you again for taking the time to build and validate this.

PR #713, opened 2026-08-22
fix(web): smooth detail action interactions

blurbery:
## Problem

On movie and series detail pages, moving across the rating stars felt delayed. The component used a 150 ms `transition-all` while also switching the SVG fill between `none` and `currentColor`, so it asked the browser to transition more than the interaction needed and changed the fill as a discrete value.

The Play, Mark Watched, Favorite, and More buttons also kept the default arrow cursor. That made clickable controls in the same action bar behave differently from the rest of the web client.

There was one more slow path in the same action bar: moving onto or away from the larger Play and Mark Watched controls could make the continuously animated detail backdrop stutter. My first attempt composited those controls and removed their background transition, but production testing showed that this only removed interpolation. Hover still changed Play from its solid primary background to the alpha `/90` color and changed the watched control's glass background from alpha `0.4` to `0.55`, so each hover still took the paint path over the moving hero artwork.

## Approach

The rating now keeps a stable `currentColor` fill and animates its opacity. The button transition is limited to color and scale, shortened to 100 ms, and uses the existing hover scale only when reduced motion is not requested. This makes moving across the stars respond cleanly without changing rating or keyboard behavior.

Pointer cursors are added only to the enabled detail-page actions involved here, including each of the three Play render paths.

The final Play and watched fix keeps each control's computed background identical before and during hover. Feedback now comes from a small `1.02` scale on the already-composited control, with a matching active scale and reduced-motion opt-out. That keeps hover work on the compositor instead of repainting a control over the animated hero. The override covers all three enabled Play branches and the shared watched toggle; click handlers, focus rings, loading and disabled states, layout, progress overlay, and artwork animation are unchanged.

Everything stays local to the detail action bar instead of changing the shared Button component or unrelated screens.

Focused tests cover hover preview and reset, selecting and clearing a rating, reduced-motion classes, every Play render path, and the watched, Favorite, and More controls. The ActionBar coverage also locks in the stable hover backgrounds and compositor-only scale treatment.

## Validation

- `corepack pnpm exec vitest run src/pages/ItemDetail/components/ActionBar.test.tsx src/components/StarRating.test.tsx` — 2 files and 7 tests passed on `dcefe29b`.
- `corepack pnpm exec eslint` on the four changed ActionBar and StarRating files with `--max-warnings=0` — passed.
- `corepack pnpm exec prettier --check` on the four changed files — passed.
- `corepack pnpm run build` — passed.
- `make verify-local-paths` — passed.
- The generated production CSS was exercised in a local browser against the real Silo backend. Play kept `rgb(232, 232, 236)` before and during hover; Mark Series Watched kept its glass background at alpha `0.4`; both reported scale `1.02`, with no console errors.
- The same portable commit was first shipped to my fork as `87bf5af3`. Full fork Docs hygiene, Web, and Go CI passed, as did the multi-architecture Docker workflow. Build 72 (`ghcr.io/blurbery/silo-server@sha256:eca0e822ae225f61d707d64b52e3c1af3c9fa9e678e308dc27fad24dae691832`) is healthy in production with the exact revision label, and I confirmed the remaining hover lag is fixed.
- Upstream CI for `dcefe29b` is running and will provide the full repository gate for the final PR head. Docs hygiene and Web are already green; Go was still in tests at the last check.

## Adversarial review

CodeRabbit previously identified two coverage gaps: the other Play render paths and reduced-motion classes were not asserted. Both were added in `29ae1986`, and its fresh incremental review through `c64eef2` produced no actionable comments.

The fresh review of the final hover increment asked for assertions covering the active scale and duration, and for the temporarily disabled watched action to avoid the enabled pointer affordance. Both grounded findings were fixed in `dcefe29b`, including pending-state coverage. The follow-up push triggered the final incremental review status.

## Risks

This changes only web-client interaction styling and tests. It does not change APIs, stored data, playback behavior, layout, hero animation, card overlays, poster actions, watched indicators, or the shared Button component. Apple and Android clients are unaffected.

The main visual tradeoff is deliberate: Play and watched no longer change background opacity on hover. Their small scale response preserves clear pointer feedback without returning to the expensive paint path. Reduced-motion users get the stable control without the scale animation.

## AI Disclosure

- Model(s): GPT-5 (the session exposed no more specific model identifier)
- Involvement: AI-assisted
- Review: CodeRabbit reviewed the final hover increment; both grounded findings were fixed in `dcefe29b`. I also reviewed the complete diff and verified the generated CSS and deployed behavior.

I reproduced the issue and tested the final deployed behavior on my own Silo server. Codex helped trace the remaining paint path, implement the focused change and regression coverage, run the validation, deploy the verified artifact to my fork, and prepare this pull request update. I reviewed the final scope and am submitting the change.

## Checklist

### DIFF
--- a/web/src/components/StarRating.test.tsx
+++ b/web/src/components/StarRating.test.tsx
@@ -36,9 +36,14 @@ describe("StarRating", () => {
       "transform-gpu",
       "transition-[color,scale]",
       "duration-100",
+      "motion-safe:hover:scale-110",
+      "motion-reduce:transition-none",
     );
     expect(fourthStar).not.toHaveClass("transition-all");
-    expect(fourthStar.querySelector("svg")).toHaveClass("transition-[fill-opacity]");
+    expect(fourthStar.querySelector("svg")).toHaveClass(
+      "transition-[fill-opacity]",
+      "motion-reduce:transition-none",
+    );
 
     fireEvent.click(fourthStar);
     expect(onChange).toHaveBeenCalledWith(4);
--- a/web/src/components/StarRating.test.tsx
+++ b/web/src/components/StarRating.test.tsx
@@ -1,0 +1,49 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import StarRating from "./StarRating";
+
+describe("StarRating", () => {
+  it("previews the hovered rating and restores the saved rating on leave", () => {
+    render(<StarRating value={2} onChange={() => {}} />);
+
+    const group = screen.getByRole("radiogroup", { name: "Rating" });
+    const stars = screen.getAllByRole("radio");
+
+    expect(stars[1]?.querySelector("svg")).toHaveAttribute("fill-opacity", "1");
+    expect(stars[2]?.querySelector("svg")).toHaveAttribute("fill-opacity", "0");
+
+    fireEvent.mouseEnter(stars[4]!);
+
+    for (const star of stars) {
+      expect(star.querySelector("svg")).toHaveAttribute("fill-opacity", "1");
+    }
+
+    fireEvent.mouseLeave(group);
+
+    expect(stars[1]?.querySelector("svg")).toHaveAttribute("fill-opacity", "1");
+    expect(stars[2]?.querySelector("svg")).toHaveAttribute("fill-opacity", "0");
+  });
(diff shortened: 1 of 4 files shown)


blurbery:
@coderabbitai rate limit

blurbery:
@coderabbitai review

PR #722, opened 2026-08-23
feat(metadata): configure image cache workers

Quick104:
## Problem

Metadata image cache and backfill tasks hardcode their worker count, so deployments cannot adjust their image-processing concurrency.

## Approach

- Read `SILO_METADATA_IMAGE_CACHE_WORKERS` when a metadata image task starts.
- Default to 4 workers. Empty, invalid, zero, and negative values fall back to that default.
- Use the same setting for queued cache drains and manual backfills.
- Keep the claim size at 2, the scheduled drain cap at 10 minutes, and manual backfills unbounded as before.
- Document the deployment override in `.env.example`.

The fixed two-job claim size still caps realized parallelism within one claimed batch at two. The setting can reduce concurrency today; values above two are passed to the runner but do not increase work within a single batch while that claim size remains fixed.

## Validation

Passed:

```text
go test -race ./internal/taskmanager/tasks/...
go build ./...
gofmt -l .
go vet ./...
make verify-local-paths
git diff --check
```

`make test-go` is not green because two existing `internal/jellycompat` process-lock tests fail reproducibly:

```text
TestBeginWebOperationRecoversDeadProcessLock
TestBeginWebOperationRejectsLiveProcessLock
```

`golangci-lint run --new-from-merge-base="origin/main" ./...` was not run because `golangci-lint` is not installed in the validation environment.

No frontend behavior changed, so frontend tests and screenshots are not applicable.

## Risks

An invalid override silently uses the default, matching existing worker-count environment overrides. No API, database, migration, or client contract changes.

## AI Disclosure

- Tool(s): Codex in T3 Code
- Model(s): gpt-5.6-sol
- Involvement: Fully AI-generated; human review pending
- Adversarial review: Reviewed the complete diff and traced the configured value through both task paths into `ImageCacheProcessor`. Verified fallback parsing, the unchanged claim and runtime arguments, and deployment documentation. The review found that the fixed claim size limits realized per-batch parallelism to two; this tradeoff is documented above because preserving the claim size is part of the requested scope.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to configure the number of concurrent metadata artwork cache workers.
  * The default concurrency is four workers when the setting is omitted or invalid.

* **Bug Fixes**
  * Improved handling of empty, non-positive, or invalid worker-count settings by safely applying the default.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -98,6 +98,9 @@ POSTGRES_SHM_SIZE=8gb
 # NODE_NAME=transcode-01
 # NODE_URL=http://transcode-01.example.com:8082
 
+# Optional metadata artwork cache concurrency. Defaults to 4 workers.
+# SILO_METADATA_IMAGE_CACHE_WORKERS=4
+
 # Optional server mode override for source runs or standalone worker nodes.
 # MODE=integrated
 #SILO_MIGRATE_TIMEOUT=60m
--- a/internal/taskmanager/tasks/cache_metadata_images.go
+++ b/internal/taskmanager/tasks/cache_metadata_images.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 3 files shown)


Quick104:
Closing as superseded by #817, which now adaptively sizes metadata image-cache workers from CPU and memory, scales claim batches with the resolved worker count, and enforces per-job timeout/lease safety.

This PR’s fixed two-job claim means worker values above two cannot increase effective concurrency, and rebasing it onto the newer implementation would risk breaking its memory and lease bounds. If an explicit operator throttle is still needed, it should be implemented as a new bounded override on top of #817.

Issue #180, opened 2026-06-18
Match Item dialog shows video provider IDs for ebook libraries


Quick104:
## Problem
When manually fixing an unmatched item from an Ebooks library, the Match Item dialog renders video-centric external ID fields: IMDb ID, TMDB ID, and TVDB ID. Those fields are irrelevant for ebook/comic metadata matching and make the manual repair workflow look unsupported or broken for ebook libraries.

## Repro
1. Create a library with type `Ebooks`, for example a comics library.
2. Add comic or ebook files such as `.cbz`, `.epub`, or `.pdf`.
3. Let autoscan run, or trigger a manual scan.
4. Open the Unmatched Items admin view.
5. Click `Match` on an unmatched ebook/comic, for example `Absolute Batman 001 (2024)`.
6. Inspect the external ID inputs in the Match Item dialog.

## Expected
The dialog adapts to the item/library type. For ebook items it should hide video-only provider fields and either show no external ID override fields or show ebook-relevant IDs supported by the active ebook metadata providers, such as ISBN, Goodreads, ComicVine, or other provider-specific IDs.

## Actual
The dialog always shows IMDb ID, TMDB ID, and TVDB ID.

## Environment
- Observed version: `8b6e53cf`
- Deployment: Docker
- Library type: `Ebooks`
- Observed file type: `.cbz`; likely also affects other ebook formats.

## Validation notes
This appears to be a real product bug, not just a missing enhancement:

- `web/src/components/MatchItemDialog.tsx` initializes and submits only `imdb_id`, `tmdb_id`, and `tvdb_id`, and renders only those three external ID fields.
- `web/src/pages/AdminLibraries.tsx` opens the dialog from Unmatched Items while casting `content_type` to the video-only union `movie | series | season | episode`, so ebook/manga types are not represented correctly at the component boundary.
- `web/src/api/types.ts` models `ItemMatchSearchRequest` with only `imdb_id`, `tmdb_id`, and `tvdb_id`.
- `internal/api/handlers/admin_match.go` accepts only those same three IDs and only injects those providers into the metadata search query.
- `internal/metadata/service.go` chooses the `series` provider chain for search unless the content type is movie/movies, so an ebook match search does not appear to resolve the `ebook` chain even though library setup supports an ebook metadata content level.
- In contrast, ebook enrichment already builds ebook search requests with content type `ebook` and persists generic provider IDs after filtering, so the manual-match path is out of sync with the ebook metadata path.

## Suggested fix direction
[...]

PR #723, opened 2026-08-23
Tokenless V3 playback, DV7 client transforms, admin transcode honesty

Quick104:
Server half of the AetherEngine player migration (client: Silo-Server/silo-apple#180).

## Playback protocol V3

- Tokenless playback: header-authenticated media with signed stream-URL reconstruction, sticky per-attempt feature sets, and tokenless subtitle delivery.
- Distributed egress for tokenless attempts: a second attempt-sticky opt-in, `authorized_media_origins_v1`, lets plans return absolute, still credential-free proxy URLs (`{proxy}/stream/v3/{session_id}` family) for direct play, progressive remux, and node-executed HLS. The proxy is told what to serve out of band — the API writes the session recipe to a Redis grant (`silo:proxygrant:`, sibling of the `noderecipe` handoff), overwritten on replan and revoked on stop/abort/rollback — and authenticates the caller itself: bearer JWT against the live signing secret plus the same `auth_sessions` liveness check the API runs, then ownership against the grant. Revoking a login stops proxy playback immediately, which signed URL tokens never could. Clients that send only `header_authenticated_media_v1`, and deployments with no proxy pool, keep the API-local behavior unchanged; the progressive-remux escalation to HLS now applies only when no proxy origin is available.
- Downloads and auth changes the same flow needs, plus access-group clause test coverage. Detailed download capabilities now fail closed to the coarse `max_resolution` ceiling when a file's probe metadata is too sparse for the per-decoder bounds walk, instead of skipping both checks.

## Admin activity

- Carry `target_audio_channels` through a new migration, session sync, the reconciler, the admin session payload, and the web types. A TrueHD 7.1 source transcoded to AAC 5.1 used to render as "AAC 7.1" because the UI borrowed the source channel count; it now shows the real output layout, or just the codec when the count is unknown.
- Rename the "Audio SW" chip to "Audio Transcode". It labels a plan decision — video copied, audio transcoded — not a client capability.

## Required by the client

Silo-Server/silo-apple#180 negotiates tokenless media and Dolby Vision Profile 7 client-transform grants that only this branch understands. Against an older server, DV7 sources fall back to remux HLS with forced AAC audio.

`authorized_media_origins_v1` is a further, separate client opt-in: without it a tokenless attempt stays entirely on the API origin (the current #180 behavior), so nothing breaks while the client adds support for absolute `/stream/v3/...` plan URLs. The contract is specified in `docs/architecture/playback-protocol-v3.md` §4.1–4.2.

## Testing

- `go build ./...`; `go test` passes for cmd/silo, api/handlers, worker, and playback (including contract and planstore).
- Web `tsc && vite build` passes; adminActivityPresentation tests 14/14.
- Verified on the dev deployment: a DV7 TrueHD remux plans `original_http` with `decision_reason=client_dv7_to_dv81` and audio claim `truehd` — no transcode.
- Proxy egress (`d6ba68b1`): `gofmt`/`go vet` clean; focused suites green for `internal/playback`, `internal/api`, `internal/proxy`, `internal/noderecipe`, `internal/nodepool`, `internal/transcodenode`, `internal/downloads`, including new grant-store, URL-builder, escalation-gating, and proxy-handler auth tests (401/403/404/503 matrix, node-relay token forwarding, mutation-tested local-fallback guard). `make verify-playback-fixtures` and `make verify-local-paths` pass; the regenerated conformance matrix fixes the earlier Go CI failure. Not yet exercised against a live multi-node deployment — unit and handler coverage only.

## Summary by CodeRabbit

* **New Features**
  * Added secure, tokenless playback through authorized media origins for direct, remuxed, and HLS content.
  * Added support for explicitly enabled and validated software video decoding.
  * Download requests now accept detailed video and decoder capability information.
  * Playback and admin activity views display delivered audio channel counts when available.
  * Added recovery for remote transcoding after service restarts.

* **Bug Fixes**
  * Improved playback authorization, ownership checks, session revocation, and rollback behavior.
  * Admin accounts can no longer be assigned access groups, with clearer validation errors.
  * Audio-only transcodes are now labeled “Audio Transcode.”


Quick104:
Review triage for the CodeRabbit and Codex findings — fixes landed in 5875949c.

**Fixed**

- **Codex P1 — grants staged before the replacement commits** (`playback_v3.go:1115`): valid. `putProxyGrantV3` now returns the grant it displaces, the transport rollbacks restore it (so a failed replan no longer 404s the restored plan's proxy URL), and the brief overwrite window before commit is documented as accepted — same session, same user, bounded by the replan.
- **CodeRabbit — stale grant after an API-local replan commit** (`playback_v3.go:1652`): valid. All three commit paths (identity, remote relay, local transcode) now revoke the grant when a proxy-egress attempt commits onto a transport the API serves itself.
- **Codex P1 — escalation gate vs. unusable proxies** (`playback_v3.go:1280`): partially taken. The gate now also requires a usable grant store — a process that can *never* authorize proxy egress escalates to HLS instead of refusing forever. Configured-but-currently-ineligible proxies (saturated, unhealthy) deliberately still get the retryable `capacity_unavailable`: that matches the legacy identity-remux behavior exactly, and retrying is the correct client response to a transient condition.
- **Codex P2 — `target_audio_channels` not advertised** (`playback_sessions.go:68`): valid; the capabilities endpoint now reports it.
- **Codex P2 — invalid `video_evidence` accepted on flat download payloads** (`resolver.go:59`): valid; a present-but-unrecognized tier is now rejected instead of silently resolving from flat claims. An omitted tier stays valid (that is what a legacy payload looks like).
- **CodeRabbit — `SessionUnauthorized` unhandled in `stream.go` / `jellycompat/streams.go`**: added as defensive hardening. Both call sites resolve a non-zero user before loading, so the status is unreachable today — but an unhandled status would nil-deref, so the case is worth its six lines.

**Not taken, with reasons**

- **CodeRabbit — reject `software_video_decode_v1` with an empty `video_decode` list** (`protocol_v3.go:326`): the feature-token-only shape is documented as valid ("feature-token-only payloads stay valid... resolve from the flat codec lists"), and it is fail-safe — with no detailed entries, strict tiers still refuse direct eligibility (`evidence_insufficient_for_direct`), so the token alone widens nothing.
- **CodeRabbit — add `codecs_video_hardware` to the download `ClientCapabilities`** (`resolver.go:39`): the download caps contract never included that field (the old mirror struct didn't have it either), and the additive vocabulary this PR documents is `client_features` + `video_evidence` + `video_decode` only. No existing or planned client sends it on downloads; adding unused surface pre-lock is the wrong direction.
- **CodeRabbit nitpick — Postgres session check per `/stream/v3` request**: accepted as-is for now; it is the exact cost profile of the API's own `RequireAuth` on every media request today. If it shows up under load, a short shared-TTL cache should cover both call sites at once.

**Deferred**

- **Codex P1 — tokenless sessions across API replicas** (`session.go:39`): real, but architectural — playback sessions have always been process-local, and the legacy stream token is what made cross-replica serving work. Tokenless mode inherits the restart trade this PR already documents; 5875949c states the replica-affinity constraint explicitly in the protocol spec (§4.1), including that proxy-origin URLs are unaffected (grants live in shared Redis). The proper fix is shared session state, which is out of scope here and worth its own issue.

Quick104:
Both round-two Codex findings were valid — fixed in 9b09ae37.

- **P1 — tokenless remote transcodes vs. node restart**: confirmed exactly as described; the relay only forwarded `X-Silo-Stream-Token` when the URL carried `?st=`, which tokenless URLs never do. Fixed with the existing jellycompat pattern rather than new machinery: the API writes the transport's recipe card to the shared `noderecipe` store (keyed by the node-facing transport id) at remote-transcode start for header-authenticated attempts, and the node's reconstruct path now falls back to the store when no token header is present — the token was always a recipe source, never that route's authorization (the static bearer is). Recipes are deleted on transport replacement, rollback, and session stop/abort; the TTL backstops a crashed API. Covered by node-side reconstruct tests and handler lifecycle tests.
- **P2 — unused proxy reservation on grant fallback**: fixed with a new `Planner.ReleaseSessionProxy` that drops only the proxy half (job slot + bandwidth charge) while the transcode node keeps its reservation. Applied uniformly — the legacy no-signable-token fallback had the same leak.

PR #726, opened 2026-08-23
fix(requests): encode empty request lists as [] instead of null

### LINKED ISSUE #577: [bug] /api/v1/requests/mine returns null instead of an empty array when no requests exist
## What happened

On a real Docker deployment, opening **Requests** in Silo for iOS 1.0.0 (build 24) displays:

> The server response was in an unexpected format.

The affected account had no media requests. Both Requests endpoints returned HTTP 200, but an authenticated replay of the `mine` endpoint returned:

```json
{"requests":null}
```

The iOS client expects `requests` to be an array, so decoding fails and the entire Requests screen displays an error.

Code inspection indicates that `ListForUser` returns a nil Go slice when no database rows exist. The handler passes that slice directly to the JSON response, producing `null`.

The web and Android clients currently tolerate this response through client-side defaults, but the backend response should consistently represent an empty collection as `[]`.

The attached iOS diagnostic bundle contained zero application log lines.

## Steps to reproduce

1. Run Silo Server using Docker.
2. Enable and configure Requests.
3. Use an account/profile that has no existing media requests.
4. Sign in with Silo for iOS 1.0.0 (build 24).
5. Tap **Requests**.
6. Observe “The server response was in an unexpected format.”
7. Send an authenticated request to `GET /api/v1/requests/mine?limit=200&offset=0`.
8. Observe HTTP 200 with `{"requests":null}`.

## Expected behavior

The Requests screen should show discovery content and an empty “My Requests” state.

When no requests exist, the endpoint should return:

```json
{"requests":[]}
```

## Silo version / commit

`fe11b2601dd339398ea9b74b18d5d0eebe6c16db`

## Deployment

Docker

## Clients affected

- [ ] Web UI
- [ ] Android
- [x] iOS/tvOS/macOS
- [ ] Jellyfin-compat client
- [ ] Other/not client-specific

## Relevant logs

Raw server log line; only the public client IP and session identifier have been redacted for privacy:

```shell
time=2026-08-10T08:07:13.383Z level=INFO msg="api request" component=api component=api request_id=0b6c0613263b/Awzqp6hGT7-007798 method=GET path=/api/v1/requests/mine path_pattern=/api/v1/requests/mine status=200 duration_ms=6 client_ip=[REDACTED] user_agent="Silo/24 CFNetwork/3860.700.1 Darwin/25.6.0" node_id=0b6c0613263b user_id=1 session_id=[REDACTED]
```

Raw authenticated replay output:

```shell
/api/v1/requests/mine?limit=200&offset=0	200
{"requests":null}
```

## AI involvement in this report

AI-assisted, human verified the repro.

## AI tool + model

OpenAI Codex / `gpt-5`

### AI Disclosure

- Tool(s): OpenAI Codex
- Model(s): `gpt-5`
- Involvement: AI-assisted; the human supplied the real-device reproduction and authorized submission.
- Adversarial review: Rechecked the raw server request, replayed the authenticated endpoint, confirmed that the account had zero media-request rows, traced the backend serialization path, and compared iOS behavior with the web and Android clients. No conflicting existing backend issue or pull request was found.

## Confirmations

- [x] I reproduced this myself on a real deployment.
- [x] Logs above are raw copy-paste, not AI-summarized; the two privacy redactions are explicitly identified.

Tobechukwu-Njoku:
Fixes #577.

## What happened

`Repository.listRequests` declares `var out []*Request` and appends per row, so it returns a nil slice when no rows match (`internal/requests/repository.go:440`). Both list handlers passed that slice straight into the response struct, and `encoding/json` encodes a nil slice as `null`:

```json
{"requests":null}
```

Silo for iOS decodes `requests` as an array, so the whole Requests screen fails with "The server response was in an unexpected format" for any account with no requests. The web and Android clients only survive it through client-side defaults.

## The change

Adds `nonNilRequests` and wraps both response sites in it, following the existing `nonNilStrings` / `nonNilUnmatchedSamples` pattern used elsewhere in the codebase (`internal/jellycompat/mapping.go:962`, `internal/historyimport/repo.go:1262`).

`HandleAdminList` shares both the response shape and the nil source with `HandleListMine`, so it had the same defect on `GET /api/v1/requests`. It's fixed alongside — same one concern, and leaving it would have meant a second PR against the identical two lines.

No client changes: this only makes the server honor the contract the clients already assume.

## Tests

Three tests in `internal/api/handlers/requests_test.go`, using the existing `fakeRequestService` (extended with `listMineFn` / `listAdminFn` hooks matching its existing `*Fn` pattern):

- `TestHandleListMineEncodesEmptyRequestsAsArray`
- `TestHandleAdminListEncodesEmptyRequestsAsArray`
- `TestHandleListMinePreservesPopulatedRequests` — guards the non-empty path

The two empty-list tests assert on the **raw response body**, not a decoded slice. `null` and `[]` both decode into a zero-length `[]*Request`, so a `len(...) == 0` check would pass against the buggy code and pin nothing.

Verified they fail before the fix and pass after. Before:

```
--- FAIL: TestHandleListMineEncodesEmptyRequestsAsArray
    requests_test.go:287: body = {"requests":null}, want {"requests":[]}
--- FAIL: TestHandleAdminListEncodesEmptyRequestsAsArray
    requests_test.go:303: body = {"requests":null}, want {"requests":[]}
```

These need no database, so they run in CI rather than skipping.

## Gate

Run against PostgreSQL 18 + Redis from `docker-compose.yml`, Go 1.26.4:

- `go build ./...` — clean
- `gofmt -l .` — no output
- `go vet ./internal/api/handlers/` — clean
- `golangci-lint run --new-from-merge-base="origin/main" ./...` — `0 issues.`
- `make test-go` — `internal/api/handlers` passes
- `cd web && pnpm run lint` — 0 errors (no web files touched)

`make test-go` does report failures in `internal/jellycompat`, `internal/playback`, and `internal/transcodenode`. I checked these against a clean `main` worktree at 91c3d7bd with no changes applied and they fail identically there, so they are pre-existing on this host and unrelated to this change:

| Package | clean `main` | this branch |
|---|---|---|
| `internal/jellycompat` | FAIL (2) | FAIL (2) |
| `internal/playback` | FAIL (4) | FAIL (4) |
| `internal/transcodenode` | FAIL (1) | FAIL (1) |
| `internal/api/handlers` | ok | ok |

They look host-specific rather than broken on main — the `playback` set is `gpudetect` expecting NVENC on an Apple Silicon machine with no NVIDIA GPU, and the `transcodenode` one is a download-prepare timing assertion. Happy to open a separate issue if they aren't already known.

## Note

Written with AI assistance (Claude), per CONTRIBUTING. I've read the change, run the gate above, and can explain the reasoning and alternatives.

## Summary by CodeRabbit

* **Bug Fixes**
  * Request list endpoints now return an empty array when no requests are available, instead of `null`.
  * Existing request data continues to be returned unchanged.

### DIFF
--- a/internal/api/handlers/requests.go
+++ b/internal/api/handlers/requests.go
@@ -263,7 +263,7 @@ func (h *RequestsHandler) HandleListMine(w http.ResponseWriter, r *http.Request)
 	}
 	writeJSON(w, http.StatusOK, struct {
 		Requests []*mediarequests.Request `json:"requests"`
-	}{Requests: requests})
+	}{Requests: nonNilRequests(requests)})
 }
 
 func (h *RequestsHandler) HandleGet(w http.ResponseWriter, r *http.Request) {
@@ -291,7 +291,7 @@ func (h *RequestsHandler) HandleAdminList(w http.ResponseWriter, r *http.Request
 	}
 	writeJSON(w, http.StatusOK, struct {
 		Requests []*mediarequests.Request `json:"requests"`
-	}{Requests: requests})
+	}{Requests: nonNilRequests(requests)})
 }
 
 func (h *RequestsHandler) HandleApprove(w http.ResponseWriter, r *http.Request) {
@@ -627,6 +627,16 @@ func toIntegrationResponses(integrations []mediarequests.Integration) []requestI
 		out = append(out, requestIntegrationResponseFrom(integration))
 	}
 	return out
+}
+
+// nonNilRequests keeps an empty request list encoding as [] rather than null.
+// The store returns a nil slice when no rows match, and clients decoding
+// `requests` as an array fail on null instead of rendering an empty state.
+func nonNilRequests(requests []*mediarequests.Request) []*mediarequests.Request {
+	if requests == nil {
+		return []*mediarequests.Request{}
+	}
+	return requests
 }
 
 func writeRequestServiceError(w http.ResponseWriter, err error) {
--- a/internal/api/handlers/requests_test.go
+++ b/internal/api/handlers/requests_test.go
@@ -20,6 +20,8 @@ type fakeRequestService struct {
 	listNetworksFn func() ([]mediarequests.DiscoverBrandCard, error)
 	listGenresFn   func() ([]mediarequests.DiscoverBrandCard, error)
 	browseFn       func(kind, slug string, mediaType mediarequests.MediaType, sort string, page int) (*mediarequests.DiscoverBrowseResponse, error)
+	listMineFn     func() ([]*mediarequests.Request, error)
+	listAdminFn    func() ([]*mediarequests.Request, error)
 }
 
 func (f *fakeRequestService) ListStudios(context.Context, mediarequests.Viewer) ([]mediarequests.DiscoverBrandCard, error) {
@@ -76,10 +78,16 @@ func (f *fakeRequestService) CreateRequest(context.Context, mediarequests.Viewer
 }
 
 func (f *fakeRequestService) ListMine(context.Context, mediarequests.Viewer, mediarequests.ListFilter) ([]*mediarequests.Request, error) {
+	if f.listMineFn != nil {
+		return f.listMineFn()
+	}
 	return nil, nil
 }
 
 func (f *fakeRequestService) ListAdmin(context.Context, mediarequests.Viewer, mediarequests.ListFilter) ([]*mediarequests.Request, error) {
+	if f.listAdminFn != nil {
+		return f.listAdminFn()
+	}
 	return nil, nil
 }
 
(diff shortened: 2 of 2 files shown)


tomislav:
A PR already exists for this. #578

Tobechukwu-Njoku:
You're right — thanks for the quick catch, and apologies for the duplicate work.

I checked #577 for assignees and scanned the recent open PRs, but #578 predated the window I looked at and I never checked the issue timeline for cross-referencing PRs, which is exactly where it shows up. My mistake.

#578 is also the better change: extracting a named `requestListResponse` type removes the duplicated anonymous struct at both call sites, where mine left it in place. Closing this in favour of it.

PR #733, opened 2026-08-23
feat(scanner): persist H.264 copy-safety verdicts and move analysis off browse paths

Quick104:
## Problem

The H.264 multi-PPS copy-safety scan (the check that decides whether a file can be stream-copied into fMP4 HLS without corrupting VideoToolbox decoders) ran while media detail pages loaded, and its result lived only in process memory. Every restart re-read the opening seconds of every browsed H.264 file. On remote/cloud storage that made first-time browsing take seconds per page — reported on Discord by a user with a remote library, and reproducible on shared dev's ceph-backed media.

## Approach

- **Persist the verdict** in three nullable `media_files` columns: `multiple_pps` plus the file size and mtime it was computed from. The verdict is *self-validating* — trusted only while the stored size/mtime still match the row — so replacing or re-encoding a file invalidates it with zero writer coordination. `Upsert` deliberately never touches these columns; writes go through a narrow targeted UPDATE (`UpdateMultiplePPS`), because `Upsert` has side effects (clears `match_suppressed_at`/`missing_since`) a scan has no business triggering.
- **Deliberately not stored in the `video_tracks` jsonb**: `models.VideoTrack` marshals directly into v1 API detail responses (no DTO layer), and several paths rebuild that jsonb from ffprobe — a tag change would both leak the field to clients and lose it on reprobe.
- **Browse pages never scan**: detail/episode/extra builders use a new `EnsureProbeOnly` (probe repair only). The scan runs from the watch page and playback start — the surfaces where a play is actually in the picture. The verdict is invisible in browse responses anyway, so scanning there was pure warm-up.
- **Scan window 15s → 5s**: affected encoders emit every PPS variant within the opening GOPs; the cost is bytes read off the media store, not demux CPU.
- **Singleflight** on concurrent first scans of the same file (previously a burst of requests each spawned ffmpeg).
- The lazy path stays strictly fail-closed and stateless: scan errors persist nothing and disable stream copy for that request only.

## Behavior for existing deployments

Purely additive migration; every existing file starts unanalyzed, is scanned once on first watch/play exactly as before, and is then remembered forever (until its bytes change). No client-visible API change; jellycompat unaffected.

## Validation

Deployed to shared dev (1.76M-file library, ceph-backed): migration applied cleanly, detail/watch endpoints return in ~400–500 ms on cold unscanned h264 files, verdicts persist with valid size/mtime bindings, revisits reuse them with no re-scan. Full gate: `go build`, scanner/catalog/playback/api suites, `gofmt`, `golangci-lint --new-from-merge-base` (0 issues), `make migrate-validate`, plus a from-zero migration chain + round-trip check against a scratch Postgres.

Related issue: N/A — narrow fix (originated from a Discord report; design discussed there).

## AI use disclosure

Implemented by Claude Code (maintainer-directed); design, review, and live validation supervised by @Quick104.

## Summary by CodeRabbit

* **Performance**
  * H.264 stream analysis now runs during playback preparation instead of page loading.
  * Reduced the analysis window from 15 seconds to 5 seconds for faster preparation, especially with remote media.

* **Reliability**
  * Analysis results are reused when file size and modification time are unchanged.
  * Stale or missing results are rescanned automatically, with retries after failures.

* **Bug Fixes**
  * Browsing and detail views no longer perform unnecessary full playback copy-safety scans.

### DIFF
--- a/docs/feature-changelog.md
+++ b/docs/feature-changelog.md
@@ -2,6 +2,13 @@
 
 ## 2026-08-23
 
+### Browsing no longer waits on H.264 stream-copy analysis
+Silo checks each H.264 file once for a bitstream quirk that makes stream-copying unsafe. That check reads the opening seconds of the file, and it used to run while a media page was loading and be forgotten on every restart — so browsing a library, especially after a reboot, re-read part of every H.264 file. On remote or cloud storage that was the difference between an instant page and a slow one.
+
+Three things change. Media pages no longer trigger the analysis at all; it now happens when a play is actually being prepared, so browsing is fast regardless of where the files live. The result is stored on the file instead of being kept only in memory, so it survives restarts and is computed at most once per file. And the check itself reads 5 seconds instead of 15.
+
+A file that changes on disk is re-checked automatically: the stored answer is only trusted while the file's size and modification time still match, so re-encoding or replacing a file in place invalidates it without any manual step. Nothing is recorded when an analysis fails, so a transient error never turns into a stale verdict — the next request simply retries. No configuration changes, and playback behavior is unchanged.
+
 ### Serve tokenless playback from proxy nodes again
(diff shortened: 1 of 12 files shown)


Quick104:
Consolidated into #734 at maintainer request — single PR for review and merge.

PR #742, opened 2026-08-24
feat(images): client-selectable artwork size via image_size

Quick104:
## Problem

On large TVs, posters and backdrops look blurry. The v1 API hardcodes which artwork variant every client receives — w300 cards and w500 posters regardless of screen — and the stored variant ladder tops out at w500 for posters, stills, and logos. A 4K TV upscaling a 500px poster is exactly the reported blur. Clients had no way to ask for more, and nothing larger existed to serve.

## Approach

Two halves, one contract:

- **Larger stored variants.** The ladder (`internal/artworkkey`, the single source of truth for generation, key expansion, and GC) gains w780 for posters and stills and w1280 for logos; backdrops already had w1920. A new `LadderVersion` constant triggers a one-shot, crash-safe background regeneration pass (singleton state table, bounded batches after the scheduled image-cache drain) so existing libraries gain the new rungs without operator action. Until a rung is generated, a presign-time S3 existence check transparently serves the next lower rung (exists cached 24h, missing 15m, errors optimistic and uncached, fallback URL cache clamped so the real rung is picked up promptly).
- **Client-selectable size.** A new leaf package `internal/imagesize` owns the size vocabulary and every (image type, size) → variant decision. Catalog, detail, seasons/episodes, sections, and the personal lists (favorites/watchlist/history) accept `image_size=small|medium|large|original`; the server bakes the chosen variant into the URLs it returns. An absent parameter is byte-identical to today's behavior (pinned by regression tests); an invalid value is a 400. `GET /api/v1/images/capability` advertises the parameter, sizes, and per-type widths live from the ladder, following the existing capability-endpoint pattern.

jellycompat parity: Jellyfin-protocol clients don't send `image_size`, but `compatRequestImageSize` gains a large bucket so `MaxWidth` requests in the 780–1200px range now map to the new rungs instead of jumping between w500 and original.

Docs: new `docs/images-api.md`, `docs/feature-changelog.md` entry. Purely additive — no entry needed in the v1-scope removals table.

## Supersedes #336

This PR supersedes and closes #336 by @raiden202, whose approach it deliberately carries forward: the ladder-rung additions for stills and logos and, in particular, the S3 existence-check fallback with asymmetric TTLs originated there. This re-implementation extends the idea to posters (the most visible blur), adds the client-facing `image_size` parameter and capability endpoint, and adds the automatic backfill. Credit to @raiden202 for identifying the problem and the fallback design.

## Coordinated client work

- `silo-apple`: tvOS sends `image_size=large` on all artwork-bearing surfaces, gated on the capability probe (404 → off). iOS/macOS unchanged.
- `silo-android`: Android TV likewise via a DI-registered preference + capability probe; phone unchanged.

Client PRs are linked in the comments below. Known follow-ups: the plugin SDK has no "large" semantic tier ( `large` maps to `featured` for plugin-resolved images) — needs a `silo-plugin-sdk` decision; the web frontend does not yet send the parameter.

## Verification

- `go test` across `imagesize`, `artworkkey`, `catalog`, `metadata`, `api`, `imagecache`, `jellycompat`, `taskmanager` — pass (two pre-existing `internal/jellycompat` process-lock test failures reproduce identically on `main`).
- Regression tests pin absent-param output byte-for-byte and the historical variant matrix for every image type.
- `make build` passes; `golangci-lint run --new-from-merge-base=main` — 0 issues; `make verify-local-paths` clean.

Related issue: N/A — supersedes #336 directly.

## AI-use disclosure

Implemented by Claude Code (maintainer-directed) with human review. The design was reviewed and approved by the maintainer before implementation; the diff was machine-reviewed for correctness, conventions, and cross-repo contract consistency, and findings were addressed.

## Summary by CodeRabbit

* **New Features**
  * Added client-selectable artwork sizes through `image_size` across catalog, library, favorites, watchlist, history, and media responses.
  * Added an image capability endpoint describing supported sizes and artwork variants.
  * Added broader artwork size options, including large logos and profile images.
  * Added automatic fallback to narrower artwork when the requested size is unavailable.
  * Added background regeneration for existing artwork caches to support the updated size ladder.

* **Bug Fixes**
  * Invalid image-size values now return HTTP 400 errors while preserving defaults when omitted.

* **Documentation**
  * Added guidance on image sizing, capability discovery, compatibility, and supported artwork widths.


Quick104:
Coordinated client PRs: Silo-Server/silo-apple#187 (tvOS), Silo-Server/silo-android#245 (Android TV).

Quick104:
Plugin-side adoption of the `large` variant tier: Silo-Server/silo-plugin-sdk#17 (canonical variant constants + contract docs), Silo-Server/silo-plugin-metadata-tmdb#16 (large → real TMDB sizes). The server now sends `variant: "large"` for plugin-resolved images (commit 10 on this PR); plugins without a large arm gracefully fall back per the SDK contract.

Quick104:
Addressed the CI failure and all review findings in d0a76b1b + 977c0c9a:

- **Lint (goconst ×11)**: named the two string vocabularies instead of suppressing — `artworkkey.ImagePoster/Backdrop/Still/Logo/Profile` and `imagesize.PluginVariant*` constants, now used at every former literal site. A knock-on `w780` finding was resolved by deriving the existence-check rung from the ladder (`imagesize.Variant(type, Large)`) so it can no longer drift from `VariantWidths`. `golangci-lint v2.12.2 --new-from-merge-base=origin/main` is clean.
- **Card backdrops (Major)**: new `sizedCardBackdropPath` reads the image type off the key, applied at both `items.go` sites — and at two equivalent spots in `favorites.go` that had the same latent bug. Regression tests cover the episode-still-as-backdrop ladder.
- **Cast/crew photos**: `personCredits` now threads `filter.ImageSize` through the profile ladder; absent-param output is byte-identical (pinned by test).
- **Task progress**: phases now map onto disjoint stretches of the bar; a monotonicity test asserts percents never decrease within one Execute.
- **Docs**: fence language tags added. The person-endpoint suggestion was *corrected* rather than applied: `/people` endpoints genuinely don't take the parameter (the only handler that threads it, `HandleGetPersonItems`, is unrouted dead code superseded by `catalog?source=person`), so the doc now states that explicitly. Follow-up decision: route or delete `HandleGetPersonItems`, and whether `/people` should honor the param.

Quick104:
Addressed all ten Codex findings in 14674a9c, 8a5cac49, ec6236cc:

- **Backfill bookkeeping cluster (cutoff re-scan, multi-node zero-candidate race, rolling-upgrade poisoning, deferred-failure latch)**: replaced job-state bookkeeping with an artwork-state predicate — a row is a candidate iff its `artwork_revision_gc_candidates.object_keys` manifest lacks the new rung (no manifest row counts as missing). Completion is a separate `HasLadderBackfillRemaining` query with no job join, so in-flight/failed/old-node work never latches the version early; an old node's re-cache without the new rungs simply leaves the row a candidate. `completedBefore` is gone. Consequence: a deployment holding permanently-unregenerable artwork never latches "done" — the sweep is paced at 15-minute intervals (persisted `last_attempt_at`) so the steady-state cost is one short-circuiting EXISTS per interval.
- **Serial S3 HEADs**: batch existence checks now run through an errgroup capped at 12, only for keys that can actually miss.
- **file:// sidecar artwork**: included in the backfill (the processor re-caches sidecars; only genuinely non-recachable schemes stay excluded).
- **audiobook-groups**: `image_size` now threads into the cover-stack resolution.
- **jellycompat**: the large bucket only promotes on width-ish constraints (`MaxWidth`/`FillWidth`); height-only requests keep prior behavior. Person photos resolve on the profile ladder, fixing the dead `profile/w780` presign.

Lint (golangci-lint v2.12.2, merge-base mode) is clean; all touched packages pass. Noted for merge: the new SQL predicates are unit-tested for rung-pattern semantics but not integration-tested against live Postgres from this environment.

PR #770, opened 2026-08-26
feat(streamtelemetry): publish what clients claim, and merge it with what was measured

### LINKED ISSUE #666: [bug] Progress updates alone keep dead sessions alive forever — ghost sessions hold transcode slots for 15h with zero byte flow

CoffeeKnyte:
The admin session list previously showed what clients reported without combining it with measured delivery. This adds a reporting publisher alongside each API process’s measuring publisher and serves a merged live view, so operators can distinguish reported playback, measured delivery, and uncertain evidence.

Related issue: #666. Extends #667, which introduced byte measurement. This PR adds visibility; it does not terminate sessions or resolve #666’s underlying playback lifecycle bug.

## Behavior

- Each API process publishes reported playback under its measuring publisher ID plus `#reported`. Reported state cannot supply byte counts or viewer addresses. Provenance is rebuilt from allowed fields, and only viewer-egress measurements contribute delivery evidence.
- Publishers declare their coverage and reporting companion. Missing, stale, or undeclared coverage prevents a complete view. `LocalHub` supports both publishers in single-process deployments; the publisher limit is 512. Wire additions remain decodable by older publishers.
- `GET /api/v1/admin/sessions/live` starts from telemetry and joins PostgreSQL display fields by session ID. The existing `/admin/sessions` response remains unchanged. Clients detect support through `stream_telemetry_live_sessions` on `/admin/sessions/capabilities`. The contract is documented in `docs/admin-api.md`.
- Admin dashboard widgets, Activity, and the sidebar consume the live view. Delivery badges distinguish evidence and view health; operators can reveal rows hidden by healthy idle classification.
- Transfers retain per-viewer identity using hashed IDs, without putting raw IP addresses or device IDs in Redis field names. Per-subject overflow preserves role and class while reducing detail. Transfer-cap exhaustion is an advisory rather than a reason to suppress classification across the fleet.
- Bitrate probes carry a separate class and are excluded from media egress totals. Interrupted stream copies check the cut between slices on both HTTP/1.1 and HTTP/2 and record the interruption. Parity checks compare independent measured evidence and withhold agreement when the comparison is incomplete or self-derived.

## Review fixes

The latest changes address all three findings from the review of `baa7f269`:

1. **Preserve potentially live sessions when a reporter is missing or stale.** A paused or fully buffered session can outlive its retired byte measurement. Both idle classifications now clear under an unhealthy view, including for pruned measurements. The client badge helper follows the same rule. An integration regression test traces a paused session through the telemetry merge and handler when its reporting publisher disappears.
2. **Refresh live membership without relying on legacy playback events.** Live-session queries poll every 30 seconds while the page is visible, focused, and not frozen. Default and revealed variants participate in dashboard refresh. A separate 60-second dashboard timer prevents session polls from postponing other sections and retries after transient failures. Automatic ticks skip refreshes while the previous automatic refresh promise is pending; the guard clears on success or failure. Tests cover both query variants, pause/resume, manual refresh, repeated dashboard refreshes, and failure recovery.
3. **Show telemetry health when the session list is empty.** Starting, partial, and stale telemetry keep the Now Playing widget expanded with its source badge. Only a trustworthy empty result, or telemetry-disabled legacy mode, displays the normal idle state. Regression tests cover all three unhealthy states.

Apple and Android were checked for consumers of the admin live endpoint and capability; none were found. These fixes do not change native-client or Jellyfin playback contracts.

## UI evidence

These screenshots render the production Now Playing component and styles with a synthetic missing-publisher response. Desktop and mobile checks verified the partial badge, reveal/hide interaction, and absence of browser errors. This is component-level browser validation, not a deployed full-stack test.

| Before | After |
| --- | --- |
| Empty partial view incorrectly presented as idle | Empty partial view retains its health badge |

Mobile partial-view state

## Validation

Automatic-refresh overlap follow-up (`92222587d`): all five dashboard tests, TypeScript build-mode checking, targeted ESLint, Prettier, and diff whitespace checks passed. Both new deferred-request cases failed before the guard and passed afterward; they cover settlement after success and failure while preserving manual refresh availability. CodeRabbit CLI was unavailable locally. The required ponytail Git gates passed.

Full validation for the preceding fix commit (`9cc419fcd`):

- `make test-go` passed. The first run timed out in two untouched policy tests; both passed individually and the full rerun passed without changes to those tests.
- `go test -race ./internal/api/handlers ./internal/streamtelemetry/...` passed.
- `make test-web`: **364 files, 3,108 tests passed**. The new regressions reproduced the reported defects before the fixes.
- Go build and vet passed; `golangci-lint run --new-from-merge-base=origin/main ./...` reported **0 issues**.
- Web build, typecheck, formatting, and lint passed; full lint retains existing warnings with no errors.
- Settings bindings, playback fixtures, local-path hygiene, and diff whitespace checks passed.
- Independent correctness and ponytail reviews completed with no remaining actionable findings.

Earlier feature validation included route-manifest goldens, multi-publisher Redis integration tests, race tests for telemetry, dashboard metrics, handlers and Jellyfin compatibility, and HTTP/1.1 versus HTTP/2 interruption tests.

The original author’s production soak at `1c1a3a7f` ran for 24 hours, August 24–25, 2026 UTC: 288 five-minute samples, healthy in every sample, zero restarts, zero dropped observations or bytes, and no publisher truncation. Reported versus legacy membership matched in 283/288 samples; the remaining differences were one row at capture boundaries. The sampler lost 38 measured-session samples because NUL-containing transfer field names confused its text processing; measured-population analysis used the remaining 250. That investigation identified the transfer-ID privacy defect corrected here.

## Risks and follow-up

- The production soak predates the later review fixes and used one publisher pair. Failover, epoch changes, and later changes have test evidence rather than a new production soak. No second soak or production deployment was performed for this update.
- Under an unhealthy view, retired rows can reappear because a missing reporter may still own playback. Counts are uncertain; the view-health warning must stay visible.
- Live membership can take up to 30 seconds to refresh, in addition to the 45-second idle window. Recently ended sessions may briefly lack PostgreSQL display metadata. Measurement tombstones expire after 35 minutes by default; fully buffered playback beyond that memory horizon remains a classification limitation.
- Reported-versus-legacy agreement checks faithful publication, not whether the client’s claim is correct. Paused sessions with dead control sockets remain outside idle classification. This PR performs no new termination or concurrency enforcement.
- Jellyfin direct play without a play-session ID can still deliver bytes without a session-manager row. Visibility improves, but its concurrency-accounting gap needs separate work.

## AI-use disclosure

- Earlier implementation and review fixes: Claude Code, `claude-opus-5` and `claude-fable-5-1`, as disclosed by the original author. Their soak figures above are retained from captured production evidence, not newly measured in this update.
- Current review, regression fixes, tests, and PR update: `gpt-6-astra` through Codex desktop, with an independent Codex review agent. Browser validation used Playwright with Google Chrome; screenshots were uploaded through the authenticated GitHub UI.
- The work was AI-assisted. Test results and browser observations above were obtained from actual runs; production-soak limitations are stated separately.

## Summary by CodeRabbit

* **New Features**
  * Added a telemetry-backed live sessions view at `/api/v1/admin/sessions/live`.
  * Admin screens now show delivery rates, byte counts, delivery status, identity conflicts, and session evidence.
  * Added controls to reveal sessions with measured activity but no currently claimed viewer.
  * Added reporting for measured versus reported-only sessions, view completeness, and telemetry advisories.
  * Bandwidth probes are identified separately and excluded from media delivery totals.

* **Bug Fixes**
  * Improved live-session classification, stale-view handling, and stream interruption reporting.
  * Parity comparisons now withhold agreement when evidence is incomplete or unreliable.

* **Documentation**
  * Expanded API and telemetry documentation for live sessions, completeness, advisories, and truncation.


Quick104:
**Code review — confirmed issues.** Multi-agent review of the full diff; every finding below was independently verified against the PR head (call sites traced, `go build` clean, focused tests pass). Review performed with Claude Code.

### Correctness / behavior

1. `internal/api/handlers/admin_live_sessions.go:283` — `no_delivery` is computed from the view's current `ViewerBytes`, which the measuring registry prunes after 5 minutes of transport idleness (`registry.go:564`) while `ReportedSessions()` keeps publishing the session. A healthy fully-buffered session (fast direct play, audiobooks, music) degrades to Reported-with-0-bytes, is past the 30 s grace, and gets classified as a ghost and hidden from the default list while the user is actively watching. ABS sessions are guaranteed victims: `abs/native_sessions.go:112` hardcodes `UpdateProgress(..., false)`, so they can never report paused.

2. `internal/streamtelemetry/registry.go:436` — `transferKey` now includes the client-supplied DeviceID and viewer IP, so one client rotating device ids (or ordinary CGNAT/IP churn the old key deliberately folded) mints unbounded transfer records. `MaxTransfers` is unchanged at 10,000; exhaustion marks the snapshot `Truncated` → `publisher_truncated` → `view.Complete=false`, and `decorateLiveSessions` then forces `NoDelivery=false` on every row — a single authenticated client can switch off the ghost detection this PR ships, fleet-wide.

3. `internal/api/handlers/playback_sessions.go:286` — the `LIMIT 200` bound is skipped for the `SessionIDs` path, and the live handler passes every id in the merged view — bounded only by `MaxMergedSessions` (50,000) — into one `ANY($n)` query with ~7 LEFT JOINs, executed on every `GET /admin/sessions/live`.

4. `web/src/components/RealtimeEventsProvider.tsx:319` — `hydrateSessions` now fires `invalidateQueries(liveSessionsRoot)` on every realtime sessions event, so every open admin dashboard refetches the endpoint above roughly every reconciler tick (~15 s, plus start/stop triggers) while anything is playing. The old `setQueryData` push path made zero network requests.

5. `internal/api/handlers/admin_live_sessions.go:220` — every normally-ended session lingers in the live list for up to 5 minutes as a measured-only phantom wearing the red "unclaimed" badge: the reporter drops it within ~1 s of stop and the Postgres row is deleted promptly, but the registry keeps its bytes until retention expiry. `NoDelivery` stays false so `include_idle` cannot hide it, and the `sessions.replaced` event from the row deletion itself triggers the refetch that displays it.

6. `internal/streamtelemetry/global.go:289` — the `missing_reported_publisher` gate cannot detect the rolling-deploy case it is documented to cover: an un-upgraded process publishes `ReportingPublisherID == ""` (the field doesn't exist in old code), the gate's `continue` skips it, and the view reports Complete while every paused and pre-delivery session that process owns silently vanishes. The mechanism only detects an upgraded process whose own reporter hasn't ticked yet.

7. `web/src/lib/sessionTelemetry.ts:48` — the destructive "unclaimed" badge is derived solely from `evidence === "measured"` with no gating on view completeness or session freshness; StreamCard never receives the envelope, which stops at NowPlayingSection. During a rolling deploy or before a reporter's first tick, healthy streaming sessions render the red badge — exactly the conditions where the server deliberately suppresses the mirror-image `no_delivery` classification.

8. `internal/streamtelemetry/parity.go:46` — `LiveSessionsFromGlobalView` gained no Reported/evidence filter, so the parity endpoint's "telemetry" side now includes sessions injected by the reporting publisher — the same session-manager source the legacy projection (`playback_sessions_sync`) is derived from. A #666 ghost that pre-PR showed as `legacy_only` now lands on both sides and reads as agreement, making the legacy-retirement gate partially self-confirming; neither code nor docs acknowledge the change.

9. `internal/api/handlers/admin_live_sessions.go:226` — the incomplete-view guard only covers missing publishers. With `SILO_STREAM_TELEMETRY_FAMILIES` scoped to a subset of route families, `Observe` returns the unwrapped handler for the rest (`writer.go:20`), nothing carries the observed-family set into the merge, and the view stays Complete — after 30 s every session on an unobserved family is flagged `no_delivery`, counted, and hidden while streaming normally.

10. `internal/streamtelemetry/global.go:762` — `normalizeProvenance` strips only Routes, ViewerIPs and BytesAccepted from a `#reported` publisher, but `mergeSession` still unconditionally folds its `LastByteAccepted`, `OpenObservations`, `RequestCount`, DeviceIDs, UserAgents and Outcomes — so a buggy or compromised claims-only reporter can still fabricate measured-looking liveness (`last_byte_at` seconds ago, open observations on a zero-byte session), which the comment and `docs/admin-api.md` claim it is unable to do.

11. `web/src/pages/AdminDashboard.tsx` — only the dashboard was repointed to `/admin/sessions/live`; AdminSidebar, ServerActivity, AdminActivity (the target of the dashboard's "View all N streams" link) and AdminStats still read legacy `/admin/sessions`, so the admin surfaces disagree about which sessions are live and how many there are.

### Cleanup (confirmed, lower priority)

- The telemetry-disabled and view-unavailable legacy-fallback blocks in `admin_live_sessions.go` (118–131 vs 141–152) are line-for-line copies — extract a serve-legacy helper.
- `transferKey` hand-rolls truncated SHA-256 + hex instead of the package's existing `digest128` (`store_redis.go:45`).
- `ReportedPublisher.Start/Stop` duplicates `Registry.Start/Stop`'s ticker/shutdown lifecycle nearly line-for-line.
- `mergeSession` calls `normalizeProvenance` twice per contribution (once discarded for the viewer-edge scan).
- `LiveSnapshot` is a single-field wrapper around `map[string]LiveByteFacts`; every caller immediately unwraps `.Facts`.
- `no_delivery_shown` in the response just echoes the request's own `include_idle` back; the web client already holds that state (its query fallback even synthesizes the field client-side).
- `sessionTelemetry` serializes six fields (`RelayBytes`, `OpenObservations`, `RequestCount`, `RealtimeAlive`, `Publishers`, `ViewerEdgePublishers`) that no client reads.

CoffeeKnyte:
**Review disposition.** Both reviews were raised against `1c1a3a7f`. Each finding below names the commit that closes it, on the rebased head `55fa23bf`.

### CodeRabbit inline

| # | Finding | Closed by |
|---|---|---|
| 1 | `CopyChunkedUntil` returns from both stop branches without calling `record`, so a cut can classify as `OutcomeCompleted` | `ede4bd93` — both paths call `record(0, err)` (`internal/httpstream/readfrom.go:73-83`, `:104-111`). Covered by `TestCopyChunkedUntilRecordsTheStopError`, alongside the existing `TestObservedWriterReadFromCutBehavesTheSameOverH1AndH2` |
| 2 | `normalizeProvenance` leaves `ViewerIPs` on a non-edge relay, and delivery-derived fields on a `#reported` publisher | `ede4bd93` — the reporting branch is now an **allowlist rebuild** rather than a list of fields to zero, and `ViewerIPs` is gated positively on a `RoleViewerEgress` route being present, not on a relay role being absent |
| 3 | Rate contributor key fingerprints all publishers, not the byte-contributing ones | `ede4bd93` — `livesessions.go` keys on `session.ViewerEdgePublishers` |
| 4 | A concurrent refresh can attach a newer view's rate to older facts | `ede4bd93` — `viewcache.go` compares `c.status.Refreshes == status.Refreshes` while holding `c.mu` and omits rates if the generation moved |
| 5 | "View all 0 streams ›" in the reveal-only state | `ede4bd93` — the link renders "View activity ›" when the list is empty. Now lives in `NowPlayingWidget.tsx` after the rebase (`55fa23bf`) |

### Correctness findings

| # | Finding | Closed by |
|---|---|---|
| 1 | Pruned `ViewerBytes` on a healthy buffered session degrades it to a ghost and hides it | `94beb01b` — retired measurements are remembered as `MeasurementPruned`, and the classification reads remembered delivery rather than only current bytes |
| 2 | Per-viewer `transferKey` lets one client exhaust `MaxTransfers` and switch ghost detection off fleet-wide | `dd57fd96`, `0e154d73` — `MaxTransfersPerSubject` plus a per-principal `overflowTransferKey` catch-all, and `TransfersTruncated` is demoted from an incompleteness **reason** to an **advisory**, so a client-reachable cap can no longer suppress classification |
| 3 | The `LIMIT 200` bound is skipped on the `SessionIDs` path | `ede4bd93` — `playback_sessions.go` binds `LIMIT $n` to `len(SessionIDs)` |
| 4 | Every realtime sessions event invalidates the live query | `5dcfa068`, `7f49ac31` — a throttle set deliberately wider than the reconciler's own tick, so sub-tick bursts coalesce |
| 5 | Ended sessions linger as measured-only phantoms wearing the "unclaimed" badge | `94beb01b` — ended-session phantoms are suppressed and counted as `unclaimed_idle`, which `include_idle` can now reach |
| 6 | The `missing_reported_publisher` gate cannot see an un-upgraded process | `8bfec3bb` — a publisher **declares its coverage**; a contribution that declares none adds `partial_family_observation`, which is exactly the un-upgraded case |
| 7 | The "unclaimed" badge is derived from `evidence` alone, with no completeness gating | `d2931dc2`, `449c13d7` — the badge reads the server's `unclaimed_idle` and is gated on `viewBlind`, mirroring the server's own `!viewComplete \|\| viewStale` suppression |
| 8 | The parity endpoint's telemetry side now includes reported sessions, so it partly agrees with itself | `ff66f1cd`, `3db8f1d9` |
| 9 | A partial `SILO_STREAM_TELEMETRY_FAMILIES` leaves the view Complete while flagging live sessions | `8bfec3bb` — same coverage declaration as #6 |
| 10 | A `#reported` publisher can still fabricate measured-looking liveness through `LastByteAccepted`, `OpenObservations`, `RequestCount` | `ede4bd93` — the allowlist rebuild in CodeRabbit #2 makes those unrepresentable rather than merely unset |
| 11 | Only the dashboard was repointed; the sidebar and activity page still read legacy `/admin/sessions` | `5dcfa068`, `a5930926`, `7a17040a`, and `55fa23bf` after the rebase — see below |

### Cleanup

Closed: the duplicated legacy-fallback blocks are one `serveLegacyLiveSessions` helper, `transferKey` uses the package's `digest128`, `LiveSnapshot` is the map itself, and `mergeSession` normalizes once per contribution — all in `ede4bd93`/`dd57fd96`.

`ced0f77b` closes the last one: `Registry` and `ReportedPublisher` had hand-rolled the identical ticker / stop-channel / wait / leave-the-roster sequence, ~50 lines across two files. Both now drive a shared `publishLoop` (`internal/streamtelemetry/publishloop.go`), which owns the exactly-once `Leave` — the part most worth having in one place, since a publisher that stops without leaving shows operators ghost sessions followed by a degraded view. No behavior change.

**Two skipped, with reasons:**

- *Six fields no client reads* — `relay_bytes`, `open_observations`, `request_count`, `realtime_alive`, `publishers`, `viewer_edge_publishers` are documented operator diagnostics (`docs/admin-api.md`). The web UI is not the only reader of an admin endpoint; an operator reading it with `curl` is the case they exist for.
- *`no_delivery_shown` echoes the request's `include_idle`* — no longer true. The legacy fallback path forces it true regardless of the request, because nothing was classified there and nothing was withheld, so the field now carries information the client cannot derive. `sessionTelemetry.ts` reads it.

### Since the review: rebase onto `main`

Finding #11 came back from a different direction. `main` rebuilt the admin dashboard as a widget grid and moved Now Playing into `NowPlayingWidget.tsx`, which — along with `ActiveStreamsStatWidget`, `TranscodeShareStatWidget` and `RecentActivityWidget` — read the legacy `/admin/sessions`. `55fa23bf` repoints all four at `useAdminLiveSessions`, so the dashboard, sidebar and activity page agree again.

That commit also fixes the web typecheck. `tsc -b` runs as part of `pnpm build` and this branch had been failing it since before the rebase: mock factories declared zero parameters while being called with the include-hidden flag, two response literals omitted the required `unclaimed_idle_*` fields, and the live-session tests indexed `FakeWebSocket.instances` under `noUncheckedIndexedAccess`. It went unnoticed because the root `web/tsconfig.json` is a solution file with `"files": []`, so `tsc -p tsconfig.json` silently checks nothing — use `tsconfig.app.json`.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` — clean
- `golangci-lint run --new-from-merge-base=origin/main ./...` — **0 issues**
- `go test ./...` green except `TestResolveCopySeekAnchorMatchesRealLongGOPHEVC`, the known ffmpeg ≥5.x requirement against this host's 4.4.2. Fails identically on `origin/main`.
- Web: `tsc --noEmit -p web/tsconfig.app.json` clean apart from packages this host's partial `node_modules` lacks; `eslint` 0 errors; `vitest` 210 tests green across every file this branch touches.

### AI-use disclosure

- Tool(s): Claude Code
- Model(s): claude-opus-5
- Involvement: AI-assisted. Every finding above was re-checked against the pushed head rather than assumed closed by its commit subject; the two skips are judgement calls stated as such.

CoffeeKnyte:
**Parking this PR. Recreate it on API v2 once v2 is finalized.**

Since this branch was cut, `main` moved the web client's admin session calls to `/api/v2/admin/sessions` (#1075). This PR builds its live view on v1 (`GET /api/v1/admin/sessions/live`), and the dashboard, Activity and sidebar read from it. A rebase conflicts in `router.go`, `admin.go`, `playback_sessions.go`, `stream_telemetry_parity.go`, `docs/admin-api.md` and four web files. Resolving those conflicts wouldn't finish the job, though: the endpoint and web hooks would also have to be ported to `internal/apiv2` and the v2 client. Rather than port against a surface that is still changing, the plan is to rebuild the feature on v2 once v2 is complete. The summary below is what the rebuild needs to cover.

## Purpose

#667 measures the bytes each stream actually delivers. The admin session list still shows what clients *report*, and those two can disagree. #666 is the clearest case: a client keeps posting progress updates long after it stops fetching media, so the session stays "playing" and keeps its transcode slot for hours. This PR puts reported state and measured delivery in one live view, so an operator can see which sessions actually receive bytes.

A 48-hour soak test of this branch on a single-process deployment found 24 sessions with the ghost pattern, about 175 phantom session-hours in total. One had reported playback for 17.6 hours with 0 bytes measured.

The PR adds visibility only. It does not end sessions and does not fix the lifecycle bug in #666.

## Features to carry forward

**Reporting publisher**
- Each API process publishes reported playback state (position, paused, activity) under its measuring publisher ID plus `#reported`.
- Reported state never supplies byte counts or viewer addresses. Provenance is rebuilt from an allow-list of fields, and only viewer-egress measurements count as delivery evidence.
- `LocalHub` carries both publishers in single-process deployments. `MaxPublishers` goes from 256 to 512, because each API process now runs two publishers.
- The wire additions stay decodable by older publishers.

**Coverage and view health**
- Publishers declare their coverage and their reporting companion instead of the merge assuming it.
- A missing, stale or undeclared publisher makes the view `partial` or `stale` instead of reading as complete.
- An unhealthy view clears the idle classifications, so a paused or fully buffered session is not hidden while evidence is missing (this covers pruned measurements too).
- If publishers disagree about a session's identity, the view records both values and raises a visible admin warning. It does not pick one: a disagreement may be an abuse signal. Any resolution logic is deferred.

**Merged live endpoint**
- A session list that starts from telemetry and joins PostgreSQL display fields by session ID. The existing session list stays unchanged.
- Feature detection through the `stream_telemetry_live_sessions` capability. On v2 this belongs in v2 capability discovery.
- Server-side idle classification. The client renders it and does not re-derive it.
- Pruned unclaimed sessions stay hidden while the view is blind, and ended sessions stay out of live counts.

**Measurement fixes found along the way**
- Transfers are keyed per viewer with hashed IDs. Raw IPs and device IDs never appear in Redis field names.
- Per-subject overflow keeps role and class on fold rows, and class is carried on transfers.
- Tombstones count toward the Redis publisher field bound, and retired measurements are remembered.
- Bitrate probes get their own class and are excluded from media egress totals.
- Interrupted stream copies check for a cut between slices on both HTTP/1.1 and HTTP/2 and record the interruption.
- Hitting the transfer cap raises an advisory instead of switching off classification across the fleet. No single client can disable ghost detection.
- The parity report compares independent measured evidence only, keeps ended sessions out, and withholds agreement when the comparison is incomplete or self-derived.

**Admin UI**
- Delivery badges separate reported, measured and uncertain evidence, and show view health.
- The dashboard's Now Playing widget and the Activity page both get a reveal control for rows hidden as healthy idle.
- An empty list under a `starting`, `partial` or `stale` view keeps its health badge. Only a trustworthy empty result, or legacy mode with telemetry disabled, shows the normal idle state.
- Live-session queries poll every 30 seconds while the page is visible and focused. A separate 60-second dashboard timer skips a tick while the previous automatic refresh is still pending. The refresh interval is longer than the reconciler's tick, which avoids a refetch storm.

## Rebuild notes

- Build on the v2 admin session surface and v2 capability discovery, not v1.
- Keep the stricter identity-conflict handling now on `main`, which records a conflict when a value moves. Don't reintroduce this branch's older "started-at is never a conflict" behavior.
- Apple and Android had no consumers of the v1 live endpoint when this was checked. Recheck both against v2 before the rebuild lands.
- This branch stays available as the reference implementation.

PR #785, opened 2026-08-27
feat(taskmanager): add bounded task history retention cleanup

Quick104:
## Problem

Task execution history grows indefinitely, increasing database storage and slowing history queries. The existing catalog search event retention task also duplicated bounded cleanup behavior that belongs in a general retention mechanism.

## Approach

Add a scheduled task-history cleanup job that runs at startup and daily. It retains the newest 1,000 executions per task, removes older records beyond 30 days, processes at most 10,000 rows across 100 batches per run, and preserves the newest execution for every task, including unregistered tasks.

Cleanup uses PostgreSQL ranking, deterministic completion-time/ID ordering, and a session-level advisory lock so only one Silo node performs cleanup at a time. The previous catalog search event retention implementation and migration are removed, and latest-history queries now use ID ordering to break timestamp ties.

## Validation

Not run; no command results were provided.

Added unit and PostgreSQL integration coverage for retention limits, age pruning, tie-breaking, batching, idempotence, concurrency, advisory-lock skipping, invalid limits, and task progress/result reporting.

## Risks

Database cleanup is bounded per run and preserves each task's newest execution. The removed catalog search retention path changes cleanup ownership but does not remove pending or recovery-sensitive search events. No migration or security impact identified.

## AI Disclosure

- Tool(s): none
- Model(s): n/a
- Involvement: AI-assisted
- Adversarial review: Not performed; the supplied diff was summarized into change request content.

## Checklist

- [ ] I read and can explain the complete diff.
- [ ] This pull request addresses one concern.

### DIFF
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -2019,6 +2019,10 @@ func main() {
 		log.Fatalf("seed activitylog defaults: %v", err)
 	}
 
+	if err := taskmanager.SeedHistoryRetentionDefaults(ctx, settingsRepo); err != nil {
+		log.Fatalf("seed task history retention defaults: %v", err)
+	}
+
 	// Seed default page sections for home and existing libraries.
 	sectionRepo := sections.NewRepository(pool)
 	var folders []*models.MediaFolder
@@ -2172,7 +2176,7 @@ func main() {
 		}
 		taskMgr.Register(tasks.NewActivityLogCleanupTask(deps.DB, settingsRepo, activityPM))
 		taskMgr.Register(tasks.NewOperationalLogCleanupTask(deps.DB, settingsRepo, opsPM))
-		taskMgr.Register(tasks.NewTaskHistoryCleanupTask(historyRepo))
+		taskMgr.Register(tasks.NewTaskHistoryCleanupTask(historyRepo, settingsRepo))
 		var diagnosticsStore diagnostics.ObjectStore
 		if deps.S3Private != nil {
 			diagnosticsStore = diagnostics.NewS3ObjectStore(deps.S3Private)
--- a/cmd/silo/main.go
+++ b/cmd/silo/main.go
@@ -2172,6 +2172,7 @@ func main() {
 		}
 		taskMgr.Register(tasks.NewActivityLogCleanupTask(deps.DB, settingsRepo, activityPM))
 		taskMgr.Register(tasks.NewOperationalLogCleanupTask(deps.DB, settingsRepo, opsPM))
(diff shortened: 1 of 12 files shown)


Quick104:
Pushed 24bcabd02 addressing the code-review findings:

**Prune redesign (perf).** The old query window-ranked the entire `task_executions` table once per batch (up to 100 times per run). Prune now enumerates distinct task keys, resolves two boundary rows per key (newest and the keep-per-task rank) with index probes, and deletes via tuple comparisons against those bounds — every lookup and delete is a range scan of the existing `idx_task_executions_key_completed`. No migration. A dedicated test proves the tuple predicate selects the exact same doomed set as the original window ranking across keep/cutoff combinations, including duplicate timestamps. The unnecessary `ORDER BY id` inside the delete is gone, and `LimitReached` is now probed rather than inferred from loop exhaustion, so an exact batch-multiple backlog no longer reports a spurious leftover warning.

**Shared advisory-lock helper (reuse).** New `internal/database/pglock` carries the hijack-and-destroy-on-unlock-failure behavior so a session-level lock can never return to the pool; both this task and the previously duplicated `SearchIndexAdvisoryLock` in `internal/catalog` now use it (the catalog copy's latent leak on unlock failure is fixed by the migration).

**Settings-driven retention (consistency).** `taskmanager.history_retention_days` (30) and `taskmanager.history_keep_per_task` (1000) are seeded and admin-editable like the sibling cleanup tasks; out-of-range values fall back to defaults. Batch size / max batches stay compile-time constants (they bound per-run I/O, not policy).

**Smaller items.** Lock-contention skips are now logged (they still persist as completed with `skipped:true` in result_data); the mirror result struct is gone (`HistoryPruneResult` is JSON-tagged directly, matching `ExecutionResult`); the unreachable nil-progress guard and the `< 1` runtime validation of compile-time constants are removed; `TaskHistoryPruner` is unexported per the `catalogSearchEventPruner` convention; the repository tests now run against the migrated `SILO_TEST_DATABASE_URL` database instead of hand-copied DDL.

The startup trigger is kept: with the per-key design a no-op boot run costs two index probes per task key, so the rolling-restart concern no longer applies.

Verified locally: `go test` green for `internal/taskmanager/...`, `internal/database/...`, `internal/config`, `internal/catalog/...` (one pre-existing failure `TestEpisodeSearchPostgresAndDocumentSource` fails identically on the parent commit); `go build ./...`, `go vet`, `gofmt` clean; `golangci-lint --new-from-merge-base=origin/main`: 0 issues; `make verify-settings-bindings` passes.

PR #790, opened 2026-08-27
fix(playback): preserve video for Android EAC3 fallback

blurbery:
## What I changed

I added an Android-mobile fallback for multichannel E-AC-3 on Bluetooth outputs that advertise no E-AC-3 passthrough. Direct play remains the first choice. If Android rejects that route, the next HLS plan copies the original video bitstream and converts only E-AC-3 audio to AAC. Full video transcoding remains available only if the direct-stream route also fails.

## Why

I reproduced this with the exact same media file on Android and iPhone. The iPhone direct-played H.264 High Level 4.0 with E-AC-3/Atmos and reached first frame. The Android phone advertised ample H.264 hardware support, but returned `ERROR_CODE_DECODING_FAILED` for both original HTTP and a codec-copy HLS remux. The old final fallback then re-encoded H.264 to H.264 as well as converting the audio.

That video encode did not address the observed failure and added unnecessary quality loss and compute. The deployed correction now reaches first frame as `server_remux_hls` with one transformation, `audio_to_aac`; the H.264 video is copied byte-for-byte. In media-server terms this is Direct Stream/Remux, not video transcoding.

The rule is capability-scoped rather than model-scoped: Android platform, mobile form factor, Bluetooth sink, empty passthrough set, E-AC-3, and at least six channels. It therefore covers Android mobile models reporting the same route facts. The live validation device was one Pixel 7; the test fixture uses a non-Pixel Android model to verify that planning does not depend on manufacturer or model.

## Validation

Passed against current upstream `main`:

```text
go test ./internal/playback -run 'Test(AndroidMobileBluetoothEAC3Fallback|AFTKRTEAC3HLSCorrection|DeviceQuirkProtocol)' -count=1
ok github.com/Silo-Server/silo-server/internal/playback 1.436s
```

The production fork passed its complete Go, Web, and docs GitHub Actions gates. After deployment, the live plan selected `server_remux_hls`, copied H.264, converted only audio to AAC, reached first frame, and stopped cycling.

## Scope and risk

A working direct route is never displaced. Android TV, speaker output, stereo E-AC-3, Apple clients, passthrough-capable routes, and other codecs keep existing planning. No API, schema, migration, configuration, or Jellyfin compatibility changes are included.

Source implementation and production validation: [blurbery/silo-server#51](https://github.com/blurbery/silo-server/pull/51).

Related issue: N/A — narrow fix reproduced on a live deployment.

## AI Disclosure

- Model(s): gpt-5.6-sol
- Involvement: AI-assisted; I designed the capability boundary, reviewed the implementation, compared both clients on the same source, and validated the deployed route
- Adversarial review: Checked that direct play remains first, the fallback copies video, and the correction cannot activate for TV, speaker, stereo, passthrough-capable, Apple, or non-E-AC-3 routes. Focused positive and negative tests passed against current upstream main.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback for E-AC-3 audio on Android mobile devices connected via Bluetooth.
  * Automatically falls back to HLS with AAC audio conversion while preserving video when required.
  * Maintains expected audio behavior for speaker outputs and Android TV devices.

### DIFF
--- a/internal/playback/device_quirks_v3.go
+++ b/internal/playback/device_quirks_v3.go
@@ -3,9 +3,10 @@ package playback
 import "strings"
 
 const (
-	QuirkFireTVAFTKRTHigh10V3  = "android.fire_tv.aftkrt.h264_high10_l52_v1"
-	QuirkFireTVAFTKRTEAC3HLSV3 = "android.fire_tv.aftkrt.eac3_7_1_hls_audio_adapt_v1"
-	QuirkFireTVDV8HDR10PlusV3  = "android.fire_tv.dv8_hdr10plus_sei_v1"
+	QuirkFireTVAFTKRTHigh10V3         = "android.fire_tv.aftkrt.h264_high10_l52_v1"
+	QuirkFireTVAFTKRTEAC3HLSV3        = "android.fire_tv.aftkrt.eac3_7_1_hls_audio_adapt_v1"
+	QuirkAndroidMobileEAC3BluetoothV3 = "android.mobile.eac3_bluetooth_hls_audio_adapt_v1"
+	QuirkFireTVDV8HDR10PlusV3         = "android.fire_tv.dv8_hdr10plus_sei_v1"
 )
 
 func high10DecodeOverrideV3(source SourceDescriptorV3, request StartRequestV3) (*AppliedQuirkV3, bool) {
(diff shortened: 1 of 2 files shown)


blurbery:
Related loop-hardening work: [#789](https://github.com/Silo-Server/silo-server/pull/789) and [Silo-Server/silo-android#250](https://github.com/Silo-Server/silo-android/pull/250). Those prevent false route churn; this PR preserves copied video when Android reports a real multichannel E-AC-3 Bluetooth decoder failure.

PR #792, opened 2026-08-27
fix(playback): refresh incomplete skip markers

blurbery:
## Problem

Skip Intro could appear only intermittently, and Skip Credits could stay missing even when an online provider had useful data. The server stopped refreshing online markers too early when only one skip segment was complete. A provider result could also arrive before the player's realtime connection was ready, leaving the active session with stale ranges.

This also needed to support public read-only marker providers without exposing contribution controls or sending submission jobs to them.

## Approach

Playback-time refresh remains eligible until both intro and credits are complete from online, plugin, or manual sources. Scanner and S3 ranges remain eligible for an online upgrade. A bounded 30-minute server-side retry cache prevents partial or missing provider data from triggering a new lookup on every play.

After the first validated realtime hello on each connection, the server loads the current persisted markers and sends them only to that exact WebSocket registration. Empty rows do not emit an all-null clearing event. Per-file epochs order snapshots with concurrent marker updates, while WebSocket writes happen outside the state lock and outside the request path. Marker ranges are built once per update rather than once per session.

Replacement connections become control-ready only after their own hello. Registration, readiness, and teardown are serialized so an old connection cannot leave a replacement marked ready or clear a replacement that has already completed its hello.

Marker ranges remain tied to a file generation, but hash changes now preserve manually entered segments. The invalidation SQL is generated from one column table and one predicate, with a database-free test proving that every marker range and provenance column is covered.

Marker plugins can declare `supports_contribution=false`. Existing plugins retain contribution support by default, while fetch-only providers expose only the provider interface.

## Client coordination

- Apple follow-up: https://github.com/Silo-Server/silo-apple/pull/203
- Android phone and TV already decode the existing `markers_updated` event and apply intro, credits, recap, and preview ranges. This PR does not change that event's schema, and empty reconnect snapshots are now suppressed, so no Android change is required.
- jellycompat already serves persisted ranges through `/MediaSegments/{id}`. The native WebSocket reconciliation does not change that surface, so no jellycompat change is required.

## Performance

Playback does not wait for provider lookups. Online refresh remains background work, duplicate work is collapsed per file, and recent partial or empty online results are negatively cached in a bounded in-memory map.

Realtime reconciliation adds at most one bounded media-row read after the first hello on a connection. Admin marker saves and provider updates no longer wait for WebSocket I/O, and unrelated files no longer share one of 64 lock buckets.

## Related changes

- TheIntroDB alternate-ID fallback, caching, and Cloudflare backoff: https://github.com/Silo-Server/silo-plugin-markers-theintrodb/pull/5
- New independent IntroDB.app provider: https://github.com/blurbery/silo-plugin-markers-introdb-app
- Tested IntroDB.app v0.1.0 release: https://github.com/blurbery/silo-plugin-markers-introdb-app/releases/tag/v0.1.0
- IntroDB.app catalog entry: https://github.com/Silo-Server/silo-plugins/pull/7
- Apple prompt and reconciliation fix: https://github.com/Silo-Server/silo-apple/pull/203

## Validation

Passed on the focused PR branch:

- Targeted race suite for scanner invalidation, marker refresh/cooldown, realtime ordering, reconnect ownership, snapshot delivery, and provider registry behavior across `internal/scanner`, `internal/playback`, `internal/api/handlers`, and `internal/markers`
- `GOWORK=off go vet ./internal/scanner ./internal/playback ./internal/api/handlers ./internal/markers`
- `make verify-local-paths`
- `git diff --check`

A full non-race run of the four relevant packages passed for scanner, API handlers, and markers. The playback package reached only the repository's macOS/NVIDIA GPU-probe failures; the changed marker and realtime tests passed separately under the race detector. `GOWORK=off go build ./internal/... ./cmd/...` was not completed locally because this clean worktree does not contain the generated `web/dist` embed target. The complete Go and Web build remains covered by GitHub Actions.

The earlier implementation was also live-tested on a Silo server: playback started immediately, online markers completed in the background, and persisted intro and credit ranges were reused by later sessions. The review-fix commit itself has not been deployed.

## Risks

The online retry cache is node-local, so a multi-node deployment can make one provider attempt per node during the retry window. It is bounded to entries attempted in the last 30 minutes and requires no migration or shared cache dependency.

Live marker invalidation events may still intentionally carry null ranges when the server has removed a previously persisted marker. Only the initial reconnect snapshot suppresses an entirely empty row.

## AI Disclosure

- Model(s): GPT-5
- Involvement: AI-assisted
- Adversarial review: I reviewed the complete review-fix diff against the reported data-loss and concurrency sequences, traced every changed call site, and ran the targeted race suite. That pass caught an initially unbounded negative-cache map, which was replaced with expiry pruning. I also checked manual-marker retention, source priority, stale snapshot ordering, connection replacement, all-null payload behavior, Android handling, and jellycompat parity. The optional companion-review runtime was not installed locally, so no companion model result is claimed.

## Checklist

- [x] This pull request contains only the focused marker refresh and realtime reconciliation work.

## Summary by CodeRabbit

* **New Features**
  * WebSocket clients receive persisted playback marker snapshots when connecting.
  * Marker updates are delivered in the correct order, improving playback state consistency.
  * Plugins that don’t support marker contributions are restricted to read-only access.

* **Bug Fixes**
  * Online marker lookups now retry when only partial skip markers are available, with a cooldown to prevent repeated attempts.
  * Marker snapshots are sent only once per connection and not to replaced connections.
  * Markers are cleared when a file is replaced, while manual markers are preserved.
  * Improved error details in marker provider logs.

### DIFF
--- a/internal/api/handlers/playback_lazy_markers.go
+++ b/internal/api/handlers/playback_lazy_markers.go
@@ -12,7 +12,10 @@ import (
 	"github.com/Silo-Server/silo-server/internal/scanner"
 )
 
-const playbackLazyMarkerTimeout = 10 * time.Minute
+const (
+	playbackLazyMarkerTimeout       = 10 * time.Minute
+	playbackLazyOnlineRetryInterval = 30 * time.Minute
+)
 
 type PlaybackIntroEligibilityChecker interface {
 	IntroDetectionEligibleForPlayback(ctx context.Context, fileID int) (bool, error)
@@ -84,7 +87,10 @@ func (h *PlaybackHandler) maybeQueueLazyPlaybackMarkers(
 	hasOnline := h.hasOnlineMarkerProviders()
 	shouldRunLocal := markers.ShouldRunLocal(mode)
 	shouldRunOnline := (mode == markers.ModeOnline || mode == markers.ModeBoth) && hasOnline
-	if shouldRunOnline && hasCompletePlaybackSkipMarkers(file) {
+	if shouldRunOnline && hasCompleteOnlinePlaybackSkipMarkers(file) {
+		shouldRunOnline = false
+	}
+	if shouldRunOnline && h.hasRecentOnlineMarkerAttempt(file.ID, time.Now()) {
 		shouldRunOnline = false
 	}
 
@@ -172,6 +178,7 @@ func (h *PlaybackHandler) runLazyPlaybackMarkers(
 		"mode", mode)
 
 	if runOnline {
+		h.recordOnlineMarkerAttempt(file.ID, time.Now())
 		wrote, err := h.fetchOnlineMarkersForPlayback(ctx, file)
 		if err != nil {
 			slog.Warn("playback lazy markers: online fetch failed",
@@ -179,7 +186,7 @@ func (h *PlaybackHandler) runLazyPlaybackMarkers(
 				"file_id", file.ID,
 				"episode_id", file.EpisodeID,
 				"mode", mode,
-				"error", err.Error())
+				"error", err)
 		}
 		if wrote {
 			if refreshed := h.reloadPlaybackMarkerFile(ctx, file.ID); hasAnyMarker(refreshed) {
@@ -320,18 +327,66 @@ func (h *PlaybackHandler) notifyPlaybackMarkers(
 		"mode", mode)
 }
 
-// hasCompletePlaybackSkipMarkers reports whether the two playback skip
-// segments supplied by online marker providers are already populated. A
-// partial online result must remain eligible: TheIntroDB can gain a missing
-// intro or credits marker after the first playback, and the next fresh
-// playback session is the inexpensive opportunity to fill it for every user.
-// Provider-side caching prevents repeated external requests for a recent miss.
-func hasCompletePlaybackSkipMarkers(file *models.MediaFile) bool {
+// hasCompleteOnlinePlaybackSkipMarkers reports whether both playback skip
+// segments are populated by a source that online lookup cannot improve. Local
+// scanner and S3 ranges remain eligible because online providers outrank them.
+func hasCompleteOnlinePlaybackSkipMarkers(file *models.MediaFile) bool {
 	if file == nil {
 		return false
 	}
+	isDurableOnlineSource := func(segmentSource *string) bool {
+		source := segmentSource
+		if source == nil {
+			source = file.MarkersSource
+		}
+		if source == nil {
+			return false
+		}
+		switch *source {
+		case models.MarkerSourceOnline, models.MarkerSourcePlugin, models.MarkerSourceManual:
+			return true
+		default:
+			return false
+		}
+	}
 	return file.IntroStart != nil && file.IntroEnd != nil &&
-		file.CreditsStart != nil && file.CreditsEnd != nil
+		file.CreditsStart != nil && file.CreditsEnd != nil &&
+		isDurableOnlineSource(file.IntroMarkersSource) &&
+		isDurableOnlineSource(file.CreditsMarkersSource)
+}
+
+func (h *PlaybackHandler) hasRecentOnlineMarkerAttempt(fileID int, now time.Time) bool {
+	if h == nil || fileID <= 0 {
+		return false
+	}
+	h.markerLazyOnlineAttemptMu.Lock()
+	defer h.markerLazyOnlineAttemptMu.Unlock()
+	attemptedAt, ok := h.markerLazyOnlineAttempts[fileID]
+	if !ok {
+		return false
+	}
+	if now.Sub(attemptedAt) >= playbackLazyOnlineRetryInterval {
+		delete(h.markerLazyOnlineAttempts, fileID)
+		return false
+	}
+	return true
+}
+
+func (h *PlaybackHandler) recordOnlineMarkerAttempt(fileID int, attemptedAt time.Time) {
+	if h == nil || fileID <= 0 {
+		return
+	}
+	h.markerLazyOnlineAttemptMu.Lock()
+	defer h.markerLazyOnlineAttemptMu.Unlock()
+	if h.markerLazyOnlineAttempts == nil {
+		h.markerLazyOnlineAttempts = make(map[int]time.Time)
+	}
+	for cachedFileID, cachedAt := range h.markerLazyOnlineAttempts {
+		if attemptedAt.Sub(cachedAt) >= playbackLazyOnlineRetryInterval {
+			delete(h.markerLazyOnlineAttempts, cachedFileID)
+		}
+	}
+	h.markerLazyOnlineAttempts[fileID] = attemptedAt
 }
 
 // hasAnyMarker reports whether the file has at least one populated marker
--- a/internal/api/handlers/playback_lazy_markers.go
+++ b/internal/api/handlers/playback_lazy_markers.go
@@ -84,7 +84,7 @@ func (h *PlaybackHandler) maybeQueueLazyPlaybackMarkers(
 	hasOnline := h.hasOnlineMarkerProviders()
 	shouldRunLocal := markers.ShouldRunLocal(mode)
 	shouldRunOnline := (mode == markers.ModeOnline || mode == markers.ModeBoth) && hasOnline
-	if shouldRunOnline && hasOnlineSourcedMarkers(file) {
+	if shouldRunOnline && hasCompletePlaybackSkipMarkers(file) {
 		shouldRunOnline = false
 	}
 
@@ -179,7 +179,7 @@ func (h *PlaybackHandler) runLazyPlaybackMarkers(
 				"file_id", file.ID,
 				"episode_id", file.EpisodeID,
 				"mode", mode,
-				"error", err)
+				"error", err.Error())
 		}
 		if wrote {
 			if refreshed := h.reloadPlaybackMarkerFile(ctx, file.ID); hasAnyMarker(refreshed) {
@@ -320,31 +320,18 @@ func (h *PlaybackHandler) notifyPlaybackMarkers(
 		"mode", mode)
 }
 
-// hasOnlineSourcedMarkers reports whether the file already has at least one
-// marker written by a non-local source. We short-circuit lazy online fetch
-// in that case to avoid refetching every playback start — TheIntroDB often
-// returns only intro+credits and never recap/preview for a given episode, so
-// requiring all four kinds before skipping would loop forever on partial data.
-// Markers from the scanner/s3 path remain refetchable since online sources
-// outrank them.
-func hasOnlineSourcedMarkers(file *models.MediaFile) bool {
+// hasCompletePlaybackSkipMarkers reports whether the two playback skip
+// segments supplied by online marker providers are already populated. A
+// partial online result must remain eligible: TheIntroDB can gain a missing
+// intro or credits marker after the first playback, and the next fresh
+// playback session is the inexpensive opportunity to fill it for every user.
+// Provider-side caching prevents repeated external requests for a recent miss.
+func hasCompletePlaybackSkipMarkers(file *models.MediaFile) bool {
 	if file == nil {
 		return false
 	}
-	isOnlineSource := func(source *string) bool {
-		if source == nil {
-			return false
-		}
-		switch *source {
-		case models.MarkerSourceOnline, models.MarkerSourcePlugin, models.MarkerSourceManual:
-			return true
-		}
-		return false
-	}
-	return isOnlineSource(file.IntroMarkersSource) ||
-		isOnlineSource(file.CreditsMarkersSource) ||
-		isOnlineSource(file.RecapMarkersSource) ||
-		isOnlineSource(file.PreviewMarkersSource)
+	return file.IntroStart != nil && file.IntroEnd != nil &&
+		file.CreditsStart != nil && file.CreditsEnd != nil
 }
 
 // hasAnyMarker reports whether the file has at least one populated marker
--- a/internal/api/handlers/playback_lazy_markers.go
+++ b/internal/api/handlers/playback_lazy_markers.go
@@ -90,7 +90,7 @@ func (h *PlaybackHandler) maybeQueueLazyPlaybackMarkers(
 	if shouldRunOnline && hasCompleteOnlinePlaybackSkipMarkers(file) {
 		shouldRunOnline = false
 	}
-	if shouldRunOnline && h.hasRecentOnlineMarkerAttempt(file.ID, time.Now()) {
+	if shouldRunOnline && h.hasRecentOnlineMarkerAttempt(file, time.Now()) {
 		shouldRunOnline = false
 	}
 
@@ -178,7 +178,7 @@ func (h *PlaybackHandler) runLazyPlaybackMarkers(
 		"mode", mode)
 
 	if runOnline {
-		h.recordOnlineMarkerAttempt(file.ID, time.Now())
+		h.recordOnlineMarkerAttempt(file, time.Now())
 		wrote, err := h.fetchOnlineMarkersForPlayback(ctx, file)
 		if err != nil {
 			slog.Warn("playback lazy markers: online fetch failed",
@@ -355,38 +355,52 @@ func hasCompleteOnlinePlaybackSkipMarkers(file *models.MediaFile) bool {
 		isDurableOnlineSource(file.CreditsMarkersSource)
 }
 
-func (h *PlaybackHandler) hasRecentOnlineMarkerAttempt(fileID int, now time.Time) bool {
-	if h == nil || fileID <= 0 {
+type playbackLazyOnlineAttemptKey struct {
+	fileID   int
+	fileHash string
+}
+
+func playbackLazyOnlineAttemptKeyFor(file *models.MediaFile) (playbackLazyOnlineAttemptKey, bool) {
+	if file == nil || file.ID <= 0 {
+		return playbackLazyOnlineAttemptKey{}, false
+	}
+	return playbackLazyOnlineAttemptKey{fileID: file.ID, fileHash: strings.TrimSpace(file.FileHash)}, true
+}
+
+func (h *PlaybackHandler) hasRecentOnlineMarkerAttempt(file *models.MediaFile, now time.Time) bool {
+	key, ok := playbackLazyOnlineAttemptKeyFor(file)
+	if h == nil || !ok {
 		return false
 	}
 	h.markerLazyOnlineAttemptMu.Lock()
 	defer h.markerLazyOnlineAttemptMu.Unlock()
-	attemptedAt, ok := h.markerLazyOnlineAttempts[fileID]
+	attemptedAt, ok := h.markerLazyOnlineAttempts[key]
 	if !ok {
 		return false
 	}
 	if now.Sub(attemptedAt) >= playbackLazyOnlineRetryInterval {
-		delete(h.markerLazyOnlineAttempts, fileID)
+		delete(h.markerLazyOnlineAttempts, key)
 		return false
 	}
 	return true
 }
 
-func (h *PlaybackHandler) recordOnlineMarkerAttempt(fileID int, attemptedAt time.Time) {
-	if h == nil || fileID <= 0 {
+func (h *PlaybackHandler) recordOnlineMarkerAttempt(file *models.MediaFile, attemptedAt time.Time) {
+	key, ok := playbackLazyOnlineAttemptKeyFor(file)
+	if h == nil || !ok {
 		return
 	}
 	h.markerLazyOnlineAttemptMu.Lock()
 	defer h.markerLazyOnlineAttemptMu.Unlock()
 	if h.markerLazyOnlineAttempts == nil {
-		h.markerLazyOnlineAttempts = make(map[int]time.Time)
+		h.markerLazyOnlineAttempts = make(map[playbackLazyOnlineAttemptKey]time.Time)
 	}
-	for cachedFileID, cachedAt := range h.markerLazyOnlineAttempts {
+	for cachedKey, cachedAt := range h.markerLazyOnlineAttempts {
 		if attemptedAt.Sub(cachedAt) >= playbackLazyOnlineRetryInterval {
-			delete(h.markerLazyOnlineAttempts, cachedFileID)
+			delete(h.markerLazyOnlineAttempts, cachedKey)
 		}
 	}
-	h.markerLazyOnlineAttempts[fileID] = attemptedAt
+	h.markerLazyOnlineAttempts[key] = attemptedAt
 }
 
 // hasAnyMarker reports whether the file has at least one populated marker
--- a/internal/api/handlers/session_ws.go
+++ b/internal/api/handlers/session_ws.go
@@ -25,6 +25,14 @@ type sessionRealtimeConn struct {
 	writeMu sync.Mutex
 }
 
+type sessionRealtimeClient struct {
+	handler       *PlaybackHandler
+	connectionCtx context.Context
+	registration  *playback.RealtimeRegistration
+	sessionID     string
+	helloReceived bool
+}
+
 func (c *sessionRealtimeConn) WriteJSON(v any) error {
 	if c == nil || c.conn == nil {
 		return playback.ErrRealtimeConnectionNotFound
@@ -85,7 +93,7 @@ func (h *PlaybackHandler) HandleSessionWebSocket(w http.ResponseWriter, r *http.
 	}
 
 	realtimeConn := &sessionRealtimeConn{conn: conn}
-	registration := h.RealtimeHub.Register(sessionID, realtimeConn)
+	registration := h.registerSessionRealtimeConnection(sessionID, realtimeConn)
 	if registration == nil {
 		conn.Close()
 		slog.WarnContext(r.Context(), "failed to register realtime websocket", "component", "api", "session", sessionID, "playback_session_id", sessionID)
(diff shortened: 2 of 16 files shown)


internal/api/handlers/session_ws.go
```diff
@@ -137,6 +138,7 @@ func (h *PlaybackHandler) handleRealtimeClientMessage(sessionID string, data []b
 			h.syncSessionsNow(context.Background(), "realtime_hello")
 		}
 		h.touchSessionActivity(sessionID)
+		go h.sendCurrentMarkerSnapshot(sessionID)
```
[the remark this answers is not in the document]
blurbery: Fixed in c6aa4f49. Snapshot loading now runs only on the first disconnected-to-connected transition. The new regression test sends two valid hello messages and verifies exactly one loader call; the focused handler build and tests pass.

internal/api/handlers/session_ws_test.go
```diff
@@ -92,6 +105,98 @@ func TestHandleSessionWebSocket_RequiresHelloBeforeRealtimeReady(t *testing.T) {
...
+	select {
+	case <-resolver.calls:
+	case <-time.After(time.Second):
+		t.Fatal("initial hello did not load a marker snapshot")
+	}
+	select {
+	case <-resolver.calls:
+		t.Fatal("repeated hello loaded a second marker snapshot")
+	case <-time.After(100 * time.Millisecond):
+	}
```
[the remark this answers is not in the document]
blurbery: Fixed in 08cea7c0. The timing-based negative assertion is gone. Hello state is scoped to the WebSocket connection, and the test synchronously proves: first hello requests a snapshot, a repeat does not, and the first hello on a replacement connection does. Focused handler tests and build pass.

internal/api/handlers/session_ws.go
```diff
@@ -103,67 +104,77 @@ func (h *PlaybackHandler) HandleSessionWebSocket(w http.ResponseWriter, r *http.
...
 			break
 		}
-		if err := h.handleRealtimeClientMessage(sessionID, data); err != nil {
+		sendMarkerSnapshot, err := h.handleRealtimeClientMessage(sessionID, data, &helloReceived)
+		if err != nil {
 			slog.WarnContext(r.Context(), "invalid realtime client message", "component", "api", "session", sessionID, "playback_session_id", sessionID, "error", err)
+			continue
+		}
+		if sendMarkerSnapshot {
+			go h.sendCurrentMarkerSnapshot(sessionID)
```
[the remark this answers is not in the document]
blurbery: Fixed in 37676c6a with the explicit disconnect sequence covered in 4a77dd96. The snapshot now carries the exact RealtimeRegistration and connection context. Cleanup cancels the load, and the hub refuses delivery if that registration no longer owns the active WebSocket, so a delayed snapshot from connection A cannot reach replacement connection B. Cleanup also changes session state only when it actually unregisters the current connection. Focused playback, handler, WebSocket, formatting, and vet checks pass.

blurbery:
Fixed the outside-diff stale-work finding in 3b66ee27 and completed the replacement wake-up path in 58d82ce3. Snapshot lock acquisition is context-cancellable, each registration now closes an ownership signal as soon as it is replaced or unregistered, and the notifier verifies ownership both before waiting and after acquiring the per-file lock. Stale jobs wake immediately and exit before GetByID, so reconnect churn cannot queue serialized database reads. The channel-synchronized tests cover replacement and cancellation while queued; they pass 100 consecutive runs, and the focused race test, handler tests, vet, and diff check pass.

Quick104 (commented):
Code review of the full merge-base diff (10 commits, 14 files). I checked out the branch and ran `go vet` plus `go test -race` on `internal/playback`, `internal/markers`, and the marker/realtime tests in `internal/api/handlers` — all green. (`go build ./...` fails only on `web/embed.go: pattern all:dist` because there's no frontend build in my worktree; unrelated.)

15 inline comments below. The two I'd want resolved before merge:

1. **Marker invalidation is source-blind and irreversible** (`internal/scanner/file_repo.go`). Any OSHash change wipes manual markers too, and the only restore path is an S3 sidecar keyed by the *new* hash. `markers.lazy_playback` defaults to `false`, so there's no playback-time refill either. A routine remux changes the hash without changing the timeline.
2. **`hasCompletePlaybackSkipMarkers` dropped the source check** (`internal/api/handlers/playback_lazy_markers.go`). The deleted comment recorded the invariant — "Markers from the scanner/s3 path remain refetchable since online sources outrank them" — and the new predicate lets locally-detected markers permanently block the higher-priority online providers. It also still loops forever on files that can only ever have one of the two segments.

The rest are lower-severity correctness notes plus cleanup/altitude observations. Disclosure: this review was produced with AI assistance (Claude Code) and the findings above were verified against the branch by hand.

internal/scanner/file_repo.go
```diff
@@ -914,6 +920,36 @@ func (r *FileRepository) Upsert(ctx context.Context, mf models.MediaFile) (*mode
 		file_size = EXCLUDED.file_size,
 		file_modified_at = EXCLUDED.file_modified_at,
 		file_hash = EXCLUDED.file_hash,
+		intro_start = CASE WHEN media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash THEN NULL ELSE media_files.intro_start END,
```
Quick104: **Data loss: a hash change wipes manual markers with no restore path.**

This clears every marker range and provenance column on any OSHash change, regardless of `markers_source`. That includes `manual` — the highest-priority source, the one `UpsertMarkers` goes out of its way to protect from provider overwrites.

Concrete path:

1. An admin hand-enters intro/credits on an episode.
2. The user remuxes h264→h265, or re-downloads the same episode. `ComputeOSHash` (`scanner.go:3887`) is size + head/tail bytes, so it changes even though the timeline usually doesn't.
3. The next scan's `Upsert` nulls all 28 marker columns.
4. `fetchMarkers` only restores markers when an S3 object exists at `markers/<newhash>.json` — for manual/plugin/online/locally-detected markers there is no sidecar, so nothing comes back.
5. `markers.lazy_playback` defaults to `"false"` (`internal/config/admin_settings.go:72`), so the playback-time refill the PR description relies on doesn't run either.

The manual work is gone silently and irreversibly. Worth gating the invalidation on source (leave `manual` alone), or at least requiring corroborating evidence beyond the hash (e.g. a `file_size` change too).

internal/api/handlers/playback_lazy_markers.go
```diff
@@ -320,31 +320,18 @@ func (h *PlaybackHandler) notifyPlaybackMarkers(
...
-// Markers from the scanner/s3 path remain refetchable since online sources
-// outrank them.
-func hasOnlineSourcedMarkers(file *models.MediaFile) bool {
+// hasCompletePlaybackSkipMarkers reports whether the two playback skip
+// segments supplied by online marker providers are already populated. A
+// partial online result must remain eligible: TheIntroDB can gain a missing
+// intro or credits marker after the first playback, and the next fresh
+// playback session is the inexpensive opportunity to fill it for every user.
+// Provider-side caching prevents repeated external requests for a recent miss.
+func hasCompletePlaybackSkipMarkers(file *models.MediaFile) bool {
```
Quick104: **The gate lost its source check, so local markers now permanently block online providers.**

The old `hasOnlineSourcedMarkers` deliberately only counted `online`/`plugin`/`manual` sources, and its comment recorded why:

> Markers from the scanner/s3 path remain refetchable since online sources outrank them.

The replacement checks only that the pointers are non-nil. So a file whose intro+credits came from the local chromaprint/chapter analyzer or an S3 sidecar (`markers_source` `scanner`/`s3`) now sets `shouldRunOnline = false` forever at line 87 — a higher-priority TheIntroDB / IntroDB.app result is never fetched, and the better ranges never replace the locally guessed ones.

Keeping the source check *and* adding the per-segment completeness check would give you the PR's intended behavior without reversing this invariant.

internal/api/handlers/playback_lazy_markers.go
```diff
@@ -84,7 +84,7 @@ func (h *PlaybackHandler) maybeQueueLazyPlaybackMarkers(
 	hasOnline := h.hasOnlineMarkerProviders()
 	shouldRunLocal := markers.ShouldRunLocal(mode)
 	shouldRunOnline := (mode == markers.ModeOnline || mode == markers.ModeBoth) && hasOnline
-	if shouldRunOnline && hasOnlineSourcedMarkers(file) {
+	if shouldRunOnline && hasCompletePlaybackSkipMarkers(file) {
```
Quick104: **Files that can never be "complete" now re-fetch on every single playback start.**

The removed comment named exactly this failure mode:

> requiring all four kinds before skipping would loop forever on partial data

Going from four kinds to two narrows the problem but doesn't close it. A movie in a library with `markers.mode=online` and `lazy_playback=true` where TheIntroDB has credits but no intro will, on *every* play, re-resolve external IDs, run `FetchMerged` concurrently across every enabled provider, and attempt an `UpsertMarkers` that writes nothing.

The PR description's mitigation — "Provider-side caching prevents repeated external requests for a recent miss" — delegates this to third-party plugin implementations, which the server can't enforce. A server-side negative cache (a `markers_last_online_attempt_at` column, or a TTL entry keyed by file ID next to `MarkerLazyInFlight`) would make the bound real rather than advisory.

internal/playback/marker_update_notifier.go
```diff
@@ -32,45 +75,128 @@ func (n *MarkerUpdateNotifier) MarkersUpdated(ctx context.Context, file *models.
...
-			return nil
-		}
-		return &TimeRangePayload{Start: *start, End: *end}
-	}
-	intro := rangePayload(file.IntroStart, file.IntroEnd)
-	credits := rangePayload(file.CreditsStart, file.CreditsEnd)
-	recap := rangePayload(file.RecapStart, file.RecapEnd)
-	preview := rangePayload(file.PreviewStart, file.PreviewEnd)
+	lock := n.fileLock(file.ID)
+	lock.lock()
```
Quick104: **This holds an uninterruptible shared lock across websocket writes, on HTTP request goroutines.**

`MarkersUpdated` had no lock before. It now takes a bucket lock (`fileID % 64`, shared library-wide) and holds it across one `conn.WriteJSON` per matching session, each bounded only by the 5s `wsWriteTimeout`.

Both production callers are synchronous on a request goroutine:

- `internal/api/handlers/markers.go:587` — manual marker save
- `internal/api/handlers/admin_intro.go:181` — in a loop over every file of an episode

So one stalled client socket blocks the admin's save for the full write deadline; with three sessions on that file, ~15s. And because the bucket is `fileID % 64`, unrelated files collide (1 and 65), so a reconnect snapshot for a completely different title blows its 3s context inside `lockContext` and is silently dropped.

Building the event under the lock and doing the sends outside it would preserve the ordering guarantee you're after without holding the lock across network I/O.

internal/api/handlers/session_ws.go
```diff
@@ -207,8 +217,42 @@ func (h *PlaybackHandler) handleRealtimeClientMessage(sessionID string, data []b
...
+	if !ok {
+		return
+	}
+	session, err := h.sessionMgr.GetSession(sessionID)
+	if err != nil || session == nil || session.MediaFileID <= 0 {
+		return
+	}
+	ctx, cancel := context.WithTimeout(connectionCtx, 3*time.Second)
+	defer cancel()
+	if err := notifier.SendSessionSnapshotFromLoader(ctx, registration, session.MediaFileID, h.fileResolver); err != nil {
```
Quick104: **The snapshot is sent unconditionally, including when the file has no markers at all.**

Every hello costs a `media_files` row read and emits a `markers_updated` whose `intro`/`credits`/`recap`/`preview` are all JSON `null`.

That's not inert on the client. `web/src/player/utils/watchPageMarkers.ts` deliberately distinguishes `undefined` ("leave alone") from `null` ("clear") — `patchVersionMarkers` assigns `nextIntro = intro` when `intro` is `null`. Any client holding ranges from another source gets them cleared by the snapshot.

Guarding the send with the `hasAnyMarker` predicate that already exists in `playback_lazy_markers.go` removes both the wasted read and the clearing event.

internal/api/handlers/session_ws.go
```diff
@@ -91,79 +92,88 @@ func (h *PlaybackHandler) HandleSessionWebSocket(w http.ResponseWriter, r *http.
 		return
 	}
 
+	configureWebSocket(conn)
+	ctx, cancelRead := context.WithCancel(r.Context())
 	defer func() {
-		if h.setRealtimeConnectionState(sessionID, false) {
+		cancelRead()
+		if h.RealtimeHub.Unregister(registration) && h.setRealtimeConnectionState(sessionID, false) {
```
Quick104: **Reconnect can leave `HasRealtimeConnection` true for a socket that hasn't sent hello.**

The `&&` short-circuit means `setRealtimeConnectionState(sessionID, false)` never runs when `Unregister` returns false.

Sequence:

1. Client A helloes — `HasRealtimeConnection = true`.
2. A's TCP connection drops.
3. Client B reconnects and calls `RealtimeHub.Register` before A's handler goroutine reaches its defer. `Register` swaps `lane.conn` to B immediately.
4. A's defer gets `Unregister(regA) == false` (generation mismatch), so the flag is never cleared.

In the window before B's hello, the session reads as control-ready and any dispatched command is written to B's socket — which is exactly what the "control-ready only after a validated hello" invariant exists to prevent. `TestHandleSessionWebSocket_RequiresHelloBeforeRealtimeReady` only covers a fresh connection, so this isn't caught.

I can see why you gated it (it does fix the opposite race, where A's teardown clobbers B's flag). Clearing the flag only when the *hub has no live registration for the session* would handle both directions.

internal/scanner/file_repo_marker_replacement_db_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"time"
+
+	"github.com/Silo-Server/silo-server/internal/models"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+func TestUpsertInvalidatesMarkersOnlyForKnownFileReplacement(t *testing.T) {
+	dsn := os.Getenv("SILO_TEST_DATABASE_URL")
+	if dsn == "" {
+		t.Skip("SILO_TEST_DATABASE_URL is not set")
```
Quick104, line 17: **The riskiest change in the PR has no CI coverage.**

`SILO_TEST_DATABASE_URL` doesn't appear anywhere in `.github/` or the `Makefile`, so this test is skipped on every CI run. The env-gated pattern is established in this repo, so that's not the issue — the issue is that the thing being gated is destructive and irreversible.

A future edit that drops the `EXCLUDED.file_hash IS NOT NULL` guard from one column, or adds a new `*_markers_*` column without a matching clause, passes `make test-go` and CI green. The regression then surfaces as silent marker deletion on a real library.

If the SQL can't be exercised in CI, it's worth extracting the condition into a Go-level helper that builds the SET clause and unit-testing *that*, so at least the "every marker column is covered" property is checked without a database.

internal/api/handlers/session_ws.go
```diff
@@ -207,8 +217,42 @@ func (h *PlaybackHandler) handleRealtimeClientMessage(sessionID string, data []b
...
+
+func (h *PlaybackHandler) sendCurrentMarkerSnapshot(
+	connectionCtx context.Context,
+	registration *playback.RealtimeRegistration,
+	sessionID string,
+) {
+	if h == nil || h.fileResolver == nil || h.MarkerUpdateNotifier == nil {
+		return
+	}
+	notifier, ok := h.MarkerUpdateNotifier.(playbackMarkerSnapshotNotifier)
```
Quick104, line 294: **Three different failures here are indistinguishable from success.**

- This type assertion returns with no log at all. If `MarkerUpdateNotifier` is ever wired as a wrapper rather than the concrete `*playback.MarkerUpdateNotifier` — a metrics adapter, a nil-safe shim — the whole reconnect-snapshot feature goes dark silently.
- `SendSessionSnapshotFromLoader` returns `nil` identically for "registration replaced", "lock wait aborted", and "loader returned a nil file", so the `err != nil` branch below never fires for any of them.

A snapshot dropped because the file lock was contended looks exactly like one that was never needed. There's no signal an operator could use to notice clients are missing markers. A debug log on the assertion failure, and a sentinel (or a `(sent bool, err error)` return) for the skip cases, would make this diagnosable.

internal/markers/types.go
```diff
@@ -467,7 +467,7 @@ func (r *Registry) logProviderError(providerID string, req Request, err error) {
 		"provider", providerID,
 		"kind", req.Kind,
 		"external_ids", sanitizeExternalIDs(req.ExternalIDs),
-		"error", err,
+		"error", err.Error(),
```
Quick104: **`err.Error()` gives up nil-safety, and the PR is inconsistent about it.**

Both current call sites are inside `if err != nil`, so this is safe today. But `"error", err` can't panic and `"error", err.Error()` can, and `logProviderError` is reached from a goroutine spawned by `FetchMerged` — a nil there takes down the process, not just the request.

If the motivation is that a struct error marshals to `{}` under the JSON handler, `slog.String("error", ...)` guarded on nil, or a small `errAttr(err)` helper, gets the same output without the footgun.

Worth noting the same PR leaves `"error", err` at `marker_update_notifier.go:190` and `:199` while changing it here and at `playback_lazy_markers.go:182` — whichever form you land on, it'd be good to apply it consistently.

internal/scanner/file_repo.go
```diff
@@ -914,6 +920,36 @@ func (r *FileRepository) Upsert(ctx context.Context, mf models.MediaFile) (*mode
...
+		recap_markers_algorithm = CASE WHEN media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash THEN NULL ELSE media_files.recap_markers_algorithm END,
+		recap_markers_detected_at = CASE WHEN media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash THEN NULL ELSE media_files.recap_markers_detected_at END,
+		preview_markers_source = CASE WHEN media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash THEN NULL ELSE media_files.preview_markers_source END,
+		preview_markers_provider = CASE WHEN media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash THEN NULL ELSE media_files.preview_markers_provider END,
+		preview_markers_confidence = CASE WHEN media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash THEN NULL ELSE media_files.preview_markers_confidence END,
```
Quick104: **The same three-clause predicate is written out verbatim 28 times.**

`media_files.file_hash IS NOT NULL AND EXCLUDED.file_hash IS NOT NULL AND media_files.file_hash IS DISTINCT FROM EXCLUDED.file_hash` appears once per marker column.

Any change to the rule — say "also require a `file_size` change" — has to be applied identically 28 times, and missing one leaves a single column stale forever with no test to catch it (see the CI-coverage comment on the test file).

Two ways out: build the SET fragment in Go by ranging over a `markerInvalidationColumns` slice, or drop the invalidation out of the upsert entirely into a separate guarded statement:

```sql
UPDATE media_files SET intro_start = NULL, ... 
WHERE file_path = $1 AND file_hash IS NOT NULL AND $2 IS NOT NULL AND file_hash IS DISTINCT FROM $2
```

Either expresses the rule once and keeps the already-200-line upsert readable.

Separately: the explanatory comment for this block sits ~85 lines further up, in the middle of the JSON-marshaling section. Moving it next to the SQL it describes would help.

internal/playback/marker_update_notifier.go
```diff
@@ -32,45 +75,128 @@ func (n *MarkerUpdateNotifier) MarkersUpdated(ctx context.Context, file *models.
...
-			)
-			continue
+		n.sendSessionSnapshotLocked(ctx, session.ID, file)
+	}
+}
+
+// SendSessionSnapshot sends the current persisted marker ranges to one live
+// playback session. Calling it after the client's hello closes the race where
+// lazy marker discovery finishes before the websocket is control-ready.
+func (n *MarkerUpdateNotifier) SendSessionSnapshot(ctx context.Context, sessionID string, file *models.MediaFile) {
```
Quick104: **`SendSessionSnapshot` has no production caller.**

The handler uses `SendSessionSnapshotFromLoader`; the only reference to this method is `TestMarkerUpdateNotifierSendsSnapshotToOnlyRequestedSession`. Its doc comment ("Calling it after the client's hello closes the race where lazy marker discovery finishes before the websocket is control-ready") describes a call site that doesn't exist.

That's an active trap: it reads like the intended entry point, but it skips the registration and staleness checks that `SendSessionSnapshotFromLoader` adds. Worth deleting along with its test.

internal/api/handlers/session_ws.go
```diff
@@ -91,79 +92,88 @@ func (h *PlaybackHandler) HandleSessionWebSocket(w http.ResponseWriter, r *http.
...
+			continue
+		}
+		if sendMarkerSnapshot {
+			go h.sendCurrentMarkerSnapshot(ctx, registration, sessionID)
 		}
 	}
 }
 
-func (h *PlaybackHandler) handleRealtimeClientMessage(sessionID string, data []byte) error {
+func (h *PlaybackHandler) handleRealtimeClientMessage(sessionID string, data []byte, helloReceived *bool) (bool, error) {
```
Quick104: **Connection-scoped state threaded through a per-message parser as an in-out param.**

`handleRealtimeClientMessage` now carries a mutable `helloReceived *bool` and returns a `bool` meaning "caller should send a snapshot", with a `helloReceived == nil` branch at line 142 that exists only so the older tests can pass `nil`.

The state is now spread across three places: the read loop's local, this parameter, and the return value. The next per-connection flag (subtitle snapshot, chapter-thumb snapshot) repeats the pattern and the signature grows again.

A small per-connection struct holding `registration`, `ctx`, and `helloReceived`, with the snapshot dispatch as a method on it, keeps the state with the connection it belongs to and drops both the out-param and the nil branch.

internal/playback/realtime_hub.go
```diff
@@ -142,3 +171,41 @@ func (h *RealtimeHub) Send(sessionID string, message any) error {
 	lane.mu.Unlock()
 	return err
 }
+
+// SendRegistered writes only when reg still owns the active connection. It is
+// used for connection-specific work that must not spill into a replacement
+// websocket after a reconnect.
+func (h *RealtimeHub) SendRegistered(reg *RealtimeRegistration, message any) error {
```
Quick104: **`SendRegistered` validates the registration twice.**

`HasRegistration` does a map read under `h.mu` plus a `lane.mu` acquire, then this function re-acquires `lane.mu` and repeats the identical `closed` / `conn == nil` / `generation` check.

The first check can't prevent anything — the registration can be replaced between the two lock windows, and the second check is what actually decides. So every realtime send pays two map lookups and two acquisitions of the hot per-session lane mutex for no added safety.

The pre-refactor version had this right: do the lookup once and validate inside the single `lane.mu` critical section. `HasRegistration` can stay for the `SendSessionSnapshotFromLoader` pre-checks without `SendRegistered` routing through it.

internal/playback/marker_update_notifier.go
```diff
@@ -32,45 +75,128 @@ func (n *MarkerUpdateNotifier) MarkersUpdated(ctx context.Context, file *models.
...
+	})
+}
+
+func (n *MarkerUpdateNotifier) sendSessionSnapshotWithLocked(
+	ctx context.Context,
+	sessionID string,
+	file *models.MediaFile,
+	send func(any) error,
+) {
+	rangePayload := func(start, end *float64) *TimeRangePayload {
```
Quick104: **The four payloads are now rebuilt per session instead of once.**

The old `MarkersUpdated` computed `intro`/`credits`/`recap`/`preview` once before the session loop. Routing through `sendSessionSnapshotWithLocked` allocates the `rangePayload` closure and all four `TimeRangePayload` values once per session.

Minor in isolation, but it's 5N allocations instead of 5 on a popular file, and it happens while the bucket lock is held — which compounds the lock-hold-time concern above. Hoisting the payloads back out, or passing a prebuilt struct into the send helper, keeps the registration-aware send without regressing this.

internal/api/handlers/session_ws.go
```diff
@@ -91,79 +92,88 @@ func (h *PlaybackHandler) HandleSessionWebSocket(w http.ResponseWriter, r *http.
...
 			break
 		}
-		if err := h.handleRealtimeClientMessage(sessionID, data); err != nil {
+		sendMarkerSnapshot, err := h.handleRealtimeClientMessage(sessionID, data, &helloReceived)
+		if err != nil {
 			slog.WarnContext(r.Context(), "invalid realtime client message", "component", "api", "session", sessionID, "playback_session_id", sessionID, "error", err)
+			continue
+		}
+		if sendMarkerSnapshot {
+			go h.sendCurrentMarkerSnapshot(ctx, registration, sessionID)
```
Quick104: **Client coordination checklist for a client-visible realtime change.**

This adds a new client-visible behavior: an unconditional `markers_updated` after every hello. `CLAUDE.md` asks that a client-visible playback/session change be handled or explicitly ruled out on each of:

> Follow-up work is done or filed for both `silo-apple` and `silo-android` — prefer coordinated multi-repo changes over leaving a platform behind.
>
> jellycompat parity was considered (does the Jellyfin surface need the same behavior?).

The Related-changes list names an Apple *prompt* fix and no Android work, and jellycompat's MediaSegments surface isn't mentioned. Given the null-vs-undefined trap the web client documents in `watchPageMarkers.ts`, an Apple or Android player that treats the snapshot payload as authoritative will null out ranges it already had — so it'd be good to file the follow-ups even if the answer for jellycompat is "not applicable".

blurbery:
Thanks for the detailed review. I addressed all 15 points from the review, plus the later CodeRabbit cooldown/queued-snapshot edge cases, in `521f14b4`, `e83031f3`, and `b78dd987`.

The two blockers are fixed:

- Hash changes now preserve manual segments and their provenance. The invalidation assignments come from one column table and one shared predicate, with a database-free regression that checks every marker column is covered.
- The completion gate is source-aware again: scanner/S3 ranges remain eligible for online upgrades, while complete online/plugin/manual intro and credits stop further lookup. Partial or missing online results use a bounded 30-minute negative cache keyed by file ID and file hash, so replacement bytes can retry immediately.

The realtime path now uses per-file epochs and performs WebSocket writes outside the state lock/request path. Empty reconnect rows are skipped, payload ranges are built once, snapshot delivery is bound to the exact registration, and queued lock waits stop on context cancellation or registration replacement. Replacement readiness is serialized around registration/hello/teardown, snapshot skips are observable, the parser state lives on a connection struct, and the dead/double-validation helpers are gone. Error logging is nil-safe and consistent.

For client coordination: the Apple follow-up is Silo-Server/silo-apple#203; Android phone and TV already decode and apply the unchanged `markers_updated` event; jellycompat already reads the same persisted ranges through `/MediaSegments/{id}`, so no compatibility change is needed.

Validation on the final head:

- focused marker/realtime race tests: pass
- focused `go vet`: pass
- GitHub Actions Go, Web, and Docs hygiene: pass
- CodeRabbit: pass, with the later minor and nitpick addressed

The PR body has been updated with the implementation details, validation limits, client coordination, and AI disclosure.

PR #795, opened 2026-08-27
feat(admin): customizable widget dashboard with insights, metrics sampler, and server-side layouts

Quick104:
## Problem

The admin dashboard was a fixed page of stat cards and lists with no insight into what the
server is actually doing: no playback or bandwidth trends, no node/scanner/error visibility,
and a Trakt card that took a third of the screen to show four counters. Every admin got the
same layout regardless of what they cared about.

## What this does

Rebuilds `/admin` as a customizable widget dashboard and adds the backend needed for real
insights. Two commits: the widget framework, then the insights phase.

**Widget framework (web):** registry-driven 12-column × 100px-row grid with a customize mode —
drag to reorder, corner-drag to resize on both axes (keyboard accessible with aria-live
announcements), add/remove via a sheet, per-widget time ranges (1h/24h/7d/30d) usable outside
customize mode. Layouts persist server-side per admin (`admin_dashboard_layouts`, opaque jsonb,
`GET/PUT/DELETE /admin/dashboard/layout`) with localStorage as instant-paint fallback and
one-time migration of existing local layouts.

**Insights backend:** a new `internal/dashmetrics` sampler records minute-resolution concurrent
streams (by play method) and egress (node egress plus per-process viewer egress from stream
telemetry, `RoleViewerEgress` only to avoid relay double-counting); replica-safe via
`INSERT … ON CONFLICT DO NOTHING`, 31-day in-band retention, degrades gracefully without
Redis/telemetry/nodes. Three aggregate endpoints follow the `AdminStatsProvider`
TTL-cache + event-bus pattern: `/admin/stats/timeseries` (adaptive peak-preserving bucketing,
60s→2h resolution by window), `/admin/stats/playback-activity` (playback history UNION live
sessions, hourly/daily buckets, reliability counters), `/admin/stats/top-activity` (top titles
with episode→series rollup, most active profiles). `/admin/server/status` gains a `health`
object (Postgres/Redis pings, 24h error/warning counts); `/admin/logs/app` accepts a
comma-separated `level` list. Two supporting index migrations.

**Widgets:** 24 total — stat tiles (streams, egress now, transcode share, profiles active,
movies, shows, users, storage), health strip, playback activity chart, concurrent
streams/egress trends, reliability, now playing, transcode nodes, scanner, libraries, users
(now sorted by last activity), top titles/profiles, Trakt sync (demoted from a 4-column card
to a one-line strip), scan activity, recent errors, recent activity. Charts are hand-rolled
inline SVG on re-stepped `--chart-*` tokens (the previous unused values failed CVD/contrast
validation; new values pass all checks per theme).

## Why this approach

- Layout blob is opaque server-side: the client owns the schema, so widget/range additions
  need no API changes. v1 pre-lock shapes designed additively; admin-web-only surface, so no
  silo-apple/silo-android coordination and no jellycompat impact.
- The sampler exists because egress/concurrency history cannot be reconstructed retroactively;
  everything reconstructable (playback activity, top lists) is computed from existing tables.
- Aggregation is peak-preserving (MAX per display bucket) — these charts exist to show spikes.
- No chart library dependency; the theme's chart tokens finally get a consumer.

Design reference: HTML mockups iterated with the maintainer (kept out of the repo per docs
hygiene rules; the PR description is the record).

## Validation

- Full CONTRIBUTING gate run on the final diff (build, gofmt, vet, golangci-lint
  new-from-merge-base, test-go, web lint/format/build, test-web, contract/fixture/local-path
  verifiers).
- Live end-to-end on a fresh local install: migrations, sampler rows (shared + proc, baseline
  behavior on first tick), all widgets with empty states, layout persistence round-trip
  (PUT/adopt/DELETE), corner-resize both axes with clamping, range switching with adaptive
  resolutions (60s/300s/1800s/7200s) verified against live Postgres including `date_bin`.
- Deployed and exercised on the shared dev instance with real catalog data (~152k movies).
- Adversarially reviewed (see disclosure); all high/medium findings fixed, including a
  right-click resize wedge, keyboard-inert handles (WCAG 2.1.1), a tablet layout regression,
  a series-link dead end, and a sampler clock-skew edge.

Screenshots: running live on the shared dev deployment; screenshots/recording can be added on
request.

## Risks / follow-ups

- Sampled charts only have history from first deploy of the sampler; week/month views fill in
  over time ("Collecting data since…" states cover this).
- Time-to-first-frame and failed-start metrics are documented as future work — no data source
  exists yet (needs playback start events).
- Cross-tab layout sync is last-write-wins; `updated_at` is returned so CAS can be added later.
- Server-side layout persistence supersedes localStorage transparently; no client breakage.

Related issue: N/A — maintainer-directed feature developed interactively; no prior issue
existed. Happy to file a tracking issue for the follow-ups above.

## AI Disclosure

- Tool(s): Claude Code
- Model(s): claude-fable-5 (orchestration, design, architecture, final review),
  claude-opus-5 (implementation agents), claude-sonnet-5 (verification/test runs, recon)
- Adversarial review: two independent high-effort review passes per phase (correctness/
  regressions lens and quality/accessibility/conventions lens) over the complete diff, plus a
  final maintainer-side review of the framework, sampler, SQL, and grid interaction code.
  13 findings total; all high/medium findings fixed and re-verified (details in commit
  history / this description), low-severity items either fixed or documented above.

## Summary by CodeRabbit

* **New Features**
  * Added a customizable admin dashboard with draggable, resizable, removable, and addable widgets.
  * Added per-account dashboard layout saving, resetting, and synchronization.
  * Added playback, top titles and profiles, stream, egress, server health, resource, scan, user, and activity insights.
  * Added interactive charts, time-range controls, loading states, and empty-state messaging.
  * Added service health checks, latency reporting, cached statistics, and comma-separated log-level filtering.
* **Documentation**
  * Documented dashboard layouts, analytics, health reporting, and log filtering.


Quick104:
All review feedback is addressed in 873bec68d. Finding-by-finding:

**Codex**
- *Capability discovery (P1)* — added `GET /admin/dashboard/capabilities` following the `/admin/sessions/capabilities` convention, with handler test and `docs/admin-api.md` section.
- *Index migration blocks writes (P1)* — the migration now uses `-- +goose NO TRANSACTION` + `CREATE INDEX CONCURRENTLY` / `DROP INDEX CONCURRENTLY`, including the invalid-leftover-index cleanup block, matching `20260819060028_add_metadata_image_cache_target_index.sql`.
- *Leading partial playback bucket (P2)* — `buildPlaybackActivityColumns` now snaps the window start to the bucket grid the server truncates on, so the bucket containing `now - hours` is included. Note this makes the visible column count `⌈window/step⌉ + 1` when now is mid-bucket (a 24 h hourly chart draws 25 columns): that is the direct consequence of no longer dropping in-window sessions.
- *Watch time vs media duration (P2)* — `total_seconds` in `/admin/stats/top-activity` is now `SUM(playback_history_admin.watched_seconds)` over the window, joined onto the plays ranking; `user_watch_history.duration_seconds` (media runtime) is no longer summed. Plays still count history rows, so marked-watched shows up as a play with 0 watched seconds. Docs updated.
- *Unscoped localStorage layout key (P2)* — the cache key now includes the account id (`silo.admin-dashboard-layout.v1.u<id>`); the unscoped key is deleted on sight and never adopted or uploaded, so one admin's cache can no longer become another account's server layout. This key only ever shipped inside this PR, so there is nothing durable to migrate.
- *In-flight save vs reset (P2)* — both mutations now share a TanStack Query mutation scope, so save/save and save/reset run serially in start order; a PUT already in flight completes before the reset DELETE starts. Tests verified to fail without the scope.

**CodeRabbit (actionable)**
- Sampler gate includes empty `SILO_MODE`, matching `isPrimaryNode`.
- Migration: same concurrent-index fix as above.
- `LineChart` (and `StackedColumnChart`, same pattern) hit layers now show a `focus-visible` ring.
- `RANGE_HOURS.month` is 720, consistent with the 30-day label and `RANGE_DAYS.month`; the sampler's 31-day retention stays deliberately one day wider than anything charted.
- Health strip counts only enabled nodes in the denominator, fallback text, and warn tone; test added for a disabled node.
- User rows are keyboard-operable (`tabIndex`, Enter/Space, `role="link"`, aria-label, focus ring), following the existing clickable-row convention in `AdminDiagnostics`.
- Overlapping saves: same shared-scope serialization as above.

**CodeRabbit (nitpicks)** — all taken: UTC location assertion in the timeseries test; deterministic tiebreakers on both leaderboard queries; the profiles query ranks on `(user_id, profile_id)` first and resolves the username/profile name once per surviving row (a renamed profile no longer splits into two rows); `DEFAULT_LAYOUT` is derived from the widget registry; the grid reads `columnGap` back from the stylesheet with the constant as fallback.

Validation on the final diff: `go build`, `gofmt`, `go vet`, `golangci-lint --new-from-merge-base` (0 issues), full `make test-go`, `make test-web` (2842 tests), web lint/format/tsc/build, `make verify-settings-bindings-all`, `make verify-playback-fixtures`, `make verify-local-paths` — all clean.

Quick104:
Second-round feedback addressed in 37837782f:

- *Return health when settings storage is down (Codex P2)* — a failed settings lookup in `/admin/server/status` now skips the jellycompat restart derivation instead of aborting with a 500, so `postgres.ok: false` reaches the health strip in exactly the outage it exists to report. Test added.
- *Truncate playback buckets in UTC (Codex P2)* — `date_trunc($2, started_at, 'UTC')`, so bucket boundaries no longer depend on the session `TimeZone` (Postgres 18 floor, so the three-argument form is available).
- *Reset layout state when the account changes (CodeRabbit)* — the hook now drops entries, sync flags, and any queued save when the authenticated user id changes while mounted, then reloads the new account's scoped cache and re-adopts its server layout. Test covers the switch, the discarded queued save, and re-adoption.
- *Unresolved node query shown as "no nodes" (CodeRabbit)* — the node cell shows an em dash while the list is loading and "unavailable" when the query failed; "none / this server transcodes" now requires a successful empty response. Tests added for both states.

Validation: `go build`, `gofmt`, `golangci-lint --new-from-merge-base` (0 issues), focused handler tests, touched vitest files (51 tests), web lint/format/tsc — all clean.

Quick104:
Codex round three (*bound the settings lookup before probing health*) addressed in the latest commit: the settings `GetAll` in `/admin/server/status` now runs under the same two-second `adminHealthProbeTimeout` as the dependency probes, so a wedged pool can no longer stall the response — the lookup times out, the jellycompat derivation is skipped, and the health probes still report `postgres.ok: false` promptly.

Quick104:
1dfdd228e retunes the default layout toward what the popular media-server monitoring dashboards (Varken/Tautulli-style Grafana boards) lead with: live activity on top, ops at the bottom.

**New default arrangement** (12-column grid):

| Row | Widgets |
|---|---|
| 1 | Six 2-wide stat tiles: streams, egress now, transcode share, profiles, movies, shows |
| 2 | Users + storage tiles (2 each) + health strip (8) |
| 3 | Now playing (12×3) — moved up from below the charts |
| 4–5 | Playback activity \| Concurrent streams; Egress \| Reliability — all 6×3, equal-height pairs (fixes the grid hole next to the old 6×2 reliability card) |
| 6 | Top titles \| Top profiles (moved above content management) |
| 7 | Libraries (7) \| Users (5) |
| 8 | Server resources (6×3, was full-width) \| Nodes (6×3) |
| 9 | Scanner (4) \| Scan activity (8, was full-width) |
| 10+ | Trakt strip (12×1, fills its row), Recent errors (12×3), Recent activity |

**Nodes widget**: renamed from "Transcode nodes" (proxy nodes live there too) and redesigned — the header now carries a fleet summary (healthy count, total jobs, summed egress), rows are tighter with the status icon leading and per-node egress on the title line. The widget id stays `transcode-nodes`, so saved layouts keep resolving, and admins with a stored layout are unaffected — this only changes the starting arrangement.

Validation: dashboard vitest suite and full `make test-web` (2845 tests) pass, `tsc` and production build clean.

Quick104:
Codex round four addressed in 0f4fd90dd:

- *Anchor playback columns to the server window* — `/admin/stats/playback-activity` now returns `from`/`to` on the database clock (mirroring the timeseries endpoint) and the chart anchors its bucket grid on `to`, falling back to the browser clock only for responses predating the fields. Docs updated.
- *Count account-profile pairs in reliability* — the sessions CTE carries `user_id` and `unique_profiles` counts `DISTINCT (user_id, profile_id)`, matching `active_profiles`.
- *Record elapsed time before summing watch time* — true elapsed viewing time cannot be derived today (sessions record no start position), so each session's contribution is now capped at its wall-clock length: `LEAST(watched_seconds, ended_at - started_at)`. A session resumed at the one-hour mark that plays ten minutes now adds ten minutes, not seventy. Documented as an estimate until playback records elapsed time; happy to file a follow-up issue for start-position capture.

Quick104:
22a8407e7 lands the maintainer-directed follow-ups:

**Sane defaults** — the default layout places 15 widgets instead of all 25: six compact tiles → full-width health strip → Now playing → Playback | Egress → Top titles | Top profiles → Libraries | Users → Recent activity. Transcode share/User count tiles, Concurrent streams, Reliability, Server resources, Nodes, Scanner, Scan activity, Watch providers, Recent errors, and Downloads ship hidden — one click away in the Add-widget sheet. Saved layouts are untouched.

**Now playing collapses when idle** — widgets can declare `collapsedRows` and report emptiness through the chrome context; the grid renders the reduced height outside customize mode. Now playing shrinks to a one-row "Nothing playing right now" strip and grows back when a session starts.

**Scanner** — queue counters and autoscan state lead; the per-library scan list sits below.

**Watch providers (was Trakt)** — `GET /admin/stats` now carries a per-provider `watch_providers` array driven by the watchsync registry: plugin providers appear the moment they register (zeros if never synced); providers with stored history but no registration stay visible flagged `registered: false`. The Trakt-hardcoded `watch_provider_activity` object is removed (admin-web was its only consumer; recorded in the v1-scope pre-lock removals table). Stored layouts with the old `trakt-sync` widget id resolve via an alias, deduped.

**Downloads visibility** — download traffic was already inside `egress_kbps` via transfer-class telemetry, so the sampler now records an additive `download_egress_kbps` split (total keeps its meaning; the subset is clamped so derived playback can never go negative; replica-safe per-process rows unchanged). The Egress chart splits Playback vs Downloads once the split is measured. New `GET /admin/stats/downloads` (TTL cache + bus invalidation) reports users with downloads, active items, bytes on devices, 24h started/completed, and a per-user top list, rendered by a new Downloads widget. Both advertised on `/admin/dashboard/capabilities` (`watch_providers`, `downloads_stats`).

**Scrollbars** — widget scroll areas now follow the app's existing `overlay-scroll` theming, declared once on the grid container.

Full gate on the final tree: build/gofmt/vet, `golangci-lint --new-from-merge-base` (0 issues), full `make test-go`, `make test-web` (343 files / 2864 tests), web lint/format/tsc/build, settings-bindings/playback-fixtures/local-paths verifiers — all green.

Quick104:
60a356c7f makes the Nodes widget adapt to its size: nodes are sorted by type then name (fleets read as groups), narrow slots keep the compact stacked rows, and once the widget's measured body fits two 260px columns the nodes render as cards — identity on top, load meter with room to breathe, vitals as labelled pairs — with the column count following the widget's actual width rather than the viewport. Tested (sorting, row mode, card-grid threshold, empty state); dashboard suite, tsc, and lint clean.

Quick104:
Two more commits:

**c8182b350 — drag widgets from the Add-widget sheet.** Sheet entries are draggable (`application/x-silo-widget-add` payload, distinct from the reorder payload so neither path can be mistaken for the other); the grid's existing drop machinery accepts them with the same insertion indicator, placing the widget in a single layout update via `addWidget(id, beforeId?)`. The sheet is non-modal already, so during a drag it just fades pointer-transparent — kept mounted, since unmounting the drag source cancels HTML5 drags in some engines — and drops are announced on the existing aria-live channel. The + button stays as the keyboard path. Eight new tests.

**f8f558e5b — Codex round five**, all four findings verified and fixed:
- *Deduplicate finalized sessions from the live union* — live rows whose `session_id` already reached `playback_history_admin` are excluded, closing the double-count during the finalize-write → sync-delete interval (and after a failed deletion).
- *One clock for every sample source* — process egress rows are now bucketed on `date_trunc('minute', now())` like the shared row, with same-minute ticks merged by `GREATEST` (peak-preserving, and per-row `download ≤ total` survives independent maxima), so host clock skew can no longer misalign streams vs egress or write future rows the server-anchored grid drops.
- *Aggregate playback before widening buckets* — rather than fixing the subtraction, the chart no longer subtracts: the main line is always the total with Downloads as a second line. Past the two-hour resolution the total and download maxima are independent peaks from possibly different minutes, so a derived "playback" line was structurally wrong. Docs now say to chart the two series, not the difference.
- *Exclude every watch-provider source* — the activity filter is now an allowlist of on-server sources (`legacy`, `manual`, `playback`, `jellycompat`); plugin providers write arbitrary keys into `source`, so the denylist would have counted any newly installed provider's imported backlog as local plays.

Validation: full handlers + dashmetrics Go suites, `golangci-lint --new-from-merge-base` 0 issues, 500 dashboard/hooks web tests, tsc/lint/format/build clean.

Quick104:
Latest commit adds drag auto-scroll: while dragging a widget (reorder or from the Add-widget sheet) or corner-resizing one, the page edge-scrolls when the pointer is within 80px of the top or bottom of the visible scroll area, with speed ramping toward the edge. One requestAnimationFrame loop, stopped on every termination path; the scroll container is resolved by walking up from the grid, so an inner scroll region would work without changes. Resize sessions compensate for scrolled distance since pointer events don't re-fire under a stationary pointer. 12 new tests (pure speed/container helpers + a wiring smoke test); dashboard suite at 149 tests, tsc/lint/format/build clean.

Also: the Build Darwin arm64 failure on the previous push was `TestMountedABSRouterFeedFileResolvesOwner` in `internal/audiobooks/abs` — untouched by this branch and passing locally on the same platform; rerun in progress as a flake.

Quick104:
Codex round six addressed in the latest commit, all three verified real:

- *Allow reorder drops in the trailing area* — a reorder dropped past the last widget now appends (matching the sheet-add path) and shows the same after-the-last-widget indicator during dragover. Test added.
- *One snapshot for playback buckets and reliability* — both statements now run in a repeatable-read read-only transaction: one snapshot, and one `now()` (the transaction timestamp), so the reported window can no longer drift off the bucket filter across a boundary.
- *Publish invalidations on the subscribed playback channel* — the reconciler and cleanup worker now publish `ChannelPlayback` whenever sessions changed instead of only when the events hub is absent; the hub (client events) and the cache bus (cross-node invalidation of the stats/activity/leaderboard/timeseries caches) are different audiences, not alternatives.

Validation: worker + handlers Go suites, `golangci-lint --new-from-merge-base` 0 issues, 150 dashboard tests, tsc/lint clean.

Quick104:
Codex round seven addressed in the latest commit, all three verified:

- *Align watched-time filtering with the ranked window* — both watched-seconds CTEs now filter on `ended_at`, the same stop instant `watched_at` records, so boundary-straddling sessions count toward both plays and watch time (and the filter now matches the existing `idx_playback_history_admin_ended` index). Docs updated.
- *Sample current counters before calculating egress* — the sampler calls `Registry.Sweep()` (which collects live counters) instead of `Snapshot()` (which reports totals as of the last telemetry sweep); with `SILO_STREAM_TELEMETRY_SWEEP_INTERVAL` above a minute the old path read zeros between sweeps and then spiked.
- *Bound database calls in the sampler loop* — every tick runs under a 30s deadline (half the interval), so a wedged pool produces a bounded chart gap instead of a permanently stuck sampler goroutine.

Validation: dashmetrics + handlers suites, `golangci-lint --new-from-merge-base` 0 issues, build/gofmt clean.

PR #798, opened 2026-08-27
fix(auth): serialize initial administrator setup

Quick104:
## Problem

Related issue: N/A — audit-validated security fix

`POST /api/v1/auth/setup` checked whether any users existed and created the first administrator in separate autocommit operations. Two requests could both observe an empty `users` table, then create different administrator accounts and login sessions.

The finding was reproduced on fresh `origin/main` at `8164fd594b9fdd8c1944bb0b6251f2d00e4a24ca`: two synchronized setup requests both returned `201`, leaving two administrators and two sessions.

## Approach

Add a repository operation dedicated to first-user creation. It opens a PostgreSQL transaction, takes a `SHARE ROW EXCLUSIVE` lock on `users`, rechecks that the table is empty, inserts through the same transaction, and commits. Competing setup requests and ordinary user writers therefore serialize on the database connection that owns the transaction. If the connection is lost, PostgreSQL rolls the transaction back and releases the lock.

The normal account-provisioning path is unchanged. Initial setup reuses its profile creation and cleanup flow but selects the database-enforced first-user operation. No migration or API shape change is required.

The regression test uses two independent one-connection pools. A query tracer holds both setup attempts after their initial zero-user checks and releases them together. It asserts one token-pair success, one `ErrSetupAlreadyComplete`, and exactly one user, administrator, and session.

## Validation

Focused and database-backed checks:

- `go test ./internal/auth ./internal/database/pglock` — pass
- `go test -race ./internal/auth ./internal/database/pglock` — pass
- `go vet ./internal/auth ./internal/database/pglock` — pass
- `go test ./internal/api/handlers -run '^TestAuthProviders' -count=1` — pass
- `SILO_TEST_DATABASE_URL=<isolated-empty-database> go test ./internal/auth -run '^TestSetupInitialUserConcurrentAcrossPoolsDB$' -count=20` — pass
- `SILO_TEST_DATABASE_URL=<isolated-empty-database> go test -race ./internal/auth -run '^TestSetupInitialUserConcurrentAcrossPoolsDB$' -count=5` — pass

Full repository gate:

- `make embed-stub` — pass
- `go build ./...` — pass
- `gofmt -l .` — pass, no output
- `go vet ./...` — pass
- `golangci-lint run --new-from-merge-base="origin/main" ./...` — pass, 0 issues; it printed one warning about an unrelated stale worktree path
- `make test-go` on macOS — not passing because of repeatable failures in untouched packages: `internal/httpstream/TestReadFromRollsDeadlineUnderProductionStep` and `internal/jellycompat/TestBeginWebOperationRecoversDeadProcessLock` / `TestBeginWebOperationRejectsLiveProcessLock`
- `go test ./...` against the exact candidate source on the Linux dev-builder — pass, including `internal/httpstream` and `internal/jellycompat`
- `pnpm install --frozen-lockfile` — pass
- `pnpm run lint` — pass with 167 inherited warnings and 0 errors
- `pnpm run format:check` — pass
- `pnpm run build` — pass with existing asset/chunk-size warnings
- `make test-web` — pass: 294 files, 2,184 tests
- `make verify-settings-bindings-all` — pass
- `make verify-playback-fixtures` — pass
- `make verify-local-paths` — pass

Isolated dev-builder validation in `audit-bootstrap-admin`:

- controller orientation and `doctor` passed before and after validation
- a normal first setup returned `201` and produced one administrator, one user, and one session
- after resetting only the synthetic test identity, synchronized competing setup requests returned `201` and `401 setup_complete`
- the competing run produced one administrator, one user, and one session

## Risks

The table lock briefly blocks writes to `users` while first-user creation hashes the password and inserts the row. This happens only while the database has no users. Setup-status reads remain available. There is no migration, stored-data rewrite, client contract change, or Jellyfin compatibility change.

Rollback is a revert of this commit; no schema or data rollback is needed.

## AI Disclosure

- Tool(s): OpenAI Codex
- Model(s): n/a (the tool did not report a model identifier)
- Involvement: AI-assisted
- Adversarial review: A fresh read-only investigator independently reproduced the count-then-create race. A separate bypass reviewer rejected an earlier advisory-lock design because a one-connection pool could deadlock and lock-session loss could release the fence before side effects completed; it also found an over-broad test interface. The implementation was replaced with a same-transaction PostgreSQL table lock and emptiness check, and the interface was narrowed. The reviewer then rechecked the replacement for concurrent setup, ordinary user writers, connection loss, pool exhaustion, and API regressions and found no remaining concrete bypass or regression.

## Checklist

### DIFF
--- a/internal/auth/account_provisioner.go
+++ b/internal/auth/account_provisioner.go
@@ -14,6 +14,10 @@ type AccountUserRepository interface {
 	Delete(ctx context.Context, id int) error
 }
 
+type initialAccountUserRepository interface {
+	CreateInitial(ctx context.Context, input models.CreateUserInput) (*models.User, error)
+}
+
 type DefaultProfileOptions struct {
 	Enabled bool
 	Name    string
@@ -43,7 +47,28 @@ func (p *AccountProvisioner) CreateAccount(
 	ctx context.Context,
 	input CreateAccountInput,
 ) (*models.User, error) {
-	user, err := p.users.Create(ctx, input.User)
+	return p.createAccount(ctx, input, p.users.Create)
+}
+
+// CreateInitialAccount provisions the first account through the repository's
+// database-enforced empty-table claim, then applies the normal profile flow.
+func (p *AccountProvisioner) CreateInitialAccount(
+	ctx context.Context,
+	input CreateAccountInput,
+) (*models.User, error) {
+	initialUsers, ok := p.users.(initialAccountUserRepository)
+	if !ok {
+		return nil, fmt.Errorf("initial account creation unavailable")
+	}
+	return p.createAccount(ctx, input, initialUsers.CreateInitial)
+}
+
+func (p *AccountProvisioner) createAccount(
+	ctx context.Context,
+	input CreateAccountInput,
+	createUser func(context.Context, models.CreateUserInput) (*models.User, error),
+) (*models.User, error) {
+	user, err := createUser(ctx, input.User)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/auth/account_provisioner_test.go
+++ b/internal/auth/account_provisioner_test.go
@@ -10,11 +10,19 @@ import (
 )
 
 type stubAccountUsers struct {
-	createFn func(context.Context, models.CreateUserInput) (*models.User, error)
-	deleteFn func(context.Context, int) error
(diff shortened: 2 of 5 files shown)


Quick104:
Closing as superseded by current main's atomic initial-admin setup. The transaction, advisory lock, recheck, and regression tests already cover the race this PR addresses.

PR #805, opened 2026-08-28
perf(web): smooth detail layout, resizing, and navigation

blurbery:
## Problem

Related issue: #804

I reproduced three connected problems in the Web detail experience on a real deployment:

- seasons with many episodes wrapped into extra rows and could push or clip the compact hero;
- opening Brave's browser-chrome sidebar made detail-heavy show, season, episode, and movie pages visibly lag during viewport changes, scrolling, and pointer movement, while Home stayed smooth;
- Series → Season and Season → Series did not match the established Home → detail and Back motion. The destination content appeared late enough to make both directions look jagged.

The target is the stable Season 4 layout and one consistent detail-navigation path: keep the artwork, keep every episode in one horizontal row, make browser viewport changes cheap, and use the same smooth motion in both directions without breaking browser history.

### Broken: 12-episode view

A 12-episode season page with the hero title clipped above the viewport and episode cards wrapping into another row

### Desired: Season 4 layout

The desired Season 4 layout with the complete compact hero and one horizontal episode row

Closes #804.

## Approach

- Reuse `MediaCarousel` for season episodes so large seasons remain one horizontal Embla row. Compact heroes now use a minimum height, so long summaries grow downward instead of clipping the breadcrumb, title, metadata, or actions.
- Keep normal cursor behavior on blank carousel track space. Only real links, buttons, cards, and controls use the pointer cursor.
- Remove shared rendering costs that showed up in hover, scroll, and viewport traces: sampled backdrop blur and image filters, broad transitions, permanent animation hints, and repeated synchronous carousel resize work. Expensive supporting sections use bounded content visibility and layout/paint containment; image decode and prefetch work is deferred or coalesced.
- Centralize same-document navigation in a commit-aware `NavigationTransitionProvider`. It tracks browser-history provenance, waits for the intended route commit, preserves the existing Home/sidebar transition boundary, and avoids the old Series/Season Back loop. Error-state Back links follow the same reuse/replace rules; genuine Back/Forward traversals accept the browser's actual destination when stored provenance is unavailable, while ordinary links still require their exact intended route so a stale commit cannot end a newer animation.
- Give nested detail navigation Home's exact desktop motion: a lightweight committed shell at `±196px`, a two-painted-frame handoff, then `300ms` of `cubic-bezier(0.32, 0.72, 0, 1)` compositor motion. Destination detail content stays gated until that motion ends, so cards and text do not expand during the final frames.
[...]

PR #814, opened 2026-08-28
feat(web): directional page motion for nested detail navigation

Quick104:
## Problem

Related issue: N/A — the last outstanding piece of #805, rebuilt against `main`.

Series → Season → Episode never moved. Those navigations do not cross the
Home/item boundary, so the sidebar collapse that gives Home → item its motion never
runs, and `main-content` fell back to the same fade-out/slide-up used for an
unrelated route change. Going *up* a breadcrumb looked identical to going down.

## Approach

`main-content` slides horizontally on the sidebar's own duration and easing tokens,
in the direction the navigation is travelling, so a breadcrumb reads as the reverse
of the push that created it. Direction is stamped on
`html[data-navigation-direction]` before the router opens the transition; when it
cannot be derived the attribute is absent and the neutral transition plays rather
than guessing a side.

**This is deliberately not a navigation provider.** #805 proposed 568 lines that
hand-rolled a view-transition commit barrier, sequence numbers, an abandonment
timeout and a WAAPI slide, because `BrowserRouter`'s POP is async. `main` has since
moved to a data router, and reading the installed react-router 8.3.0 source shows it
owns all of that: `RouterProvider` holds its update callback until
`state.location.key === pendingState.location.key` (`lib/components.js:237`),
restarts on interruption, and replays transitions on POP from
`appliedViewTransitions` (`lib/router/router.js`). What is left is the one thing the
router does not do — which way the page is moving — and that is all this adds.

**Two decisions worth reviewing:**

*Root snapshot scoping.* The sidebar collapse is a live CSS transition running across
the same navigation, and a frozen root snapshot cross-fades the old 260px sidebar
over it. The root group is held still **inside the sidebar shell only**, keyed off
`data-app-shell`. `main-content` is named in `Layout` and nowhere else, so
suppressing the root globally would leave `/watch`, `/reader/*`, `/admin/*` and the
auth screens with no captured element at all — trading one visual defect for a
broader one.

*No history provenance map.* A path→index mirror would let a breadcrumb
`navigate(-n)` and unwind to an entry already behind the user instead of pushing a
duplicate. It cannot be made sound here: React Router derives a push index as
`getIndex() + 1` over `(globalHistory.state || { idx: null }).idx`
(`lib/router/history.js:282-284`, `:300`), so an entry landing without an `idx`
makes `null + 1 === 1` and restarts the numbering while the real stack keeps
growing. This app ships the trigger — the "Skip to content" links in `Layout` and
`AdminLayout` are raw `<a href="#main-content">`, which push an unindexed entry and
fire `hashchange`, which the router does not listen for. A stale delta lands the user
on a page they never asked for; a duplicate history entry is merely untidy.
Breadcrumbs push.

## Resolved during review: the Back half

The first revision of this branch shipped with browser Back falling back to the
neutral animation. Recorded here because the failure mode is worth knowing about.

`popstate` is a discrete event, so React flushes the router's state update
*synchronously inside the router's own listener*, which is registered at router
creation where ours could only be registered later from an effect. Ours always ran
second, after the location effect had advanced the tracked index to the destination,
so every pop compared an index against itself and cleared the attribute.

The flaw was two writers to the tracked index with no guaranteed order. The fix is
one writer: `resolveCommittedDirection` runs from the location effect, the only place
that sees every commit exactly once, and the popstate listener is gone. Fixed in
21d853ad and verified in a real browser — link click stamps `forward`, browser Back
stamps `back`, browser Forward stamps `forward`.

Worth knowing for how you weigh the tests: **every direction test uses `MemoryRouter`,
which registers no popstate listener at all**, so the entire suite passed while half
the feature was dead. The tests now drive real router navigations and say so in the
file header. They pin the arithmetic; the ordering is structural rather than racy,
which is what makes them meaningful.

## Validation

- `pnpm run build` — passed
- `pnpm exec vitest run` — 2667 passed
- `pnpm run lint` — 0 errors (168 warnings, unchanged from this branch's base)
- `pnpm run format:check` — passed

Verified live in a signed-in session: `data-app-shell` is set inside `Layout`, the
scoped root rule applies, the root still computes `view-transition-name: root` (so
routes outside `Layout` keep their snapshot), and a forward navigation stamps
`data-navigation-direction="forward"`.

## Risks

The direction attribute's whole value depends on landing before the browser
snapshots. Both directions are confirmed live, but only in Chrome — the ordering
argument is about React's discrete-event flushing rather than anything
browser-specific, so it should hold, but it has not been checked elsewhere.

`data-app-shell` is a new coupling between `Layout` and `app.css`. It is asserted in
`navigationTransitionCss.test.ts`, but that test greps stylesheet source — it cannot
prove a rule matches at runtime.

## Provenance

Replaces the navigation half of #805 by @blurbery. The profiling and the original
problem statement are theirs; the implementation is new, because the premise that
branch was built on (BrowserRouter, no native commit barrier) no longer holds.

## AI Disclosure

- Tool(s): Claude Code
- Model(s): claude-opus-5[1m]
- Involvement: Fully AI-generated, human verified — @Quick104 chose to drop the
  history-provenance unwind after review and directed the scope.
- Adversarial review: built by a multi-agent workflow (recon → implement → verify)
  whose review agent found a blocker and three majors against the first draft. The
  blocker (the provenance map sending users to the wrong page) and two majors were
  resolved by dropping the unwind entirely; the third by scoping the root
  suppression. The router-ownership claims were verified against the installed
  react-router source, not documentation, and a second opinion from a different
  model independently confirmed them before any code was written. The Back-half
  defect above was then found by live browser verification after the gate was green.

## Checklist

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -28,6 +28,7 @@ import { queryClient } from "@/lib/query-client";
 import { AuthProvider, useAuth } from "@/hooks/useAuth";
 import { useCurrentProfile } from "@/hooks/useCurrentProfile";
 import { useIsActingAdmin } from "@/hooks/useIsActingAdmin";
+import { useNavigationDirection } from "@/hooks/useNavigationDirection";
 import { ThemeProvider } from "@/hooks/useTheme";
 import { DateTimeFormatProvider, useDateTimeFormat } from "@/hooks/useDateTimeFormat";
 import { CustomThemeProvider } from "@/contexts/CustomThemeProvider";
@@ -200,6 +201,17 @@ function RouteAnnouncer() {
 
 function ScrollRestorationManager() {
   useScrollRestoration();
+  return null;
+}
+
+/**
+ * Tracks history provenance and the direction the page is moving in. A leaf
+ * rather than a call inside `AppShell`: the hook reads the location, and
+ * subscribing `AppShell` to it would re-render the whole provider stack on
+ * every navigation.
+ */
+function NavigationDirectionManager() {
+  useNavigationDirection();
   return null;
 }
 
@@ -760,6 +772,7 @@ function AppShell() {
                         <RouteChunkPrewarmer />
                         <RealtimeEventChannels />
                         <ScrollRestorationManager />
+                        <NavigationDirectionManager />
                         <RouteAnnouncer />
                         <QueryCacheManager />
                         <AppChrome />
--- a/web/src/app.css
+++ b/web/src/app.css
@@ -242,6 +242,30 @@
   }
 }
 
+/* Directional page motion for nested navigation. `--nav-slide-shift` is the
+   distance the OUTGOING page travels; the incoming page always enters from the
+   opposite side, so one pair of keyframes covers both directions.            */
+@keyframes nav-slide-out {
+  from {
+    opacity: 1;
+    transform: translateX(0);
+  }
+  to {
+    opacity: 0;
+    transform: translateX(var(--nav-slide-shift));
+  }
+}
+@keyframes nav-slide-in {
+  from {
+    opacity: 0;
+    transform: translateX(calc(-1 * var(--nav-slide-shift)));
+  }
+  to {
+    opacity: 1;
+    transform: translateX(0);
+  }
+}
+
 /* ================================================================
    VIEW TRANSITIONS — Page navigation animations
    ================================================================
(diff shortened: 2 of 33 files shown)


Quick104:
Fixed in 21d853ad. Good catch — the diagnosis was exactly right, including the
part about MemoryRouter being unable to reproduce it.

**Root cause.** Two writers to the tracked index with no guaranteed order between
them. `popstate` is a discrete event, so React flushes the router's state update
synchronously inside the router's own listener, and the router registers that
listener at creation, where ours could only be registered later from an effect.
Ours always ran second, after the location effect had advanced the index to the
destination, so every pop compared an index against itself.

**Fix.** One writer. `resolveCommittedDirection` runs from the location effect —
the only place that sees every commit exactly once — and the popstate listener is
gone. `useNavigationType()` distinguishes a POP, which has no click to have
stamped it and must be derived, from a push or replace, which the caller already
stamped with intent the index cannot recover. The ordering question disappears
rather than being won.

**Verified in a real browser**, since the suite structurally cannot:

| action | `data-navigation-direction` |
| --- | --- |
| link click | `forward` |
| browser Back | `back` (was cleared) |
| browser Forward | `forward` |

**On the tests.** They now drive real router navigations instead of dispatching
synthetic popstate events, and the file header says plainly what they cannot
prove: MemoryRouter registers no popstate listener, so all of them passed while
this was broken. They pin the arithmetic; the ordering is now structural rather
than racy, which is what makes them worth having at all.

Marking ready for review — the known limitation the description opened with is
resolved, and I've updated the body to match.

PR #843, opened 2026-08-30
fix(tonemap): enlarge the NVDEC probe fixture

Baine:
## Problem

The embedded HEVC Main 10 decoder fixture is 64x64. Pascal-generation NVIDIA NVDEC rejects that frame before tone mapping is exercised because its advertised minimum decode size is 144x144. On a GeForce GTX 1050 Ti this produced:

```text
Video width 64 not within range from 144 to 8192
```

The failed smoke test removes the NVENC hardware tone-map capability even though the GPU, driver, CUDA filter, and NVENC encoder can handle the real stream. Playback then reports that no validated HDR-preserving or tone-map recipe is installed, or that tone-map capability discovery is temporarily unavailable.

## Approach

Replace the deterministic 64x64 HEVC Main 10 frame with a deterministic 256x256 HEVC Main 10 frame. The probe remains self-contained: it does not require a media mount or an encoder to generate its input.

Add a regression test that inspects the embedded frame with `ffprobe` and requires:

- exactly one video stream;
- width and height of at least 144 pixels; and
- `yuv420p10le` pixel format.

256x256 is intentionally small while clearing Pascal NVDEC's minimum dimension. The `ffprobe` test skips when that external binary is unavailable; the normal unit tests remain usable in minimal development environments.

## Validation

Real deployment validation used Silo in Docker on Unraid with NVIDIA driver 580.178.04 and a GeForce GTX 1050 Ti:

- The previous fixture failed deterministically with `Video width 64 not within range from 144 to 8192`.
- The replacement fixture completed the same CUDA decode / `tonemap_cuda` / `h264_nvenc` smoke path successfully.
- With this fixture and the companion cold-start timeout fix, Silo reported `HW NVENC` and `HW Tone Map` while converting HEVC 2160p to H.264 1608p HLS. Software HDR tone mapping was disabled. The simultaneous TrueHD Atmos 7.1 to AAC 5.1 conversion is an independent audio transcode.

Local validation on Linux, Go 1.26.4:

```text
$ go test -v ./internal/tonemap -run '^TestDecodeProbeFixtureCoversPascalNVDECMinimum$' -count=1
--- PASS: TestDecodeProbeFixtureCoversPascalNVDECMinimum (0.03s)

$ go test ./internal/tonemap
ok  github.com/Silo-Server/silo-server/internal/tonemap

$ go build ./...
PASS

$ go vet ./...
PASS

0 issues.

$ make verify-settings-bindings-all verify-playback-fixtures verify-local-paths
settings bindings are current
web settings binding is current
playback fixtures are current
```

`make test-go` passed every package except the existing `TestRemoteTranscodeStartTimeoutCoversColdProbePreflightAndReadiness` failure in `internal/jellycompat` (`3m43s`, expected `3m29s`). The same test was run on the unmodified `upstream/main` commit and failed with the identical values, confirming that it is a baseline failure unrelated to this diff.

## Risks

The embedded fixture is larger, increasing the binary by less than 1 KiB. Probe behavior is otherwise unchanged. GPUs that accepted the 64x64 frame also accept the standards-compliant 256x256 frame.

## AI Disclosure

- Model(s): GPT-5
- Involvement: AI-assisted
- Adversarial review: Reviewed the complete diff and probe call path; validated the fixture metadata with real `ffprobe`; ran the exact CUDA decode, tone-map, and NVENC encode path on a GTX 1050 Ti; separated an independent cold-start timeout into a second pull request; and reproduced the only full-suite failure unchanged on `upstream/main`.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hardware-accelerated video decoding detection for compatible NVIDIA Pascal GPUs.
  - Prevented capable GPUs, including the GeForce GTX 1050 Ti, from being incorrectly identified as unsupported.

- **Tests**
  - Added validation that the decoding probe uses sufficient video dimensions and the expected 10-bit pixel format.

### DIFF
--- a/internal/tonemap/probe.go
+++ b/internal/tonemap/probe.go
@@ -32,13 +32,16 @@ const (
 	probeRequestSlack  = 5 * time.Second
 )
 
-// One deterministic 64x64 Main 10 HEVC frame. Keeping the compressed fixture
+// One deterministic 256x256 Main 10 HEVC frame. Keeping the compressed fixture
 // in the binary lets production probes exercise the real decoder without
 // depending on a media mount or generating source files with an encoder whose
-// availability is itself under test.
+// availability is itself under test. The dimensions deliberately exceed the
+// 144x144 minimum reported by Pascal-generation NVIDIA NVDEC; the old 64x64
+// fixture produced a false negative on otherwise capable GPUs such as the
+// GeForce GTX 1050 Ti.
 const decodeProbeFixtureBitDepth = 10
 
-const decodeProbeFixtureBase64 = "AAAAAUABDAH//wIgAAADAJAAAAMAAAMAHpWUCQAAAAFCAQECIAAAAwCQAAADAAADAB6gIIEE2WVlSkwvAWgIAAADAAgAAAMACEAAAAABRAHAc8CJAAABKAGsTtcff/U+nK/q+A=="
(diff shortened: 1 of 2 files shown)


Baine:
Addressed the diff-scoped docstring warning in 5a58b621 by documenting both analyzed test functions. The focused tonemap test and merge-base golangci-lint run pass. CodeRabbit currently reports its follow-up review as rate limited, so the warning in the earlier summary may remain stale until its next refresh.

PR #844, opened 2026-08-30
fix(tonemap): allow cold NVENC probes to initialize

Baine:
## Problem

Related issue: #843

Tone-map smoke commands currently share a 5-second per-command deadline across every backend. That is long enough for warm NVENC probes, but not for CUDA/NVENC initialization on some cold systems.

On an Unraid Docker host with NVIDIA driver 580.178.04 and a GeForce GTX 1050 Ti, the exact single-frame PQ CUDA/NVENC smoke command completed successfully in 14.916 seconds from a cold start. Under the existing 5-second deadline, valid PQ and HLG probes were canceled. Hardware tone mapping was therefore omitted from the capability inventory and playback returned `tone map capability discovery is temporarily unavailable`.

This is independent of the undersized decoder fixture fixed by #843. Both conditions were present on the reported Pascal system.

## Approach

Give only NVENC hardware smoke commands a 30-second deadline. Listing commands, software tone mapping, QSV, VAAPI, and VideoToolbox retain the existing 5-second deadline because no evidence supports widening them.

The timeout is attached to each smoke command rather than made global. A successful command returns immediately; 30 seconds is only the failure ceiling. The extra headroom avoids setting a 15-second limit immediately adjacent to the measured 14.916-second cold start.

The change also recalculates every composed budget from the backend-specific command deadline:

- total local probe timeout;
- remote endpoint timeout;
- advertised request timeout; and
- the maximum accepted remote-node timeout.

The adversarial review found that the last ceiling was still priced from QSV. That would have allowed a node to advertise the larger NVENC budget and then had the caller clamp it below the node's own deadline. `MaxProbeRequestTimeout` now uses NVENC, the slowest supported command matrix, and the device-cap regression test covers QSV, VAAPI, NVENC, and VideoToolbox.

## Validation

Measured on the affected GTX 1050 Ti deployment with the production FFmpeg command shape and the corrected fixture from #843:

```text
source kind  5-second result  elapsed
pq           timeout (124)    5017 ms
hlg          timeout (124)    5019 ms
hlg_bt709    timeout (124)    5014 ms
sdr_bt709    success          2548 ms
sdr_bt2020   success          3138 ms
```

The cold PQ command completed successfully when allowed to continue:

```text
exit: 0
elapsed: 14.916341445 s
```

A subsequent warm PQ run completed in 839 ms. This confirms that the failure was initialization latency rather than an unsupported decoder, filter, or encoder.

End-to-end deployment validation with software HDR tone mapping disabled reported:

```text
container: HLS
video:     HEVC 2160p -> H.264 1608p
mode:      HW NVENC
tone map:  Hardware
audio:     TrueHD Atmos 7.1 -> AAC 5.1
```

The audio conversion is independent of the video capability fix.

Local validation on Linux, Go 1.26.4:

```text
$ go test ./internal/tonemap -count=1
ok  github.com/Silo-Server/silo-server/internal/tonemap

$ go build ./...
PASS

$ go vet ./...
PASS

0 issues.

$ pnpm run build
PASS

$ pnpm run lint
0 errors, 174 pre-existing warnings

$ pnpm run format:check

$ pnpm test
360 test files passed; 3019 tests passed

$ make verify-settings-bindings-all verify-playback-fixtures verify-local-paths
settings bindings are current
web settings binding is current
playback fixtures are current
```

`make test-go` exercised the complete suite but was not fully green locally:

- `TestNonToneMapStartDoesNotProbeLiveMetadata` encountered `text file busy` while executing a newly written temporary script; the isolated rerun passed.
- `TestRemoteTranscodeStartTimeoutCoversColdProbePreflightAndReadiness` expects `3m29s` but receives `3m43s`. The isolated test fails with the identical values on unmodified `upstream/main`, so this is an existing baseline mismatch.

An earlier resource-constrained container run also timed out in unrelated Go and Web packages; each reported Go package and all four reported Web files passed when rerun separately. The final local Web run above passed in full.

## Risks

A genuinely broken NVENC path can now take up to 30 seconds per failed source-kind smoke command to classify. Discovery is shared and cached, successful commands return immediately, and other backends retain their previous limit. Remote endpoint and request budgets grow consistently with the work they permit, avoiding a caller-side timeout before the node's probe can finish.

No settings, API schema, migrations, or playback command graphs change.

## AI Disclosure

- Model(s): GPT-5
- Involvement: AI-assisted
- Adversarial review: Reviewed the complete diff and all timeout compositions; compared cold and warm production FFmpeg timings; limited the widened deadline to NVENC; rejected a global timeout increase; found and fixed the stale remote maximum-budget calculation; expanded backend/device-cap tests; and separated the decoder-fixture compatibility issue into #843.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling for hardware video probing.
  * NVENC detection now allows additional time for slower commands while other hardware backends retain their existing limits.
  * Probe request timeout calculations now more accurately reflect the selected hardware backend.
  * Improved validation across all supported hardware backends.

### DIFF
--- a/internal/tonemap/probe.go
+++ b/internal/tonemap/probe.go
@@ -20,8 +20,13 @@ import (
 
 const (
 	probeCommandTimeout = 5 * time.Second
-	probeNegativeTTL    = 15 * time.Second
-	probeTimeoutSlack   = time.Second
+	// An NVENC smoke test initializes the NVIDIA driver, decoder, CUDA tone-map
+	// filter, and encoder in a fresh FFmpeg process. Older cards can take longer
+	// than the ordinary command budget on the first CUDA use even when the same
+	// pipeline is fully supported once initialized.
+	nvencProbeCommandTimeout = 30 * time.Second
+	probeNegativeTTL         = 15 * time.Second
+	probeTimeoutSlack        = time.Second
 	// probeEndpointSlack covers what a capability endpoint spends around the
 	// tone-map matrix itself: one bounded hardware detection walk (30s in
 	// playback.hwAccelWalkTimeout), the transformation registry's three 3s
@@ -272,13 +277,14 @@ func capabilityCoversAllSourceKinds(capabilities Capabilities, mode Mode, backen
 // ProbeTotalTimeout budgets one bounded deadline for every listing and smoke
 // command the selected backend and device set can execute.
 func ProbeTotalTimeout(hardwareBackend, hardwareDevice string) time.Duration {
-	commandCount := 2 + len(AllSourceKinds())
+	total := time.Duration(2+len(AllSourceKinds())) * probeCommandTimeout
 	backend := strings.ToLower(strings.TrimSpace(hardwareBackend))
 	switch backend {
 	case BackendQSV, BackendVAAPI, BackendNVENC, BackendVideoToolbox:
-		commandCount += len(AllSourceKinds()) * len(probeDevices(hardwareDevice, backend))
+		hardwareCommandCount := len(AllSourceKinds()) * len(probeDevices(hardwareDevice, backend))
+		total += time.Duration(hardwareCommandCount) * hardwareProbeCommandTimeout(backend)
 	}
-	return time.Duration(commandCount)*probeCommandTimeout + probeTimeoutSlack
+	return total + probeTimeoutSlack
 }
 
 // ProbeEndpointTimeout includes auto-backend discovery and response overhead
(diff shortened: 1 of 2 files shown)


Baine:
Addressed the diff-scoped docstring warning in d6724dda by documenting the four previously undocumented helpers and tests. The focused tonemap test, full build, go vet, and merge-base golangci-lint run pass. CodeRabbit currently reports its follow-up review as rate limited, so the warning in the earlier summary may remain stale until its next refresh.

PR #854, opened 2026-08-31
fix(web): refresh access token on 401 in playerFetch and clarify expired session message (fixes #851)

### LINKED ISSUE #851: [bug] Playback dies mid-session with "You do not have permission to play this item" when the access token expires — the player never refreshes it, and reports 401 as a permissions error
## What happened

Playback in the web player fails mid-session with:

> **Playback unavailable**
> You do not have permission to play this item. *(Go Back)*

It appears during an established stream rather than at start, most often after the video has been **paused for a while and then resumed**. Reloading the page fixes it immediately and the same item plays fine, so it is not an entitlement problem with the item or the account.

## Where it goes wrong

Two separate defects combine here — one functional, one in the message.

### 1. The player has no token refresh, so an expired access token is terminal

`web/src/player/player-fetch.ts` is a deliberately isolated wrapper — its header says it "never imports app-specific code". It reads the token once, attaches it, and on any non-2xx throws immediately:

```ts
const token = config.getAccessToken();
if (token) headers["Authorization"] = `Bearer ${token}`;
…
if (!res.ok) {
  …
  throw new PlayerFetchError(res.status, message, code, text);
}
```

There is no 401 branch. The app's own client has one — `web/src/api/client.ts:500-512` refreshes on 401 and replays the request — but `playerFetch` never reaches it, which is the cost of that isolation. So a request that any other part of the app would have silently recovered from is fatal inside the player.

Refresh is also purely reactive: it fires on a 401 response, with no proactive timer. A paused player issues no requests, so nothing renews the token while it sits. On resume, the first request carries a token that expired during the pause, and the player has no path back.

Reloading the page works because startup re-authenticates through the app-shell path, which does refresh.

This matches the pause-dependence in the report: the longer the pause, the likelier the token has expired.

### 2. A 401 is reported as a permissions problem

`web/src/player/playback-errors.ts` maps both statuses to the same text:

```ts
if (error.status === 401 || error.status === 403) {
  return {
    title: "Playback unavailable",
    message: "You do not have permission to play this item.",
  };
}
```

401 means the credential expired; 403 means the account is not allowed. Collapsing them tells the user their permissions are wrong when their session simply aged out, which sends anyone diagnosing this toward access groups and entitlements rather than auth lifetime.

Notably the app's own mapper already draws the distinction correctly (`web/src/api/client.ts:299-303`):

```ts
if (res.status === 401) return "Authentication required.";
if (res.status === 403) return "You do not have permission to perform this action.";
```

The player module also already has suitable copy for the expiry case — the `session_expired` terminal renders "This playback session is no longer active. Start it again to keep watching."

## Expected behavior

1. `playerFetch` should refresh the access token and replay the request once on a 401, matching `api/client.ts`. A token expiring mid-session should be invisible to the viewer.
2. Failing that, a 401 should not be presented as a permissions error. Separating it from 403 would at least point at the real cause.

## Steps to reproduce

1. Start playback in the web player.
2. Pause for longer than the access-token lifetime (`auth.access_token_expiry`; 1 hour on the server I saw this on, 8h by default), leaving the tab idle so no other request triggers a refresh.
3. Resume.
4. The error appears; reloading the page restores playback.

## Environment

- Silo `506 · 8592f439`
- Web player, Chrome
- `auth.access_token_expiry` set to 1h on this deployment; access-token payloads confirm a 3600s lifetime

Hostname, tokens, and account identifiers are omitted — none are needed to describe the failure.

rangoDJ:
### Summary
Fixes #851.

When web player requests returned 401 (e.g. after pausing video longer than access token expiry), \playerFetch\ had no token refresh handler and failed immediately. Additionally, \describePlaybackTransportError\ mapped 401 and 403 to the same permissions error message (\You do not have permission to play this item.\).\
\
### Changes\
- Added optional \
efreshToken?: () => Promise<boolean>\ callback to \PlayerConfig\ in \web/src/player/context/PlayerConfigContext.tsx\.\
- Exported \ttemptRefresh\ in \web/src/api/client.ts\ and wired \
efreshToken: () => attemptRefresh()\ into \playerConfig\ in \web/src/playback/WatchPlaybackChrome.tsx\.\
- Intercepted 401 responses in \playerFetch\ to automatically invoke \config.refreshToken()\ and replay the request once on success.\
- Updated \describePlaybackTransportError\ in \web/src/player/playback-errors.ts\ to present 401 as \Authentication required\ / \Your playback session has expired...\ and 403 as \Playback unavailable\.\
- Added unit tests in \web/src/player/player-fetch.test.ts\ and \web/src/player/playback-errors.test.ts\.

## Summary by CodeRabbit

- **New Features**
  - Playback now automatically refreshes expired access tokens and retries requests once when authentication expires.
  - Added support for playback environments to provide token-refresh behavior.

- **Bug Fixes**
  - Improved playback error messaging by distinguishing authentication-required errors from permission-denied errors.
  - Users are prompted to restart playback or sign in when authentication is required.
  - Playback requests now correctly retry with refreshed authentication details.

### DIFF
--- a/internal/jellycompat/handlers_items_test.go
+++ b/internal/jellycompat/handlers_items_test.go
@@ -810,7 +810,6 @@ func TestHandleEpisodes_UnknownSeasonIDReturnsNotFound(t *testing.T) {
 	}
 }
 
-
 // TestHandleUpcoming_InvalidSeriesId_ReturnsEmptyNot404 — same contract for
 // undecodable IDs. Decode failure must NOT 404 for the same reason.
 func TestHandleUpcoming_InvalidSeriesId_ReturnsEmptyNot404(t *testing.T) {
--- a/internal/jellycompat/parent_browse_test.go
+++ b/internal/jellycompat/parent_browse_test.go
@@ -2,7 +2,6 @@ package jellycompat
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http/httptest"
 	"net/url"
@@ -11,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Silo-Server/silo-server/internal/catalog"
 	"github.com/Silo-Server/silo-server/internal/config"
 	"github.com/Silo-Server/silo-server/internal/models"
 )
@@ -22,7 +22,7 @@ func (f *fakeSeasonByIDRepo) GetByID(_ context.Context, id string) (*models.Seas
 	if s, ok := f.seasons[id]; ok {
 		return s, nil
 	}
-	return nil, errors.New("season not found")
+	return nil, catalog.ErrSeasonNotFound
 }
 
 // fakeSeasonEpisodeRepo implements episodeRepoForBatchLoader, serving episodes by
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -206,7 +206,7 @@ function getDeviceHeaders(): Record<string, string> {
   };
 }
 
-async function attemptRefresh(): Promise<boolean> {
+export async function attemptRefresh(): Promise<boolean> {
   const rt = getRefreshToken();
   if (!rt) return false;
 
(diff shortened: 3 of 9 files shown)


Quick104:
Closing as superseded by #939, which added player token refresh and session error handling now present in main.

Issue #888, opened 2026-09-01
Schema workstream: 1.0 hygiene and migration reset (umbrella)


Quick104:
Part of the API v2 program epic #135. Durable architecture: `docs/architecture/api-contract.md` (PR #880). Runs in parallel with the API phases; everything below ships in or before the bridge except the reset itself.

Grouped deliverables:
1. **Dead-table drops** — `plex_sync_*` ×4, `user_playback_sessions`, `content_id_migration_map`: PR #879.
2. **Integrity** — the 14 missing `user_id` foreign keys (partitioned `policy_decisions*` may stay exempt), leading indexes for the 46 unindexed FK columns (prioritize `auth_sessions.user_id`, `downloads.media_file_id`, `library_collection_items.media_item_id`), `users.role`/`users.enabled` NOT NULL.
3. **Standardization** — 36 serial-style tables to `GENERATED AS IDENTITY`; 14 `varchar(n)` columns in notifications/push to `text`; consistency renames (`oauth_session`, `oauth_completion`, `playback_history_admin`).
4. **Bigint widenings** — `users.id`, `media_files.id`, `media_folders.id`: table rewrites; separate PR with an explicit maintenance-window release note.
5. **Bridge-timing** — drop `user_settings` + `user_setting_migration_rejects` after the Go settings backfill; bridge is the last release with the one-way SQLite→Postgres userdb import.
6. **1.0 reset** — migration squash to one baseline + guard (empty or bridge-tip only), seed-data fidelity audit (~53 seeding migrations incl. backfill-completion flags), CI parity gate comparing normalized schema **and** seed data; the parity gate must exist before the baseline PR opens (Phase 6).

**Blocked on maintainer decisions (do not implement until recorded on #135):** which library-restriction layers survive (per-user `library_ids`, access-group list, per-profile allow list, profile-scoped setting), and whether the per-user policy override columns on `users` remain a documented override chain or fold into access groups.

Migration discipline is absolute per AGENTS.md gotchas: `make migrate-create` only, no `goose fix`, no paired up/down files, no renumbering, no encrypted `server_settings` renames.

AI disclosure: drafted by Claude Code (claude-fable-5) orchestrating the #135 program.

Quick104:
[...]

PR #879, opened 2026-09-01
chore(db): drop dead plex_sync, user_playback_sessions, and content_id_migration_map tables [schema-hygiene 1/4]

Quick104:
## Problem

Related issue: #135

The 1.0 schema workstream locks the database schema alongside the API v2 contract, and the squashed 1.0 baseline should not inherit dead tables. The effective schema still carries six: the four `plex_sync_*` tables (superseded by `webhook_sync_*`; migration 060 already copied their data), `user_playback_sessions` (001-era, referenced by no code), and `content_id_migration_map` (artifact of the completed deterministic-content-ID migration).

Base branch: `main`. This is the first schema-hygiene PR; later schema PRs (integrity FKs/indexes, type/identity standardization, bigint widenings) stack on it.

## Approach

One Goose migration (`migrations/sql/20260901152535_drop_dead_tables.sql`, created with `make migrate-create`) drops the six tables, children before parents. The last vestigial code references go in the same change: two `NOT EXISTS` guards in the catalog orphan-cleanup predicate and five plex_sync merge steps in metadata provider-ID integrity (the `webhook_sync_item_state` step remains). The regression test now asserts the predicate must *not* reference the dropped tables while keeping all 13 durable guards.

The Down recreates all six tables empty with their original DDL — including the `COLLATE "C"` that migration `20260612130000` applied to `media_item_id` columns — following the migration 156 precedent. The migration comment and `docs/architecture/deterministic-content-id.md` record that the content-ID reverse remap becomes a no-op on databases past this point.

Deadness evidence: repo-wide sweeps (Go including string-built SQL, SQL, web, tests) hit only historical migrations; `git log -S` shows the plexsync writer package was deleted in the same commit that added migration 060's data copy. The `legacyPlexSync*` identifiers in `webhook_sync.go` are JSON DTO names over the live `webhook_sync_*` tables.

## Validation

- Full chain from an empty scratch database: applied cleanly; `migrate-status` 357 applied / 0 pending; all six tables absent (`to_regclass` NULL).
- `make migrate-down-to VERSION=20260830022104`: six tables recreated; `pg_dump --schema-only` byte-identical to the pre-drop baseline (columns, defaults, collation, constraint/index/FK names). Re-up clean (up→down→up idempotent).
- Independent review reproduction: plex_sync rows seeded at migration 59, chain run to HEAD — data present in `webhook_sync_connections` / `webhook_sync_profile_mappings` / `webhook_sync_item_state` before the drop applies.
[...]

Issue #135, opened 2026-06-11
API v2 program epic: /api/v2 contract reset on Huma, bounded bridge, v1 retirement at 1.0


Quick104:
## Program decision (supersedes the original proposal, preserved below)

Silo 1.0 ships `/api/v2` as the first stable native API, implemented with Huma v2 on the
existing chi router. `/api/v1` is a frozen alpha contract: carried through exactly one published
pre-1.0 bridge release, then retired — the 1.0 main API listener serves a `410 Gone` tombstone
(`client_upgrade_required`, `Cache-Control: no-store`) for the whole `/api/v1` namespace. The
database schema resets at 1.0 with a squashed baseline; the bridge is the mandatory upgrade
waypoint (0.x → bridge → 1.0).

This replaces the original incremental-v1 / generated-native-client proposal:

- v2 major with a hard cutover, not additive-only v1 evolution;
- web uses generated TypeScript types from the committed OpenAPI 3.1 artifact; Apple and
  Android keep handwritten lenient models proven against vendored fixtures with a `SOURCE`
  commit pin (no Swift/Kotlin codegen);
- deterministic operation registration, committed byte-reproducible spec, semantic-diff CI;
- no compatibility shims except the one bounded bridge release;
- Postgres is the only per-user store at 1.0 (SQLite userdb retired; the bridge keeps the
  one-way import).

Durable architecture: `docs/architecture/api-contract.md` and the draft administrator guide
`docs/update-to-1.0.md`, landed by PR #880. This issue and its child issues are the durable
program state; each phase ends with a checkpoint comment here.

## v1 feature freeze

Declared as of Phase 0: `/api/v1` receives critical bridge fixes only; all new contract work
targets v2. Scope-freeze exceptions (already-approved v1 work allowed to finish): none
recorded yet — maintainer to confirm.

## Phases and child issues

| Phase | Scope | Issue |
|---|---|---|
| 0 | Program setup: durable docs, policy updates, this epic | PR #880 |
| 1 | Exhaustive route/consumer inventory, migration ledger, tier-1 catalogs | #881 |
| 2 | Contract foundation: Huma mount, wire rules, error adapter, OpenAPI CI, fixtures | #882 |
| 3 | Client harnesses (web/Apple/Android) + pilot slice | #883 |
| 4 | Section-by-section vertical migration (one PR per section; section issues from the Phase 1 ledger) | #884 |
| 5 | Bridge release | #885 |
[...]

PR #889, opened 2026-09-01
chore(db): add missing user FKs, FK leading indexes, and users role/enabled NOT NULL [schema-hygiene 2/4]

Quick104:
## Problem

Account deletion is a bare `DELETE FROM users WHERE id=$1`, but 17 user-referencing columns had no foreign key — every deletion leaks rows (live push routes, webhooks, notification prefs, cached recommendations, in-flight Discord link state, and more). Separately, 46 FK leading columns had no usable index (so FK cascade checks sequential-scan child tables, worst on `auth_sessions.user_id` and `downloads.media_file_id`), and `users.role`/`users.enabled` allowed NULL that no code path writes or handles. The 1.0 squashed baseline should not inherit any of this.

**Base branch: `schema/hygiene-drop-dead-tables` (PR #879)** — second PR in the schema-hygiene stack; only the last two commits are new here.

## Approach

Three migrations, all derived from an empirical audit of the migrated schema (audit queries are committed in the migration comments):

1. **`20260901161657_fk_leading_indexes.sql`** — 57 `CREATE INDEX CONCURRENTLY` (`-- +goose NO TRANSACTION`), covering the 46 audited unindexed FK leading columns, 10 newly constrained columns, and 2 columns whose only leading index is partial with a predicate on a *different* column (`watch_provider_scrobble_sessions.connection_id`, `page_sections.library_id`). The audit rule is strict: partial indexes never count as FK coverage unless their predicate is `<column> IS NOT NULL` (nine such documented exceptions, argued in the migration).
2. **`20260901161700_user_fk_integrity.sql`** — 17 FKs to `users(id)`: 14 `ON DELETE CASCADE` (row belongs to the account) and 3 `ON DELETE SET NULL` (attribution on records that outlive the account: `activity_log.impersonator_user_id`, `subtitle_ai_jobs.requested_by`, `metadata_translation_jobs.requested_by`). `auth_sessions.impersonator_user_id` is deliberately CASCADE — SET NULL would silently convert a live impersonation session into an ordinary session for the impersonated user. Each `ADD CONSTRAINT` is preceded by an orphan sweep scoped exactly to rows whose parent no longer exists (targeted `DELETE`, or `UPDATE … SET NULL` for the attribution columns). Documented exclusions: `policy_decisions*` (per program plan: hottest write path, rows are historical evidence), `notification_server_channels.created_by_user_id` (informational-only by that table's recorded convention), foreign-system identity columns, and two text-typed Silo user-id columns that need a retype first (noted in #888).
3. **`20260901161702_users_role_enabled_not_null.sql`** — backfills `role='user'` / `enabled=true` (the code-side defaults), then `SET NOT NULL`.

`migrations/schema_integrity_test.go` (new, in the established migrations contract-test style) pins the 17 exact `ON DELETE` actions, each paired orphan sweep, the backfill values, the 57 index names, and that no `CREATE INDEX` statement gains a `WHERE` predicate (case-insensitive) — mutation-tested: flipping a CASCADE or lowercasing a predicate fails the test.

## Validation

- Full chain from an empty scratch DB (including the PR #879 drops): clean apply; migrate-status captured; down → re-up of all three migrations clean.
- Orphan-seeded upgrade run: orphan rows planted in representative tables plus NULL `role`/`enabled`; migration sweeps exactly the orphans, constraints then hold. Independently reproduced by the destructive-safety reviewer with its own seed data (parents/children/orphans) — only orphans touched.
- `EXPLAIN` confirms the prioritized indexes are present and usable.
- `make migrate-validate`, `make embed-stub`, `go build ./...`, `gofmt` (empty), `go vet`: pass.
- `golangci-lint run --new-from-merge-base="origin/main"`: 0 issues.
- `make test-go`: all packages ok (includes the new contract tests). `make verify-settings-bindings-all`, `verify-playback-fixtures`, `verify-local-paths`: pass.
- Web untouched on this branch (same web tree as PR #879, whose Web CI is green).

## Risks

- The FK migration runs transactionally and holds `SHARE ROW EXCLUSIVE` on `users` and each child during validation — on large tables that blocks writes for the duration. Accepted as documented maintenance-window behavior per the program's upgrade posture (a `NOT VALID`/`VALIDATE` split was considered and declined for pre-1.0 simplicity; the migration comments note it).
- The index migration uses `CONCURRENTLY` outside a transaction; the Up includes recovery handling for invalid leftover indexes if a run is interrupted.
- Orphan sweeps delete garbage rows (unreachable by any code path since their account is gone) and null out attribution for deleted accounts; the Down restores constraints/nullability but deleted rows are not restorable — restore from backup, as the migration states.

## Checklist

## AI Disclosure

- Harness: Claude Code
- Model(s): claude-opus-5 (implementation), claude-fable-5 (orchestration and all review agents)
- Adversarial review: two rounds of independent blind panels (destructive-safety at maximum effort, migration discipline, audit fidelity, test adequacy). Round 1 found one major (the index audit counted partial indexes as FK coverage — `watch_provider_scrobble_sessions.connection_id` was uncovered on an unboundedly-growing table) and three minors; all fixed. Round 2 verified closure and mutation-tested the new contract tests, yielding two minor prescriptions (partition-aware audit-count comment; case-insensitive predicate guard), both applied. The `NOT VALID` lock-window suggestion was declined with a written reason (documented maintenance-window posture).

### DIFF
--- a/internal/playback/planstore/postgres_test.go
+++ b/internal/playback/planstore/postgres_test.go
@@ -92,7 +92,7 @@ func newPlanstoreFixture(t *testing.T) *planstoreFixture {
 		t.Fatalf("insert fixture media folder: %v", err)
 	}
 	if err := pool.QueryRow(ctx, `
-		INSERT INTO users (username) VALUES ($1) RETURNING id`, unique).Scan(&f.userID); err != nil {
+		INSERT INTO users (username, role) VALUES ($1, 'user') RETURNING id`, unique).Scan(&f.userID); err != nil {
 		t.Fatalf("insert fixture user: %v", err)
 	}
 	if err := pool.QueryRow(ctx, `
--- a/migrations/schema_integrity_test.go
+++ b/migrations/schema_integrity_test.go
@@ -1,0 +1,360 @@
+package migrations
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// The three migrations in this release are load-bearing in ways a reader cannot
+// see from the file name: the ON DELETE action chosen per table decides whether
+// deleting an account destroys or merely detaches a row, each ADD CONSTRAINT is
+// only appliable because a sweep for orphans runs first, and the two backfill
+// values decide what a pre-existing NULL becomes forever. These tests pin all of
+// that as exact text, so changing an action, dropping a sweep, or reordering the
+// pair fails here instead of on someone's database.
+
+const (
+	fkLeadingIndexesMigration = "sql/20260901161657_fk_leading_indexes.sql"
+	userFKIntegrityMigration  = "sql/20260901161700_user_fk_integrity.sql"
+	usersNotNullMigration     = "sql/20260901161702_users_role_enabled_not_null.sql"
+)
+
+var sqlLineComment = regexp.MustCompile(`(?m)--.*$`)
+
+// readMigrationSections returns the Up and Down halves of a migration with line
+// comments removed and whitespace collapsed, so assertions can quote statements
+// as single-line SQL and compare positions without prose interfering.
+func readMigrationSections(t *testing.T, name string) (up, down string) {
+	t.Helper()
+	raw, err := FS.ReadFile(name)
+	if err != nil {
+		t.Fatalf("read migration %s: %v", name, err)
+	}
+	// Strip goose annotations first; they are themselves SQL comments.
+	body := string(raw)
+	parts := strings.SplitN(body, "-- +goose Down", 2)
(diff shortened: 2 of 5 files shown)


Quick104:
Verified the Codex P2: `newPlanstoreFixture` inserted a user without a role, so the DB-gated `TestPostgresPlanStore` would fail a not-null violation once this migration is applied. Fixed in 5bad72589 (the fixture now supplies `role = 'user'`); this was the only test-file user insert in the tree without a role.

Quick104:

PR #893, opened 2026-09-01
chore(db): convert serial ids to identity, widen notification varchars, rename tables [schema-hygiene 3/4]

Quick104:
## Problem

Before the 1.0 schema baseline freezes the shape of every table, three inconsistencies are still cheap to fix: 32 id columns use legacy `serial` defaults while 20 newer tables already use identity columns; the notifications/push tables carry 14 `varchar(n)` columns whose limits the application already enforces (and duplicates) in Go; and three tables break the schema's naming convention (`oauth_session`, `oauth_completion` singular in a plural schema; `playback_history_admin` puts the scope qualifier last where every sibling — `admin_jobs`, `admin_dashboard_layouts`, `user_watch_history` — puts it first).

**Base branch: `schema/hygiene-integrity` (PR #889)** — third PR in the schema-hygiene stack; only the top commit is new here.

## Approach

Three migrations, lists derived empirically from a migrated scratch database (audit queries in the migration comments):

1. **`20260901180714_serial_ids_to_identity.sql`** — converts the 32 serial `id` columns (36 with the four `activity_log` partitions that inherit) to `GENERATED BY DEFAULT AS IDENTITY`. `BY DEFAULT` keeps serial parity: explicit-id inserts (used by tests and id-carrying restores) keep working. One DO loop handles every table through a single code path: lock via `DROP DEFAULT`, read the old sequence position, drop it, add identity, `setval` the new sequence — so no table can restart at 1. The Down mirrors it (recreating the sequence `AS <column type>`) and takes an explicit `ACCESS EXCLUSIVE` lock before reading the position.
2. **`20260901180715_notification_varchar_to_text.sql`** — the 14 `varchar(n)` columns to `text`; metadata-only (relfilenodes verified unchanged). Every length rule survives in Go (`validateChannelName` ≤64, device_id ≤128, `left($n, 256)` on failure messages, etc.); no code depends on SQLSTATE 22001.
3. **`20260901180726_rename_tables_for_consistency.sql`** — `oauth_session → oauth_sessions`, `oauth_completion → oauth_completions`, `playback_history_admin → admin_playback_history`, with every index/constraint renamed and a catalog-driven sweep for the named NOT NULL constraints Postgres 18 leaves behind on `RENAME TABLE` (no-op on older servers). All Go, test, and doc references updated; repo-wide grep clean.

**Encryption safety of the renames** (checked before any rename stood): `secret.RowAAD` has no call site on these tables; `oauth_completion.token_ciphertext` derives its key from a fixed label and uses `code_hash` as AAD — no table identifier anywhere. Proven by a survival test: a token encrypted into `oauth_completion` decrypts through the real `GetAndDeleteCompletion` path after the rename.

**Frozen-v1 wire check**: the touched web widget and `docs/admin-api.md` edits are comment/prose only; no JSON field, query parameter, or TS type carried the old table name. No `/api/v1` wire change.

`migrations/schema_standardization_test.go` pins the 32-table list in both directions, the `BY DEFAULT` choice, the ordered continuity steps, the 14 columns with their restore lengths, the rename pairs with every index/constraint rename, and the NOT NULL sweep — mutation-tested (20 mutations, all caught).

## Validation

- Full chain from empty (pgvector pg18): 364 versions applied; `migrate-down-to 20260901161702` then re-up clean; each migration cycled down/up individually; `goose validate` clean.
- Sequence continuity: rows seeded across `users` (with a deleted-row gap), `media_files`, `activity_log`, `task_triggers`, `stream_nodes`, plus advanced-but-empty sequences and an uncalled one; the 46-sequence snapshot was byte-identical across Up and Down; post-Up/post-Down/re-Up inserts continued without collision. Independently reproduced by the destructive-safety reviewer.
- Encrypted-row survival across the rename: pass (above).
- `make embed-stub`, `go build ./...`, `gofmt` (empty), `go vet`: pass. `golangci-lint run --new-from-merge-base="origin/main"`: 0 issues.
- `make test-go`: all ok. Web: `pnpm run lint`, `format:check`, `build`, `make test-web`: pass.
- `make verify-settings-bindings-all`, `verify-playback-fixtures`, `verify-local-paths`: pass.
- Focused DB-gated runs with `SILO_TEST_DATABASE_URL` on the scratch DB: `internal/auth`, `internal/catalog/reattribute`, `internal/metadata`, `internal/api`, `internal/api/handlers`: pass.

## Risks

- The identity conversion takes `ACCESS EXCLUSIVE` per table for a catalog-only change (no rewrite); brief, but it is a maintenance-window migration like the rest of the stack.
- Renames are all-or-nothing within one transaction; the Down reverses them and the NOT NULL sweep.
- Two pre-existing failures noted, not introduced or fixed here: `internal/notifications` push-device tests (their TEMP shadow table lacks `fcm_token_ciphertext`) and `internal/catalog` `TestEpisodeSearchPostgresAndDocumentSource`. Both reproduce at the base commit.

## Checklist

## AI Disclosure

- Harness: Claude Code
- Model(s): claude-opus-5 (initial draft), claude-fable-5 (draft verification and completion, orchestration, all review agents)
- Adversarial review: four independent blind lenses (destructive-safety/encryption at maximum effort, migration discipline, rename-and-wire correctness, test adequacy). One major found and fixed: the Down read the sequence position before taking the table lock (fixed with an explicit lock). Two minors: cosmetic edits to two already-dead notification tests were reverted; a DB-gated continuity test was deferred in favor of the text-pinned contract test. Discipline and rename/wire lenses were clean.


Quick104:
Re the Codex P1 (preserve old table names through rolling upgrades): verified the mechanics — the first upgraded primary applies the rename at startup while any not-yet-upgraded replica keeps issuing SQL against the old names — and declining the compatibility-view/expand-contract approach with reasons, same as the drop migration in #879:

- Mixed-version API replicas during a migration have never been supported for any shape-changing migration in this chain (drops, NOT NULL, renames all fail the same way), and the documented posture is a verified-backup maintenance-window update (`docs/release-versioning.md` Backups and rollback; `docs/wiki/deployment/docker.md` Backups and updates), not a zero-downtime rolling upgrade of API replicas.
- The rename is a single transaction, so the switch is atomic: an old replica fails closed with `relation does not exist` rather than writing to a wrong table. OAuth sessions are short-lived and restart cleanly; playback history writes on an old replica error rather than being silently lost.
- Compatibility views would have to be carried into the squashed 1.0 baseline or removed by yet another migration inside the bridge, working against the point of this workstream (a clean schema before the bridge freeze).

If the maintainer prefers expand/contract for renames specifically, the alternative is to ship the Go-side dual-name support one release before the rename; happy to restructure. Recorded on #888.

Quick104:
Re the Codex P2 (rename the primary-key constraints, not only their indexes): verified and dismissed as incorrect for PostgreSQL. `ALTER INDEX ... RENAME` on an index that backs a UNIQUE / PRIMARY KEY / EXCLUDE constraint renames the constraint as well (PostgreSQL docs, ALTER INDEX, RENAME form). Reproduced on the deployed engine version:

```
PostgreSQL 18.6
CREATE TABLE t(id int PRIMARY KEY, u text UNIQUE);
ALTER INDEX t_pkey RENAME TO t2_pkey; ALTER INDEX t_u_key RENAME TO t2_u_key;
SELECT conname, conindid::regclass FROM pg_constraint WHERE conrelid='t'::regclass;
   conname  |  index
 t2_pkey    | t2_pkey
 t2_u_key   | t2_u_key
```

And on the scratch database after this migration ran, `pg_constraint` reports `oauth_sessions_pkey`, `oauth_completions_pkey`, and `admin_playback_history_pkey` as the constraint names. The catalog-driven sweep in the migration exists for the one thing `RENAME TABLE` and `ALTER INDEX RENAME` do leave behind on PG 18: the named NOT NULL constraints.

Quick104:

PR #901, opened 2026-09-02
fix(web): repair high contrast tokens dropped as cyclic CSS

zerodayz1:
## Problem

Related issue: N/A — narrow bug fix.

High contrast mode has never worked. Seven of the ten declarations in the
`html[data-high-contrast="true"]` block referenced **themselves**:

```css
--border: color-mix(in srgb, white 34%, var(--border));
```

Per [CSS Custom Properties L1 §3](https://www.w3.org/TR/css-variables-1/#cycles) a
custom property whose value references itself is in a dependency cycle and is
invalid at computed-value time, so the declaration is discarded and the token
computes to nothing. This holds **even though an inherited value exists** — the
inherited value is not consulted — so the intended mix never resolved.

Themes and this block both match `<html>` and both live in `@layer base`, so
`html[data-high-contrast="true"]` (0,1,1) wins over `[data-theme="…"]` (0,1,0) and
replaced each theme's good value with nothing.

Measured on a running instance (`midnight-cinema`, Chrome), reading computed values
off `document.documentElement` with high contrast on: all seven self-referencing
tokens computed **empty** (`--border --input --surface --surface-hover
--surface-raised --accent --muted-foreground`), while the non-self-referencing
`--foreground` and `--ambient-glow-opacity` applied correctly. The split falls
exactly along the self-reference line.

Two visible consequences, both making high contrast *worse* than standard mode:

1. **Panels lost their background.** `--surface` invalid, so `.surface-panel` /
   `.surface-panel-subtle` / `.surface-panel-raised` computed
   `background-color: rgba(0, 0, 0, 0)` and rendered transparent.
2. **Every border turned pure white.** `border-color` fell back to `currentColor`,
   and the block sets `--foreground: #ffffff`, so borders rendered
   `rgb(255, 255, 255)` instead of `rgb(40, 40, 46)`.

`git log -L` traces the block unchanged to c085b12f, so this never worked. It likely
went unreported because any profile with a stored `ui.custom_theme_vars` override
never sees it: those are injected unlayered by `CustomThemeProvider` and win over
this block.

**Second defect, found in review.** Every override mixed toward hard-coded white,
which raises contrast on a dark theme and destroys it on a light one. Computed
against cinema-light's `#f4f4f6` page:

| token | normal | high contrast |
|---|---|---|
| `--muted-foreground` | 5.42:1 | **1.36:1** |
| `--foreground` | — | **1.10:1** (`#ffffff` on `#f4f4f6`) |

The accessibility mode drove body text to near-invisible on that theme. The
`--foreground` half was already live before this PR, since it never self-referenced;
repairing the cycle would have activated the rest.

## Approach

**The cycle.** The mix needs an input that is not the property being assigned. Each
theme defines the literal under `--<token>-base`, the semantic token is an alias,
and high contrast mixes from the base.

Three shapes were tried in a browser; only the base token works:

```css
/* A — current shape */   html[data-x] { --border: color-mix(in srgb, white 34%, var(--border)); }
/* → (empty)  ✗ */
/* B — base token (this PR) */
html[data-x] { --border-base: #28282e; --border: color-mix(in srgb, white 34%, var(--border-base)); }
/* → color-mix(in srgb, white 34%, #28282e)  ✓ */
/* C — move to a descendant so var() inherits */
body[data-x] { --border: color-mix(in srgb, white 34%, var(--border)); }
/* → (empty)  ✗ still cyclic */
```

C is worth recording because it is the intuitive fix and does **not** work.

The alternative — dropping `color-mix` and hardcoding a lightened value per theme —
also works, but it is 7 × 5 hand-computed literals that drift as themes change.

**The direction.** Each theme declares `--contrast-boost` — white on the four dark
themes, black on cinema-light — and every override mixes toward that. On
cinema-light this yields 15.49:1 for muted text and 19.12:1 for body text. Dark
themes are unchanged except `--ring`, which moves from `oklch(0.95 0 0)` to the
boost: negligible there, and the difference between a visible and an invisible focus
ring on the light theme. `:root` defaults the boost to white so a theme that omits
it degrades to the previous behaviour rather than to an invalid value.

## Validation

```
$ npx tsc -b                    # clean
$ npx vite build                # ✓ built in 25.11s
$ npx eslint <changed files>    # clean
$ npx prettier --check <changed files>
Checking formatting...

$ npx vitest run
 Test Files  3 failed | 362 passed (365)
      Tests  5 failed | 3132 passed (3137)
```

The 5 failures are pre-existing and unrelated: `PersonDetail > formatBirthDate` ×3,
`SectionItemCard > renders premiere metadata`, `CardOverlays > scales legacy
browsers`. I reran those three files against a clean `main` and got the identical 5;
they look locale/timezone dependent.

Compiled output loaded into a real document root: all seven tokens resolve with high
contrast both off and on. No self-referencing declaration remains anywhere under
`web/src`.

Two suites, both mutation-checked rather than assumed:

- `highContrastCss.test.ts` — fails the whole stylesheet on any self-referencing
  custom property, checks the mix, the aliases, and that every theme defines every
  base token. Reintroducing the old declaration fails 2 of 4.
- `highContrastContrast.test.ts` — computed WCAG ratios per theme: the boost must
  contrast with its own page, body text must clear AA, and neither muted text nor
  borders may come out worse in high contrast than without it. Reverting
  cinema-light to boosting toward white fails 4 of 20. Asserting the declarations
  merely exist could never have caught the direction bug — the tokens were all
  present and correct-looking.

**UI evidence not attached.** The visible change is a theme-wide colour shift behind
an accessibility toggle; I can attach before/after captures per theme if that is
required before merge.

## Risks

- **Token rename.** `--<token>-base` is introduced for seven tokens across five
  themes. Anything outside `app.css` referencing those seven names directly would
  need updating; nothing does, and the semantic names are unchanged for consumers.
- **`--ring` on dark themes** shifts from `oklch(0.95 0 0)` to `#ffffff` in high
  contrast only. Called out above rather than buried.
- **Migration / security / operational:** none identified.

## Checklist

## AI Disclosure

- **Harness:** Claude Code
- **Tool(s):** Claude Code
- **Model(s):** claude-opus-5
- **Involvement:** AI-assisted — diagnosis, patch, tests and this description were
  AI-authored and human-verified. The bug was found while diagnosing an unrelated
  theming problem on a live instance.
- **Adversarial review:** The three candidate fixes were tested in a real browser
  rather than reasoned about, which is how the plausible-but-wrong descendant-element
  fix was rejected. Review findings on this PR — the light-theme contrast inversion
  and two `tsc` errors in the test added by the first commit — were reproduced and
  quantified against the code before being fixed in the second commit. I had not run
  `tsc -b` on this branch before opening the PR; `vite build` does not typecheck,
  which is why the errors were not caught earlier.

### DIFF
--- a/web/src/app.css
+++ b/web/src/app.css
@@ -457,6 +457,12 @@ html[data-navigation-direction]::view-transition-new(main-content) {
      It reads `--border-base` instead.
      ───────────────────────────────────────────────────────── */
   :root {
+    /* The direction high contrast pushes toward: away from the page. White on
+       the dark themes, black on a light one. Defaulting here means a theme
+       that forgets to set it degrades to the dark-theme behaviour rather than
+       to an invalid value. */
+    --contrast-boost: #ffffff;
+
     --muted-foreground: var(--muted-foreground-base);
     --border: var(--border-base);
     --input: var(--input-base);
@@ -476,6 +482,7 @@ html[data-navigation-direction]::view-transition-new(main-content) {
     --font-display: "Outfit", sans-serif;
 
     /* Surfaces: near-black with subtle cool undertone */
+    --contrast-boost: #ffffff;
     --background: #141417;
     --foreground: #e8e8ec;
     --card: #1c1c20;
@@ -540,6 +547,7 @@ html[data-navigation-direction]::view-transition-new(main-content) {
     --font-display: "Outfit", sans-serif;
 
     /* Surfaces: warm off-white with cool undertone */
+    --contrast-boost: #000000;
     --background: #f4f4f6;
     --foreground: #1a1a1e;
     --card: #ffffff;
@@ -609,6 +617,7 @@ html[data-navigation-direction]::view-transition-new(main-content) {
     --radius: 0.75rem;
     --font-body: "Outfit", sans-serif;
     --font-display: "Outfit", sans-serif;
+    --contrast-boost: #ffffff;
     --background: #101722;
     --foreground: #f4f8ff;
     --card: #151e2b;
@@ -654,6 +663,7 @@ html[data-navigation-direction]::view-transition-new(main-content) {
     --radius: 0.75rem;
     --font-body: "Outfit", sans-serif;
     --font-display: "Outfit", sans-serif;
+    --contrast-boost: #ffffff;
     --background: #171113;
     --foreground: #f8f2f3;
     --card: #20181b;
@@ -699,6 +709,7 @@ html[data-navigation-direction]::view-transition-new(main-content) {
     --radius: 0.75rem;
     --font-body: "Outfit", sans-serif;
     --font-display: "Outfit", sans-serif;
+    --contrast-boost: #ffffff;
     --background: #101715;
     --foreground: #f2f8f5;
     --card: #16201d;
@@ -774,16 +785,22 @@ html[data-navigation-direction]::view-transition-new(main-content) {
     --font-weight-semibold-ui: 600;
     --font-weight-bold-ui: 700;
   }
+  /* Every token here pushes toward --contrast-boost, never toward white
+     directly. Hard-coded white raises contrast on a dark theme and destroys it
+     on a light one: on cinema-light it drove --muted-foreground to 1.36:1 and
+     --foreground to 1.10:1 against the page, i.e. high contrast mode made the
+     theme unreadable. Mixing toward the boost gives 15.49:1 and 19.12:1 there
+     while leaving the dark themes as they were. */
   html[data-high-contrast="true"] {
-    --foreground: #ffffff;
-    --ring: oklch(0.95 0 0);
-    --muted-foreground: color-mix(in srgb, white 72%, var(--muted-foreground-base));
-    --border: color-mix(in srgb, white 34%, var(--border-base));
-    --input: color-mix(in srgb, white 16%, var(--input-base));
-    --surface: color-mix(in srgb, white 10%, var(--surface-base));
-    --surface-hover: color-mix(in srgb, white 16%, var(--surface-hover-base));
-    --surface-raised: color-mix(in srgb, white 20%, var(--surface-raised-base));
-    --accent: color-mix(in srgb, white 14%, var(--accent-base));
+    --foreground: var(--contrast-boost);
+    --ring: var(--contrast-boost);
+    --muted-foreground: color-mix(in srgb, var(--contrast-boost) 72%, var(--muted-foreground-base));
+    --border: color-mix(in srgb, var(--contrast-boost) 34%, var(--border-base));
+    --input: color-mix(in srgb, var(--contrast-boost) 16%, var(--input-base));
+    --surface: color-mix(in srgb, var(--contrast-boost) 10%, var(--surface-base));
+    --surface-hover: color-mix(in srgb, var(--contrast-boost) 16%, var(--surface-hover-base));
+    --surface-raised: color-mix(in srgb, var(--contrast-boost) 20%, var(--surface-raised-base));
+    --accent: color-mix(in srgb, var(--contrast-boost) 14%, var(--accent-base));
     --ambient-glow-opacity: 0.04;
   }
   html[data-high-contrast="true"] .surface-panel-subtle,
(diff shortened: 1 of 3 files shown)


zerodayz1:
Good catch — reproduced and fixed in fbd726a4, and it was worse than the comment
describes.

Computed against cinema-light's `#f4f4f6` page:

| token | normal | high contrast (before) |
|---|---|---|
| `--muted-foreground` | 5.42:1 | **1.36:1** |
| `--foreground` | — | **1.10:1** (`#ffffff` on `#f4f4f6`) |

So it was not only muted text and borders: `--foreground` itself was white on a
near-white page. That half was already live before this PR, since it never
self-referenced — repairing the cycle would have activated the rest and made a bad
situation considerably worse.

Rather than a cinema-light-specific override, each theme now declares
`--contrast-boost` (white on the four dark themes, black on cinema-light) and every
override mixes toward that. This generalises to any future light theme instead of
needing a parallel block per theme. On cinema-light it gives 15.49:1 for muted text
and 19.12:1 for body text. Dark themes are unchanged except `--ring`, which moves
from `oklch(0.95 0 0)` to the boost — negligible there, and the difference between a
visible and an invisible focus ring on the light theme. `:root` defaults the boost to
white, so a theme that omits it degrades to the previous behaviour rather than to an
invalid value.

Added the computed contrast checks you asked for: `highContrastContrast.test.ts`
walks every theme and asserts the boost contrasts with its own page, body text clears
AA, and neither muted text nor borders come out worse in high contrast than they were
without it. Reverting cinema-light to boosting toward white fails 4 of its 20 cases.
Worth noting that no assertion about the declarations existing could have caught
this — the tokens were all present and looked correct.

Also fixed in the same commit: two `tsc -b` errors in the test added by the first
commit, which meant `npm run build` was failing on this branch. `vite build` does not
typecheck, which is why I missed them. The new suite was written against the same
strict settings.

Validation after the fixes: `tsc -b`, `vite build`, `eslint` and `prettier --check`
clean; `vitest run` 3132 passed with 5 pre-existing failures (`PersonDetail >
formatBirthDate` ×3, `SectionItemCard`, `CardOverlays`) that reproduce identically on
a clean `main`.

I have also rewritten the PR description to follow `PULL_REQUEST_TEMPLATE.md`, which
I had missed entirely when opening this.

Quick104 (commented):
## Review summary: `fbd726a4`

No blocking findings. The fix is correct, the approach fits the project, and local validation passes. I pushed no changes.

**Verified**
- **The root cause.** In Chrome, `html[data-x] { --border: color-mix(in srgb, white 34%, var(--border)); }` computes to an empty value, as the description says.
- **The fix.** I built this head merged with current `main` and loaded the compiled CSS in Chrome. `--foreground`, `--muted-foreground`, `--border`, `--input`, `--surface`, `--surface-hover`, `--surface-raised`, `--accent`, and `--ring` resolve in all five themes, with high contrast both on and off. On cinema-light with high contrast, the page stays `#f4f4f6`, body text is `#000`, and muted text computes to about `#1c1c1e`. The dark themes lighten as before.
- **The alias placement.** `useTheme` only sets `data-theme` on `<html>`, so the `:root` aliases always resolve against the active theme's `-base` values. No subtree sets its own theme and inherits a stale alias.
- **Custom theme overrides.** `CustomThemeProvider` injects `ui.custom_theme_vars` as unlayered `:root` rules, and a user override of a semantic token such as `--surface` still wins in high contrast, as it did before. `TokenEditor` reads `--<token>` through `getComputedToken`, so the semantic names it uses are unchanged.
- **The earlier CodeRabbit finding** about light-theme direction is fixed by `--contrast-boost`.

**Validation** (this head merged with `main`, which merges cleanly)
- `make test-web`: 603 files and 4659 tests passed.
- `tsc -b`, `vite build`, `prettier --check`, and `eslint` on the changed files: clean.

**Optional, non-blocking**
- `highContrastContrast.test.ts` hard-codes the mix percentages (`mix(72, …)`, `mix(34, …)`) instead of reading them from `app.css`. Its doc comment says a changed mix percentage cannot slip through, but editing the CSS percentages would not fail this test. Parsing the percentages from the high-contrast block would make that claim true.
- In the same file, "pushes away from the page" and "keeps body text readable" make the same assertion (`>` vs `>=` 4.5).

**Merge condition:** the `CI` workflow for this fork PR is `action_required` and has not run. A maintainer must approve it, and it must pass.

**Mergeability: 8/10, conditional.** The fix is small, correct, and well explained, and the browser checks back up the description. The only gap is that the required CI has not run yet.

_Review performed with Claude Code (claude-opus-5-5)._

Issue #715, opened 2026-08-22
Improve poster card quick actions, watched controls, and overlays


blurbery:
## What I would like to improve

On the web client, poster cards make several common actions harder to reach than they need to be. Favouriting and watched state are available elsewhere, but they are not consistently visible or reversible from the card itself. Movie and series card menus also do not expose the existing Edit Metadata and Match Item workflows.

There are a few related interaction and layout problems:

- Closing the three-dot menu with the mouse can leave its trigger visible after the pointer leaves the poster.
- Carousel swipes can interfere with card action clicks.
- The card and detail menus are wider than their longest entry and do not consistently use the shared action icons.
- Standard hover controls overlap on the narrower Favourites and Watchlist posters.
- Episode watched indicators compete with thumbnail overlays and untouched episodes do not always expose a quick watched action.
- Bottom-left and bottom-right card overlays can be displaced by hover controls.
- Saved overlay defaults can appear one save behind because the browser keeps the shared configuration response cached.
- Windows hybrid-pointer browsers can report a non-hovering primary pointer even while an attached mouse is hovering, which can hide all card actions.

## Proposed design

- Add a bottom-left favourites heart and watched-state eye to movie and series posters, matching the existing glass-style three-dot control.
- Use a crossed white eye for unwatched items and a green eye for watched items. The heart turns red when active. Both use a short state-change animation.
- Keep favourite/unfavourite and watched/unwatched changes in sync with the card menu, item detail page, Favourites, Watchlist, recommendations, and other cached media surfaces.
- Make both actions optimistic and reversible without showing a loading cursor.
- Use the same pointer path as the three-dot control so taps remain reliable after horizontal carousel movement, while real drags remain swipes.
- Add the watched eye to Continue Watching without adding a favourites shortcut there.
- Add the watched eye to episode cards while keeping the compact watched check beside the episode number. Untouched episodes should work even when the API omits `user_data`.
- Use smaller hover controls on Favourites and Watchlist posters and keep all hover actions above configured card overlays.
- Add Edit Metadata and Match Item to movie and series card menus using the same permissions, item context, dialogs, and match results as the detail page.
- Use the shared Lucide action icons in both card and detail menus and size each menu to its longest entry.
- Support Windows hybrid-pointer browsers by basing hover visibility on any attached precise pointer and retaining touch and keyboard fallbacks.
- Preserve both pointer-release and browser-click activation so watched, favourite, and three-dot actions stay reliable across Windows browsers without double toggles.
[...]

PR #907, opened 2026-09-02
fix(web): reveal hover affordances on hybrid-pointer devices

zerodayz1:
## Problem

Related issue: N/A — narrow bug fix. Follow-up to the gate introduced in #715.

Card hover controls never appear on some Windows machines with a touchscreen, in
Chrome, Brave and Edge. The three-dot "More actions" and play triggers rest at
`opacity: 0; pointer-events: none` and are revealed by a rule behind
`@media (any-hover: hover) and (any-pointer: fine)`. On those machines Chromium
reports **`any-hover: none` and `any-pointer: coarse` while a mouse is attached and
actively hovering the card** — it has enumerated the touch digitizer and nothing
else — so the query never matches and the controls stay hidden permanently.

Measured on an affected machine, normal window, mouse hovering the page:

| feature | affected machine | ordinary desktop |
|---|---|---|
| `(hover: hover)` | `false` | `true` |
| `(any-hover: hover)` | **`false`** | `true` |
| `(pointer: fine)` | `false` | `true` |
| `(any-pointer: fine)` | **`false`** | `true` |
| `navigator.maxTouchPoints` | `10` | `0` |

The `any-*` features exist to describe *any* attached device, so there is nothing
more permissive to widen to: the answer itself is wrong. #715 already moved this
gate from `hover` to `any-hover` for this class of device, which was the right
instinct and is not enough.

Keyboard and open-menu reveals were never affected — those paths are unconditional,
which is why the component always tested fine. `Tab` + `Enter` opens the menu
correctly on the affected machine.

Tailwind v4 compiles every `hover:` and `group-hover:` utility into
`@media (hover: hover)`, which is false there too, so **all** hover styling in the
app is inert on these machines, not just the card controls.

## Approach

Observe the pointer rather than ask the OS about it. A `pointermove` carrying
`pointerType: "mouse"` is ground truth whatever the media query claims.

`lib/pointerCapability.ts` publishes `data-fine-pointer` on `<html>`: seeded
synchronously at boot from the same media query, then driven by observed
`pointermove` / `pointerdown` — mouse and pen set it, touch clears it, last pointer
wins. Behaviour is unchanged wherever the media query was already correct. Touch
devices keep clean artwork and the long-press sheet, because a touch press clears
the flag before Chromium's emulated `:hover` lands.

Three consumers follow it:

- the bespoke card reveal rules in `app.css`
- `@custom-variant hover`, so every Tailwind `hover:` / `group-hover:` utility is
  gated on the same signal instead of `@media (hover: hover)`
- `MediaItemMenu`, via the `useFinePointer` hook, since whether the watched and
  favorite shortcut trees are *mounted* is a JavaScript decision
- the `.pointer-reveal` row-action rule that `main` added in 2112f0752, merged
  into this branch so it follows the same gate

**The gate must contribute zero specificity.** Tailwind orders state variants by
specificity, so an unwrapped `:root[data-fine-pointer="true"]` prefix made `hover:`
utilities 0-4-0 against 0-2-0 for `disabled:`, `active:` and `focus-visible:` —
hover then won regardless of intended ordering, on every device. `:where()` fixes
this. That was caught in review; see the third commit.

Alternatives rejected: widening the media query further (nothing left to widen to);
bare `:hover` with no gate (reintroduces tap-emulated sticky hover on phones);
Tailwind `pointer-fine:` variants (compile to `@media (pointer: fine)`, also false
here, and blocked by the existing className assertions).

## Validation

```
$ npx tsc -b                    # clean
$ npx vite build                # ✓ built in 23.70s
$ npx eslint <changed files>    # 0 errors (1 pre-existing react-refresh warning
                                #   on MediaItemMenu.tsx, unrelated to this diff)
$ npx prettier --check <changed files>
Checking formatting...

$ npx vitest run
 Test Files  3 failed | 363 passed (366)
      Tests  5 failed | 3131 passed (3136)
```

The 5 failures are pre-existing and unrelated: `PersonDetail > formatBirthDate` ×3,
`SectionItemCard > renders premiere metadata`, `CardOverlays > scales legacy
browsers`. I reran those three files against a clean `main` and got the identical 5;
they look locale/timezone dependent.

Specificity verified on the production build:

```
$ grep -o ':root\[data-fine-pointer=true\]' index-*.css | wc -l          264
$ grep -o ':where(:root\[data-fine-pointer=true\])' index-*.css | wc -l  264
$ grep -o '@media *(hover:hover)' index-*.css | wc -l                      0
```

All 264 gates inside `:where()`, none leaking specificity. `hover:` utilities are
0-2-0, matching `active:` and `disabled:`. `group-hover:` composes correctly as
`.group-hover\:scale-105:is(:where(:root[data-fine-pointer=true]) :where(.group):hover *)`.

24 tests added across three suites, each mutation-checked rather than assumed:
reverting `pointerCapability` to media-query-only fails 5 of its 15; removing the
`@custom-variant` override fails the CSS guard; unwrapping `:where()` fails the
zero-specificity guard.

Verified on an affected machine against a running instance: the attribute seeds
`"false"` (the media query being wrong) and **a real mouse move flips it to `"true"`
while the query still reports `false`** — the exact path the fix depends on. The
reporter confirmed the controls then appear on hover.

**UI evidence not yet attached.** The visible change is a hover state on hardware
that reproduces the bug, which I cannot capture as a still from this environment. I
can attach a recording from the affected machine if that is required before merge.

## Risks

- **Scope.** The `@custom-variant` override changes how every `hover:` and
  `group-hover:` utility compiles — 264 rules. Behaviour is unchanged wherever the
  media query was already correct, since the attribute is seeded from that query,
  but the blast radius is app-wide and worth weighing. The first commit stands
  alone if you would rather take only the narrow card fix.
- **JS dependency.** Hover reveals now require `initPointerCapability` to have run.
  It is called before `createRoot(...).render(...)` and this is a React SPA, so
  there is no meaningful no-JS path to preserve. If it never ran, `readFinePointer`
  reports undefined and callers keep their prior defaults.
- **Migration / security / operational:** none identified.

## Checklist

## AI Disclosure

- **Harness:** Claude Code
- **Tool(s):** Claude Code
- **Model(s):** claude-opus-5
- **Involvement:** AI-assisted — the diagnosis, patch, tests and this description
  were AI-authored and human-verified. The original bug report came from a human
  reproducing it on their own machine, and a human confirmed the fix there.
- **Adversarial review:** The bug brief backing this change was produced in a
  separate session with browser access but no repository access; several of its
  stated file facts (line counts and line numbers) were wrong and were corrected
  against the source before implementation. Review findings on this PR — the
  specificity inflation, the unmigrated `MediaItemMenu` media-query read, and the
  media-query-overwrites-touch case — were each independently reproduced against
  the code and the compiled stylesheet before being fixed in the third commit,
  rather than applied on assertion.
- **Maintainer review follow-up:** Claude Code (claude-opus-5-5), run by a
  maintainer, merged current `main` into the branch. It moved `.pointer-reveal`
  onto the gate after the branch's guard test failed on the merge, and reworded
  two stale comments. Validation on the merged tree: the full web suite passed
  with the Makefile exclusions (4639 tests), and `tsc -b`, `eslint` and `vite build`
  were clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hover behavior on touchscreen and hybrid devices by detecting the pointer currently in use.
  * Prevented hover styles from overriding disabled, active, and keyboard focus states.
  * Ensured hover-dependent controls are correctly initialized before the first screen render.

* **Tests**
  * Added coverage for pointer detection, capability changes, cleanup, hover gating, and long-press interactions.

### DIFF
--- a/web/src/pointerGateCss.test.ts
+++ b/web/src/pointerGateCss.test.ts
@@ -28,14 +28,26 @@ describe("pointer-capability gating", () => {
   it("overrides Tailwind's hover variant onto the attribute", () => {
     // Without this, every `hover:` and `group-hover:` utility compiles into
     // `@media (hover: hover)` and is inert on the affected devices.
-    expect(css).toMatch(/@custom-variant hover \(:root\[data-fine-pointer="true"\] &:hover\);/);
+    expect(css).toMatch(
+      /@custom-variant hover \(:where\(:root\[data-fine-pointer="true"\]\) &:hover\);/,
+    );
+  });
+
+  it("keeps the gate at zero specificity everywhere it appears", () => {
+    // Tailwind orders its state variants by specificity. An unwrapped gate
+    // would make every hover: utility out-weigh disabled:, active: and
+    // focus-visible: on every device, not just the ones this works around.
+    const all = css.match(/:root\[data-fine-pointer="true"\]/g) ?? [];
+    const wrapped = css.match(/:where\(:root\[data-fine-pointer="true"\]\)/g) ?? [];
+    expect(all.length).toBeGreaterThan(0);
+    expect(wrapped).toHaveLength(all.length);
   });
 
   it("gates the card reveals on the attribute", () => {
     for (const selector of [
-      String.raw`:root\[data-fine-pointer="true"\] \.group\\/card:hover \.media-card-action-trigger`,
-      String.raw`:root\[data-fine-pointer="true"\] \.group\\/media:hover \.media-card-play-trigger`,
-      String.raw`:root\[data-fine-pointer="true"\] \.group\\/media:hover \.media-card-hover-dim`,
+      String.raw`:where\(:root\[data-fine-pointer="true"\]\) \.group\\/card:hover \.media-card-action-trigger`,
+      String.raw`:where\(:root\[data-fine-pointer="true"\]\) \.group\\/media:hover \.media-card-play-trigger`,
+      String.raw`:where\(:root\[data-fine-pointer="true"\]\) \.group\\/media:hover \.media-card-hover-dim`,
     ]) {
       expect(css).toMatch(new RegExp(selector));
     }
--- a/web/src/pointerGateCss.test.ts
+++ b/web/src/pointerGateCss.test.ts
@@ -1,0 +1,56 @@
+// @vitest-environment node
+
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+/**
+ * Hover reveals are gated in CSS on `data-fine-pointer`, which
+ * `lib/pointerCapability.ts` publishes from observed pointer events. These are
+ * contract tests over app.css, the same shape as
+ * `navigationTransitionCss.test.ts`.
+ *
+ * What matters: nothing here may fall back to a hover/pointer media feature.
+ * Chromium on some Windows machines with a touchscreen reports
+ * `any-hover: none` and `any-pointer: coarse` while a mouse is actively
+ * hovering, so any rule left behind such a query is silently dead on exactly
+ * the devices this gate exists to serve.
+ */
+const source = readFileSync(fileURLToPath(new URL("./app.css", import.meta.url)), "utf8");
+
+/**
+ * Comments here discuss the very media features these tests forbid, so the
+ * negative assertions below run against declarations only.
+ */
+const css = source.replace(/\/\*[\s\S]*?\*\//g, "");
+
+describe("pointer-capability gating", () => {
+  it("overrides Tailwind's hover variant onto the attribute", () => {
+    // Without this, every `hover:` and `group-hover:` utility compiles into
+    // `@media (hover: hover)` and is inert on the affected devices.
+    expect(css).toMatch(/@custom-variant hover \(:root\[data-fine-pointer="true"\] &:hover\);/);
+  });
+
+  it("gates the card reveals on the attribute", () => {
+    for (const selector of [
+      String.raw`:root\[data-fine-pointer="true"\] \.group\\/card:hover \.media-card-action-trigger`,
+      String.raw`:root\[data-fine-pointer="true"\] \.group\\/media:hover \.media-card-play-trigger`,
+      String.raw`:root\[data-fine-pointer="true"\] \.group\\/media:hover \.media-card-hover-dim`,
+    ]) {
+      expect(css).toMatch(new RegExp(selector));
+    }
+  });
+
+  it("leaves no hover or pointer media query gating a reveal", () => {
+    expect(css).not.toMatch(/@media[^{]*\bany-hover\b/);
+    expect(css).not.toMatch(/@media[^{]*\bany-pointer\b/);
+    expect(css).not.toMatch(/@media[^{]*\(\s*hover\s*:/);
+  });
+
+  it("never forces the controls back to hidden on a coarse pointer", () => {
+    // Such a rule would out-specify the `[data-state="open"]` and
+    // `:focus-visible` reveals and hide the trigger while its own menu is open
+    // on a touch device.
+    expect(css).not.toMatch(/\[data-fine-pointer="false"\]/);
+  });
+});
--- a/web/src/app.css
+++ b/web/src/app.css
@@ -1,5 +1,17 @@
 @import "tailwindcss";
 
+/* Tailwind v4 wraps every `hover:` and `group-hover:` utility in
+   `@media (hover: hover)`. That feature is false on Windows machines whose
+   Chromium under-reports pointer capability — the same devices the card
+   controls gate works around — so every hover style in the app is silently
+   inert there, not just the card controls.
+
+   Gate on the observed-pointer attribute instead (see lib/pointerCapability.ts,
+   which seeds it from the equivalent media query at boot). Behaviour is
+   unchanged wherever the media query was already right; touch devices still
+   get no hover styling, because a touch press clears the attribute. */
+@custom-variant hover (:root[data-fine-pointer="true"] &:hover);
+
 /* ================================================================
    § 1  THEME BRIDGE — Tailwind ↔ CSS Custom Properties
    ================================================================
--- a/web/src/app.css
+++ b/web/src/app.css
@@ -9,8 +9,15 @@
    Gate on the observed-pointer attribute instead (see lib/pointerCapability.ts,
    which seeds it from the equivalent media query at boot). Behaviour is
    unchanged wherever the media query was already right; touch devices still
-   get no hover styling, because a touch press clears the attribute. */
-@custom-variant hover (:root[data-fine-pointer="true"] &:hover);
+   get no hover styling, because a touch press clears the attribute.
+
+   The gate MUST stay inside :where(). Tailwind orders its state variants by
+   specificity, so an unwrapped `:root[data-fine-pointer="true"]` prefix would
+   make every hover: utility 0-4-0 against 0-2-0 for disabled:, active: and
+   focus-visible: — hover would then beat them on every device, not just the
+   ones this works around. :where() contributes nothing, leaving the compiled
+   selectors at exactly the specificity Tailwind intends. */
+@custom-variant hover (:where(:root[data-fine-pointer="true"]) &:hover);
 
 /* ================================================================
    § 1  THEME BRIDGE — Tailwind ↔ CSS Custom Properties
@@ -1641,11 +1648,11 @@ html[data-navigation-direction]::view-transition-new(main-content) {
      back to 0: at this specificity it would out-weigh the `[data-state="open"]`
      and `:focus-visible` rules above and hide the trigger while its own menu is
      open on a touch device. Rest state is the base rule's job. */
-  :root[data-fine-pointer="true"] .group\/media:hover .media-card-hover-dim {
+  :where(:root[data-fine-pointer="true"]) .group\/media:hover .media-card-hover-dim {
     opacity: 1;
   }
-  :root[data-fine-pointer="true"] .group\/card:hover .media-card-action-trigger,
-  :root[data-fine-pointer="true"] .group\/media:hover .media-card-play-trigger {
+  :where(:root[data-fine-pointer="true"]) .group\/card:hover .media-card-action-trigger,
+  :where(:root[data-fine-pointer="true"]) .group\/media:hover .media-card-play-trigger {
     opacity: 1;
     pointer-events: auto;
   }
--- a/web/src/app.css
+++ b/web/src/app.css
@@ -1617,15 +1617,25 @@ html[data-navigation-direction]::view-transition-new(main-content) {
   .group\/media:focus-within .media-card-hover-dim {
     opacity: 1;
   }
-  @media (any-hover: hover) and (any-pointer: fine) {
-    .group\/media:hover .media-card-hover-dim {
-      opacity: 1;
-    }
-    .group\/card:hover .media-card-action-trigger,
-    .group\/media:hover .media-card-play-trigger {
-      opacity: 1;
-      pointer-events: auto;
-    }
+  /* Gated on an attribute set from observed pointer events
+     (lib/pointerCapability.ts), not on a media query. Chromium on some Windows
+     hybrid machines reports `any-hover: none` and `any-pointer: coarse` even
+     while a mouse is actively hovering the card, so this block never matched
+     and the controls stayed at `opacity: 0` permanently. The attribute is
+     seeded from that same media query at boot, so behaviour is unchanged
+     wherever it was already right.
+
+     Deliberately no `[data-fine-pointer="false"]` counterpart forcing opacity
+     back to 0: at this specificity it would out-weigh the `[data-state="open"]`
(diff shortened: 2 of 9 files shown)


Quick104 (commented):
I found two high-confidence correctness issues that should be addressed before merge.

1. **[P2] Preserve Tailwind’s state-variant specificity**

   [`@custom-variant hover`](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/app.css#L13) adds both `:root` and `[data-fine-pointer]` to every generated `hover:` and `group-hover:` selector. The production build emits selectors such as `:root[data-fine-pointer=true] .hover\:opacity-90:hover` with specificity `0-4-0`, while `disabled:`, `active:`, `focus-visible:`, and data-state variants remain `0-2-0`.

   Hover therefore wins regardless of Tailwind’s intended variant ordering. Existing concrete regressions include the [disabled MarkerEditPanel control](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/player/components/MarkerEditPanel.tsx#L226) becoming 90% opaque while hovered, [card play controls](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/components/CardPlayOverlay.tsx#L73) remaining at `scale: 1.10` instead of applying `active:scale-95`, and the [intro-skip button](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/player/components/IntroSkipButton.tsx#L58) losing its pressed background under the pointer. This affects ordinary desktops too. The capability gate needs zero specificity, for example by wrapping it in `:where(...)`.

2. **[P2] Feed observed pointer state into React-mounted quick actions**

   The [pointer observer](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/lib/pointerCapability.ts#L57-L59) updates only `<html data-fine-pointer>`. `MediaItemMenu` still [reads the known-bad media query](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/components/MediaItemMenu.tsx#L81-L86) and [uses that result to decide whether watched/favorite shortcut trees are mounted](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/src/components/MediaItemMenu.tsx#L768-L783).

   On the exact Windows configuration this PR targets, a mouse move restores More Actions, play, and Tailwind hover styles, but the watched and favorite shortcuts remain absent. An isolated integration test reproduced this: after `data-fine-pointer` became `"true"` while `matchMedia` remained false, the DOM still exposed only More Actions. This also conflicts with the [documented media-card behavior](https://github.com/Silo-Server/silo-server/blob/3ff075ada1e71890620670f30e22b9e830c98ead/web/docs/design-system.md#L314).

I also reproduced the already-open CodeRabbit finding where a later media-query promotion overwrites an observed touch, so that thread remains valid.

Separate repository requirements remain unmet: the PR has no mandatory AI disclosure despite Claude co-authorship, its title is not Conventional Commit format, `Related issue: N/A` conflicts with the stated #715 follow-up, and the required UI evidence and complete pre-PR validation evidence are absent.

Validation: the exact-head production frontend build and all 15 added tests passed. The compiled stylesheet and isolated diagnostic tests exposed the two issues above.

zerodayz1:
Thanks — both findings reproduced, and both are fixed in 12b3f2ae.

**1. Variant specificity.** Confirmed exactly as described. The unwrapped gate made
`hover:` utilities 0-4-0 against 0-2-0 for `disabled:`, `active:` and
`focus-visible:`, so hover won regardless of Tailwind's ordering — on every device,
which made this a worse regression than the bug being fixed. The gate is now wrapped
in `:where()`. On the production build: 264 gate occurrences, all inside `:where()`,
`hover:` back to 0-2-0, and `group-hover:` still composing as
`.group-hover\:scale-105:is(:where(:root[data-fine-pointer=true]) :where(.group):hover *)`.
Added a guard asserting every gate occurrence is wrapped, so an unwrapped one fails
the suite rather than silently reordering variants again.

**2. React-mounted quick actions.** Also confirmed — I gated the CSS and missed that
`MediaItemMenu` reads the same query in JS to decide whether the shortcut trees are
mounted at all. The capability is now subscribable (`readFinePointer` /
`subscribeFinePointer`, surfaced as `useFinePointer` over `useSyncExternalStore`,
matching `useMediaQuery`'s shape) and `MediaItemMenu` consumes that. An absent
attribute reports undefined so callers keep their own default, preserving the
established quick actions in SSR, tests and browsers where init has not run. I
checked for other consumers of the query and there are none left.

The coarse-pointer long-press test now declares the device by publishing the
capability instead of stubbing matchMedia. Its contract is unchanged — desktop-only
shortcut trees omitted, long-press sheet intact — only the source of truth moved.

**3. CodeRabbit's media-query-overwrites-touch thread.** Valid, fixed in the same
commit: promotion now applies only until the first observed pointer, so a later
change event cannot undo an observed touch. Covered by two new tests.

**Repository requirements.** All fair, and my omissions:

- PR description rewritten to the template — Problem / Approach / Validation / Risks
  / Checklist / AI Disclosure. I had missed `PULL_REQUEST_TEMPLATE.md` entirely.
- Title corrected to Conventional Commit form.
- `Related issue: N/A` now reads as a narrow fix that is a follow-up to the gate
  introduced in #715, rather than implying #715 is a linked issue.
- Full validation output pasted, including the 5 pre-existing failures by name and
  the evidence that they reproduce on a clean `main`.
- UI evidence is the one item still outstanding: the visible change is a hover state
  on hardware that reproduces the bug, which I could not capture as a still. I can
  attach a recording from the affected machine if that is needed before merge.

Validation after the fixes: `tsc -b`, `vite build`, `eslint` and `prettier --check`
clean; `vitest run` 3131 passed, same 5 pre-existing failures. 24 tests now, each
mutation-checked — reverting `pointerCapability` to media-query-only fails 5 of 15,
removing the `@custom-variant` override fails the CSS guard, and unwrapping
`:where()` fails the new specificity guard.

zerodayz1:
Fixed in 22406882.

You are right that the default parameter defeats the fallback the docstring
promises: `readFinePointer(doc: Document = document)` evaluates its default at the
call site, so `readFinePointer()` in an environment without a document throws a
`ReferenceError` before it can return undefined.

Worth being straight about the reachability: this is a pure SPA with no server
rendering, and `useFinePointer` routes the server case through `getServerSnapshot`,
which never calls this — so nothing in the app can hit it today. But it is exported
API carrying a documented promise it would not keep, and narrowing the docstring
instead would be the worse fix. The document is now resolved inside the function,
guarded on `typeof`.

Added coverage that stubs the global away; restoring the default parameter fails it
with the ReferenceError, so the guard is doing real work rather than asserting
current behaviour.

`tsc -b`, `vite build`, `eslint` and `prettier --check` clean; `vitest run` 3132
passed with the same 5 pre-existing failures.

zerodayz1:
Independent reproduction on a second affected machine, since UI evidence was the one outstanding item.

Chromium on a touchscreen laptop, mouse attached and actively hovering a card. Read from the live page:

| feature | value |
| --- | --- |
| `(hover: hover)` | false |
| `(hover: none)` | true |
| `(pointer: fine)` | false |
| `(pointer: coarse)` | true |
| `(any-hover: hover)` | false |
| `(any-pointer: fine)` | false |
| `(any-pointer: coarse)` | true |
| `navigator.maxTouchPoints` | 10 |

That matches the table in the description exactly, including the `any-*` features reading false while a mouse is hovering the card.

DOM state captured at the moment of that hover:

```
card .group/media matches :hover     true
.media-card-action-trigger opacity   0
```

So the trigger is present in the DOM, the card is genuinely in `:hover`, and `@media (any-hover: hover) and (any-pointer: fine)` does not match — the controls stay at `opacity: 0` with no pointer-driven way to reveal them. Keyboard and open-menu reveals behave as described.

No screenshot attached: the still would just be a populated personal library, and it carries nothing the two readings above do not.

Separately re-verified against current `main` today: the branch rebases with no conflicts, the four touched test files pass 71/71, and the full web suite shows the same pre-existing failures as a clean checkout — I ran the four not listed in `WEBTEST_KNOWN_FAILURES` against an unmodified `origin/main` worktree and they fail identically there.

AI disclosure: measurements gathered and this comment drafted with Claude Code (claude-opus-5), run by the author against their own deployment.

web/src/pointerGateCss.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    for (const selector of [
+      String.raw`:where\(:root\[data-fine-pointer="true"\]\) \.group\\/card:hover \.media-card-action-trigger`,
+      String.raw`:where\(:root\[data-fine-pointer="true"\]\) \.group\\/media:hover \.media-card-play-trigger`,
+      String.raw`:where\(:root\[data-fine-pointer="true"\]\) \.group\\/media:hover \.media-card-hover-dim`,
+    ]) {
+      expect(css).toMatch(new RegExp(selector));
+    }
+  });
+
+  it("leaves no hover or pointer media query gating a reveal", () => {
```
Quick104, line 63: **[P2] This guard fails once the branch merges with current `main`.** `main` commit 2112f0752 (Sep 21) added a `.pointer-reveal` row-action rule in `app.css` behind `@media (any-hover: hover) and (any-pointer: fine)`. Git auto-merges `app.css` cleanly, but this test then fails with `expected ... not to match /@media[^{]*\bany-hover\b/`. I reproduced it by merging `origin/main` into this head and running the web suite: 1 failed, 4637 passed. Merging as-is would turn `make test-web` red on `main`.

It is also the same bug this PR fixes, with the polarity flipped. On the affected hybrids the query is false, so Watch Together row actions (`SuggestionList`, `SeriesDrilldown`) never hide for a mouse user.

Fix I'm pushing: merge `main`, then move `.pointer-reveal` onto `:where(:root[data-fine-pointer="true"])` at the same specificity it had inside the media query. I'm also adding a positive guard for the row reveal next to the card one.

Quick104 (commented):
Re-review of `22406882`. The earlier findings hold up as fixed: every compiled `hover:`/`group-hover:` gate sits inside `:where()` (specificity 0-2-0, same as `active:`/`disabled:`), `MediaItemMenu` follows the observed pointer, and an observed touch can no longer be overwritten by a media-query change. At this head, `tsc -b`, `vite build` and the focused suites pass.

Two new items, both from `main` moving since the branch point. I'm pushing the fix for both to this branch.

1. **[P2] Merge with current `main` fails the branch's own guard.** See the inline comment.
2. **[P3] Two comments now describe the gate this PR removed.** `web/src/app.css:1619-1623` says the card reveal uses a direct `:hover` because Tailwind's `group-hover` is limited to `(hover: hover)`, and `web/src/components/CardPlayOverlay.tsx:70-72` says the rule is "gated on any-hover/any-pointer". After this PR, neither is true. I'm rewording both to point at `lib/pointerCapability.ts`.

The CI workflow for this head is still `action_required` (fork approval), so required checks have not run on it yet.

_Review by Claude Code (claude-opus-5-5) on behalf of @Quick104._

web/src/pointerGateCss.test.ts
Quick104: Fixed in https://github.com/Silo-Server/silo-server/pull/907/commits/fec5e4e225af98e909f4d15c9642bff6771125d1, on top of a merge of current `main` (6223b87c).

- `.pointer-reveal` now hides and reveals under `:where(:root[data-fine-pointer="true"])`, at the same 0-1-0 / 0-2-0 / 0-3-0 specificity it had inside the media query. Touch and unpublished states keep the row actions visible, as before.
- New guard: `gates the row reveals on the attribute`.
- Also reworded the two stale comments from the review body (`app.css` card reveal, `CardPlayOverlay.tsx`).

Verified on the merged tree: full web suite with the Makefile exclusions, 602 files / 4639 tests passed; `tsc -b`, `eslint` and `prettier` clean; `vite build` emits 260 gate occurrences, all inside `:where()`, and no `any-hover` or `hover:` media queries.

Quick104:
**Review summary, head `fec5e4e2`**

Addressed:
- The P2 merge-skew guard failure: `.pointer-reveal` from `main` is now on the pointer gate (thread resolved).
- The P3 stale comments in `app.css` and `CardPlayOverlay.tsx`.
- Still fixed from the earlier round: `:where()` specificity, `MediaItemMenu` following the observed pointer, and observed touch no longer being overwritten.

Validation on the merged tree: web suite 602 files / 4639 tests passed with the Makefile exclusions; `tsc -b`, `eslint` and `vite build` clean. All 260 compiled gates sit inside `:where()`.

Remaining:
- CI for `fec5e4e2` is `action_required` and needs maintainer approval to run the fork workflow. Required checks have not run.
- Non-blocking and out of scope: the player's `useCoarsePointer` still reads `(pointer: coarse)`, which is true on the affected hybrids. Worth a follow-up if player controls act touch-first for mouse users there.

**Mergeability: 8/10 — conditional.** The approach is sound and the scope is right. It is ready once CI passes on `fec5e4e2`.

_Claude Code (claude-opus-5-5) on behalf of @Quick104._

Issue #909, opened 2026-09-02
Preserve viewer IP behind Cloudflare Tunnel


MrDoudou:
## Observed behavior
When Silo is accessed through a Cloudflare Tunnel running in Kubernetes, request logs record the cloudflared/Kubernetes peer address (for example, `10.42.0.169`) instead of the real viewer IP.

## Context
The deployment trusts the Kubernetes proxy ranges through `SILO_TRUSTED_PROXIES`. Cloudflare Tunnel forwards the viewer address in `CF-Connecting-IP`, but Silo currently resolves only `X-Forwarded-For` and `X-Real-IP`.

## Expected behavior
Use `CF-Connecting-IP` when the transport peer belongs to the configured trusted proxy ranges, while ignoring the header from untrusted peers.

## AI Disclosure

- Harness: Codex desktop app
- Tool(s): `exec_command`, `apply_patch`
- Model(s): GPT-5
- Involvement: AI-assisted
- Adversarial review: trusted-peer and untrusted-peer header cases were exercised in focused tests; no findings requiring changes

MrDoudou:
Closing after reproducing the request through the same Cloudflare Tunnel and Traefik route.

`CF-Connecting-IP` reaches the origin with the correct viewer address, while Traefik replaces `X-Forwarded-For` and `X-Real-IP` with the cloudflared pod IP because the entrypoint has no `forwardedHeaders.trustedIPs` configured.

The deployment fix is to trust the cloudflared pod CIDR on the relevant Traefik entrypoint. This lets Traefik preserve and extend the forwarding chain, which Silo already validates from right to left. The insecure forwarded-header mode is not required.

Thanks @Quick104 for the review and for steering this toward the correct trust-boundary fix.

PR #910, opened 2026-09-02
fix(clientip): honor Cloudflare Tunnel client IP

MrDoudou:
## Problem
Requests arriving through Cloudflare Tunnel are logged with the cloudflared/Kubernetes peer address (for example, `10.42.0.169`) instead of the viewer address.

## Solution
- accept Cloudflare Tunnel's `CF-Connecting-IP` header after the existing trusted-proxy boundary check
- keep untrusted peers from overriding the transport address with the header
- document the supported forwarding headers alongside `SILO_TRUSTED_PROXIES`

## Validation
- `go test ./internal/clientip` — passed
- `go test -race ./internal/clientip` — passed
- `go vet ./internal/clientip` — passed
- GitHub CI on the fork PR — Go, Web, and Docs hygiene passed

Related issue: #909

## AI Disclosure

- Harness: Codex desktop app
- Tool(s): `exec_command`, `apply_patch`
- Model(s): GPT-5
- Involvement: AI-assisted
- Adversarial review: exercised trusted-peer and untrusted-peer header cases in focused tests; no findings requiring changes

## Summary by CodeRabbit

* **New Features**
  * Trusted proxies can now provide client IP addresses through Cloudflare Tunnel’s `CF-Connecting-IP` header.

* **Documentation**
  * Updated the configuration example to document supported forwarded-IP headers for trusted peers.

* **Bug Fixes**
  * Untrusted peers’ `CF-Connecting-IP` values are ignored, preserving secure client IP detection.

### DIFF
--- a/internal/clientip/resolver.go
+++ b/internal/clientip/resolver.go
@@ -33,6 +33,14 @@ func (r *Resolver) ClientIP(req *http.Request) string {
 		return normalize(remoteIP)
 	}
 
+	// Cloudflare Tunnel preserves the viewer address in CF-Connecting-IP.
+	// This header is trusted only after the peer passed the proxy boundary above.
+	if cloudflareIP := strings.TrimSpace(req.Header.Get("CF-Connecting-IP")); cloudflareIP != "" {
+		if parsed := net.ParseIP(cloudflareIP); parsed != nil {
+			return normalize(parsed)
+		}
+	}
+
 	// Check X-Forwarded-For: walk right-to-left, return first untrusted IP.
 	if xff := req.Header.Get("X-Forwarded-For"); xff != "" {
 		ips := strings.Split(xff, ",")
--- a/.env.example
+++ b/.env.example
@@ -83,6 +83,8 @@ POSTGRES_SHM_SIZE=8gb
 # validated and written to server_settings, so the Admin UI shows it). Leave
 # unset to manage it from Admin > Settings > General instead. Defaults trust
 # RFC 1918 private ranges + loopback.
+# When the peer is trusted, Silo accepts X-Forwarded-For, X-Real-IP, and
+# Cloudflare Tunnel's CF-Connecting-IP forwarding headers.
 # SILO_TRUSTED_PROXIES=10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,127.0.0.0/8,::1/128,203.0.113.7/32
 
 # Run from source / advanced overrides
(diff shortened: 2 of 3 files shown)


internal/clientip/resolver.go
```diff
@@ -33,6 +33,14 @@ func (r *Resolver) ClientIP(req *http.Request) string {
 		return normalize(remoteIP)
 	}
 
+	// Cloudflare Tunnel preserves the viewer address in CF-Connecting-IP.
+	// This header is trusted only after the peer passed the proxy boundary above.
+	if cloudflareIP := strings.TrimSpace(req.Header.Get("CF-Connecting-IP")); cloudflareIP != "" {
```
Quick104, line 38: [P1] Do not let `CF-Connecting-IP` override the validated forwarding chain

This trusts `CF-Connecting-IP` from every peer in `SILO_TRUSTED_PROXIES`, not specifically from Cloudflare. In an existing nginx, Traefik, or similar deployment, the trusted proxy can add a trustworthy `X-Forwarded-For` entry while passing an otherwise unknown client-supplied `CF-Connecting-IP` through unchanged. This early return then selects the forged value before the right-to-left XFF validation.

For example, with trusted peer `10.42.0.5`, `X-Forwarded-For: 203.0.113.50`, and attacker-supplied `CF-Connecting-IP: 198.51.100.42`, the base resolves `203.0.113.50`; this change resolves `198.51.100.42`. Rotating the header can bypass the native per-IP authentication limits and falsify request, activity, playback, and session IP attribution.

Please treat this header as a fallback after the validated standard headers, or require a separate Cloudflare-header trust opt-in/source restriction.

MrDoudou:
Thanks @Quick104 for the review and for questioning the trust boundary.

I confirmed the behavior with a whoami service behind the same Cloudflare Tunnel and Traefik instance:

- `CF-Connecting-IP` contains the correct public client IPv6.
- `X-Forwarded-For` and `X-Real-IP` are replaced with the cloudflared pod IP (`10.42.0.162`).
- A client-supplied `X-Forwarded-For` produces the same origin headers.
- A client attempt to supply `CF-Connecting-IP` is rejected by Cloudflare.

The Traefik entrypoint does not currently configure `forwardedHeaders.trustedIPs`, so it does not trust cloudflared and rebuilds the standard forwarding headers. The appropriate fix is to trust the cloudflared pod CIDR on the Traefik entrypoint, without enabling the insecure forwarded-header mode. Once Traefik preserves the chain, Silo's existing right-to-left XFF validation resolves the viewer correctly.

This is therefore a deployment configuration issue rather than something Silo should work around. I am closing this PR and will apply the fix in the Kubernetes/Traefik configuration. Thank you again for catching the unsafe generic behavior.

PR #919, opened 2026-09-02
fix(playback): make GPU transcoding adaptive, bounded, and observable

### LINKED ISSUE #918: [bug] auto hardware transcode does not retry CPU decode with GPU encode
## What happened

On a real Silo deployment with an NVIDIA Tesla P4 and Jellyfin FFmpeg 7.1.4, automatic hardware transcoding selected a complete hardware pipeline for a source that the GPU could encode but could not decode through the selected hardware path.

FFmpeg exited before producing the first playable manifest. Playback then failed instead of trying CPU decoding while retaining the working GPU encoder.

The client-visible failures observed during these sessions included:

~~~text
All compatible playback recipes have already failed for this output route
~~~

and:

~~~text
The playback transport failed before media became ready
~~~

The same source shape was successfully processed on the same server when decoding on the CPU, converting/uploading frames as NV12, scaling with CUDA, and encoding with NVENC. This shows that the GPU encoder remained usable even though the full hardware path did not.

## Steps to reproduce

1. Run Silo with a supported NVIDIA GPU and Jellyfin FFmpeg 7.1.4.
2. Configure Playback → Hardware acceleration as <code>Auto</code>.
3. Select an H.264 High 10 1080p source that is not accepted by the available NVIDIA hardware decoder.
4. Request a lower-resolution H.264 transcode, such as 720p.
5. Observe that the automatically selected full-hardware FFmpeg process exits before creating its first manifest.
6. Observe that playback fails without retrying the viable CPU-decode → GPU-encode path.
7. Run the same media shape using CPU decode, NV12 CUDA upload, CUDA scaling, and NVENC output.
8. Observe that FFmpeg produces playable output on the same GPU.

## Expected behavior

When <code>playback.hw_accel=auto</code>, a failure of the complete hardware path should not immediately discard a working hardware encoder.

Silo should attempt progressively safer startup paths:

1. GPU decode → GPU encode.
2. CPU decode → GPU encode.
3. CPU decode → CPU encode.

Fallback should happen only after FFmpeg exits before producing the first manifest. A process that is still running but slow must not cause a duplicate encoder to start.

Explicit hardware selections, direct play, stream copy/remux, and validated HDR/tone-map recipes should keep their existing behavior.

## Silo version / commit

<code>90b5c0f4c44fcccebefa2822098638d6a2fc099a</code>

## Deployment

Docker container deployed on Kubernetes.

## Clients affected

- [x] Web UI
- [x] Android
- [x] iOS/tvOS/macOS
- [x] Jellyfin-compat client
- [x] Other/not client-specific

The failure is in FFmpeg startup and can affect any client route that requests the incompatible automatic transcode.

## Relevant logs

~~~shell
No raw server logs are available from the pre-fix reproduction. The original pod was replaced before this issue was filed.
~~~

The client-visible messages above are copied from the reproduced playback failures; they are not presented as server log lines.

## Technical notes

Automatic selection currently resolves one hardware backend and treats that result as valid for the complete decode/filter/encode pipeline. Hardware support is not uniform across decode codec, profile, bit depth, pixel format, filters, driver, FFmpeg build, GPU generation, and encoder.

A bounded automatic fallback should:

- use the real manifest startup as validation instead of launching a separate probe;
- close a failed process before starting the next path;
- retain GPU encoding when only hardware decoding fails;
- cache only a path that actually produced a manifest;
- scope any cache entry to the media and executable recipe;
- expire cached fallback choices so a recovered GPU is retried;
- apply consistently to local, remote-node, Jellyfin-compatible, and reconstructed playback.

The tested implementation uses a 15-minute, 256-entry, concurrency-safe successful-path cache and avoids the concrete device that just failed when another configured hardware device is available.

## AI harness

Codex desktop app

## AI tool(s)

Codex <code>exec_command</code>, <code>apply_patch</code>, and in-app Browser

## AI model(s)

GPT-5 (Codex)

## AI involvement

AI-assisted

## Independent or adversarial review

The complete implementation diff was reviewed across automatic eligibility, startup process lifecycle, media-specific cache isolation, cache concurrency and bounds, multi-device avoidance, explicit acceleration behavior, tone-map exclusions, local/remote/Jellyfin/reconstruction parity, throttle defaults and hysteresis, segment cleanup, FFmpeg log volume, session correlation, and Admin rendering.

Review findings were resolved by restricting adaptive fallback to ordinary <code>auto</code> video transcodes, refusing to retry a process that remains alive, writing cache preferences only after manifest readiness, bounding cache lifetime and size, preserving specialized tone-map paths, propagating resource policy through durable recipes, emitting full diagnostics only on lifecycle events, and moving log-sink calls outside session locks.

## Confirmations

- [x] I reproduced this myself on a real deployment.
- [x] No synthesized or AI-summarized server logs are presented as raw evidence.
- [x] The absence of retained pre-fix server logs is stated explicitly.

MrDoudou:
## Problem

Related issue: #918

Automatic hardware selection can abort playback when the selected GPU cannot decode a source even though the same GPU can encode CPU-decoded frames. Transcodes can also run far ahead of the viewer, and the Activity view does not expose enough of the effective FFmpeg recipe to diagnose the selected path.

Fixes #918

## Approach

This change makes automatic FFmpeg hardware transcoding adaptive instead of treating one resolved hardware path as universally compatible.

When <code>playback.hw_accel=auto</code>, Silo now validates the real playback startup and progressively falls back through safer execution paths:

1. GPU decode → GPU encode.
2. CPU decode → GPU encode.
3. CPU decode → CPU encode.

The first path that actually produces a playable manifest is used. This keeps GPU encoding available when a source cannot be decoded or filtered safely on the GPU, while still preserving a final software path for machines or media combinations that cannot use hardware acceleration.

The same behavior is applied consistently to native playback, Jellyfin-compatible playback, local execution, remote transcode nodes, and reconstructed sessions.

The PR also bounds work and temporary storage by enabling transcoding throttling and short segment retention by default, then adds a complete FFmpeg diagnostics panel to Admin → Activity so operators can see exactly which path and recipe were selected.

## Why this is needed

The previous automatic selection resolved one FFmpeg/hardware backend and assumed that the complete pipeline would work for every compatible-looking source.

That assumption is too broad. Hardware support varies independently across:

- decode codec;
- source profile and bit depth;
- pixel format;
- filters and scale path;
- driver;
- FFmpeg build;
- GPU generation;
- output encoder;
- remote-node hardware.

A GPU can often encode the requested output even when it cannot decode the source. For example, rejecting a hardware decode should not immediately discard a working NVENC, QSV, VA-API, or VideoToolbox encoder.

Before this change, such a startup failure could surface to the client as a transport error even though this valid path existed:

~~~text
source → CPU decode → compatible CPU pixel format → GPU upload/filter → GPU encode
~~~

Using full software transcoding immediately would restore playback, but would unnecessarily discard the GPU and create significantly more CPU load. A fixed, GPU-specific workaround would also not be portable across backends and deployments.

The adaptive sequence keeps the fast path first, preserves GPU encoding for the common partial-compatibility case, and only uses full software when both hardware paths fail.

## Scope

This change covers:

- automatic hardware fallback;
- bounded transcode generation and segment retention;
- local/remote/Jellyfin/reconstruction propagation;
- FFmpeg operational diagnostics;
- Admin Playback warnings and Admin Activity diagnostics;
- related tests.

### Eligibility

The adaptive pipeline is intentionally narrow. It is enabled only when all of the following are true:

- the administrator selected <code>auto</code>;
- the request is a video transcode;
- automatic resolution selected a real hardware backend;
- the video target is not stream copy;
- the request is not using a frozen HDR/tone-map recipe.

Explicit administrator selections such as <code>nvenc</code>, <code>qsv</code>, <code>vaapi</code>, <code>videotoolbox</code>, or <code>none</code> keep their existing behavior.

Direct play, remux/video copy, and the existing validated tone-map execution paths are also left unchanged.

### Fallback order

For an eligible request, the server builds the following ordered candidates:

| Stage | Decode | Encode | Purpose |
| --- | --- | --- | --- |
| Full hardware | GPU | GPU | Preferred low-CPU path |
| Mixed | CPU | GPU | Keeps hardware encoding when hardware decode is incompatible |
| Software | CPU | CPU | Last-resort compatibility path |

If source-safety logic has already required software decoding, the sequence begins at the mixed path and then falls back to full software.

### Startup validation

No separate FFmpeg probe or duplicate pre-encode is introduced.

The actual playback FFmpeg process is the validation:

- Silo starts the current candidate.
- It waits for the first playable manifest.
- If the manifest is ready, that candidate is accepted.
- If FFmpeg exited before producing the manifest, the failed process is closed and the next candidate is attempted.
- If FFmpeg is still running but merely slow, Silo does not start a duplicate process and reports the normal readiness timeout.

This distinction prevents multiple encoders from processing the same media concurrently.

An eligible request can make at most two additional startup attempts beyond the preferred full-hardware path.

### Multiple GPU devices

When a concrete hardware device fails during startup, the next hardware-capable attempt records that device as one to avoid.

The existing allocator can therefore prefer another configured device when one is available. The avoidance hint is cleared for the final software path.

This avoids repeatedly selecting the same failing device without permanently disabling it.

## Successful-path cache

A small in-memory cache prevents every playback of the same problematic source from repeating known failed startup paths.

A fallback is cached only after the real FFmpeg process produced a valid manifest. Failed attempts and timeouts are never cached as successes.

The signature includes the byte- and executor-relevant shape:

- FFmpeg executable identity;
- source file path;
- resolved hardware backend;
- configured hardware device set;
- source codec;
- source profile;
- source bit depth;
- source resolution;
- target video codec;
- target resolution;
- target bitrate;
- subtitle burn-in state;
- subtitle codec.

This is deliberately media-specific. One corrupt, unusual, or unsupported file cannot downgrade unrelated titles that happen to use the same codec.

Cache bounds:

- TTL: 15 minutes;
- maximum entries: 256;
- expired entries are removed during access/update;
- the oldest entry is evicted when the bound is reached;
- access is concurrency-safe;
- a later successful full-hardware start clears the previous fallback preference.

The short TTL avoids a permanent software downgrade after a driver, FFmpeg, device, or deployment recovery.

## Playback paths covered

The adaptive startup and resource policy are wired through all relevant execution paths:

- native local playback;
- native remote transcode-node playback;
- Jellyfin-compatible local playback;
- Jellyfin-compatible remote playback;
- local session reconstruction;
- transcode-node session reconstruction;
- restart hooks after seek/audio-track changes.

For remote execution, the API server sends the original playback session identity separately from the node transport identity. FFmpeg logs therefore remain correlated with the user-visible Activity session instead of being split under an internal transport ID.

The source resolution and resolved throttle policy are carried through:

- transcode-node start requests;
- recipe cards;
- stream-token claims;
- reconstructed <code>TranscodeOpts</code>.

This keeps live starts and post-restart reconstruction behavior equivalent.

### Forward work: throttling

Transcoding throttling is now enabled by default.

Default policy:

- enabled: <code>true</code>;
- forward buffer: 120 seconds;
- accepted configured range: 60–86,400 seconds.

FFmpeg pauses when generated media is at least the configured threshold ahead of the client.

It resumes only when the lead falls to half the threshold or below. For the 120-second default, this means:

- pause at 120 seconds ahead;
- resume at 60 seconds ahead.

The half-threshold hysteresis prevents a fast encoder from repeatedly pausing and resuming around one boundary.

The throttler is reattached after FFmpeg restarts and reconstructed sessions, so seek and audio-track changes retain the same resource policy.

### Backward cache: segment retention

The default HLS segment retention changes from 600 seconds to 120 seconds.

Accepted values:

- <code>0</code>: disable segment cleanup explicitly;
- 120–86,400 seconds: retain a bounded rewind window.

Downloaded segments older than the configured rewind window are pruned during playback. The complete temporary transcode directory is still session-scoped and removed when playback ends, so a later viewing regenerates the media instead of retaining a permanent converted copy.

Together, the two bounds control both sides of the client position:

~~~text
pruned old segments ← 120 s back buffer | viewer | 120 s forward buffer → FFmpeg paused
~~~

This prevents an abandoned playback from encoding an entire title at full speed and consuming unnecessary CPU, GPU time, and temporary disk space.

### Existing installations and explicit opt-outs

The settings remain administrator-controlled:

- an existing explicit throttle value is preserved;
- an existing explicit disabled throttle remains disabled;
- <code>playback.segment_retention_seconds=0</code> remains a valid cleanup opt-out;
- missing, invalid, or temporarily unavailable throttle settings fall back to the protective defaults.

No persistent media copy or long-lived conversion cache is added.

## Admin Playback UI

The Playback settings explain the new behavior and make unsafe resource choices visible.

Changes include:

- the Hardware acceleration description now explains the <code>auto</code> fallback behavior;
- enabled throttling displays “Recommended protection enabled”;
- disabled throttling displays a strong CPU/GPU/disk warning;
- the main Transcoding row warns when transcoding is enabled without throttling;
- Buffer ahead recommends 120 seconds;
- Transcode back buffer explains the rewind window and session cleanup lifecycle;
- disabling segment cleanup displays a temporary-storage growth warning.

The UI does not add another tone-map toggle. The existing hardware and software HDR tone-map settings retain their separate purpose and behavior.

### Event diagnostics

FFmpeg lifecycle events now include an event-only diagnostic snapshot derived from the exact argument vector passed to the process.

Snapshots are recorded for meaningful lifecycle points such as:

- process starting;
- process started;
- process restart;
- process exit;
- process exit error;
- resource-policy configuration.

The diagnostic payload is intentionally not repeated on every FFmpeg stderr line. This keeps the operational log volume bounded while retaining the complete recipe needed for debugging.

Throttle pause/resume events are lightweight and contain the current paused state and lead in seconds without duplicating the full FFmpeg command.

### Captured fields

The diagnostics include:

- FFmpeg executable path;
- exact FFmpeg argument vector;
- effective execution pipeline;
- node type, node ID, and execution mode;
- resolved hardware backend and concrete device;
- video and audio encoders;
- complete video filter graph;
- explicit pixel-format transitions;
- source codec, profile, bit depth, resolution, and audio channels;
- target codecs, resolution, video bitrate, audio channels, and audio bitrate;
- audio/subtitle track indexes;
- subtitle burn-in and subtitle codec;
- HDR/tone-map policy, mode, source kind, filter, recipe version, and preflight requirement;
- video bitstream filter and sample entry;
- HLS segment type and duration;
- segment retention;
- total media duration;
- fast-start and seek state;
- segment generation;
- last requested and delivered segments;
- restart count;
- throttle configured/enabled/paused state;
- throttle threshold and current lead;
- input and output paths;
- process exit error where applicable.

The pipeline is reported with an operator-readable distinction:

- GPU decode → GPU encode;
- CPU decode → GPU encode;
- CPU decode → CPU encode;
- video copy + audio transcode;
- remux / stream copy.

### Log correlation and safety

Native and Jellyfin-compatible paths now use the same FFmpeg log sink, including integrated and remote nodes.

Logs prefer the original playback session ID for correlation. Internal transport IDs remain available for execution but no longer fragment the Activity view.

Sink calls are made after releasing the transcode-session mutex. This avoids blocking process/session state while the logging backend writes an event.

Existing stderr line and byte bounds remain in place. Once the cap is reached, one capped event records the dropped-line count rather than growing operational storage without limit.

## Admin Activity UI

Expanding a transcoding session in Admin → Activity now displays a dedicated **Transcode diagnostics** panel before the live stderr console.

The panel groups facts into:

- Execution;
- Video;
- Audio & subtitles;
- HDR / tone map;
- HLS & cache;
- Paths.

It also provides two expandable views:

- the exact reconstructed FFmpeg command with argument boundaries safely quoted;
- the complete raw diagnostic attributes as formatted JSON.

The model merges newest-first diagnostic events and fills missing values from the live session summary. This allows partial/older events to remain useful while always preferring the newest concrete runtime facts.

## Implementation notes

The main implementation units are:

- <code>internal/playback/auto_transcode_pipeline.go</code>: candidate ordering and successful-path cache;
- <code>internal/playback/transcode_startup.go</code>: first-manifest startup validation and retries;
- <code>internal/playback/ffmpeg_diagnostics.go</code>: exact recipe/pipeline extraction;
- <code>internal/config/transcode_resource_policy.go</code>: shared live throttle policy;
- <code>internal/playback/transcode.go</code>: session lifecycle, diagnostics, throttle state, and device avoidance;
- <code>internal/playback/throttle.go</code>: pause/resume hysteresis and transition events;
- <code>internal/playback/recipecard.go</code> and <code>internal/streamtoken/token.go</code>: reconstruction persistence;
- native API, Jellyfin compatibility, and transcode-node handlers: consistent propagation;
- <code>web/src/pages/adminActivityTranscodeDebug.ts</code>: presentation model and formatting;
- <code>web/src/pages/AdminActivity.tsx</code>: diagnostics panel;
- <code>web/src/pages/admin-settings/PlaybackSettings.tsx</code>: behavior descriptions and warnings.

### Unsupported GPU decode, working GPU encode

~~~text
Attempt 1: NVDEC/QSV/VA-API/VideoToolbox decode + GPU encode
Result: FFmpeg exits before manifest

Attempt 2: CPU decode + GPU encode
Result: first manifest produced

Playback: starts with GPU encoding
Cache: mixed path preferred for this exact signature for 15 minutes
~~~

### Hardware unavailable for the complete request

~~~text
Attempt 1: full hardware fails before manifest
Attempt 2: CPU decode + GPU encode fails before manifest
Attempt 3: CPU decode + CPU encode produces manifest

Playback: starts in software
Cache: software path preferred briefly for this exact signature
~~~

### Slow but still-running FFmpeg

~~~text
Attempt 1: process remains alive but does not become ready before timeout
Result: readiness error; no duplicate fallback process is started
~~~

### Explicit hardware selection

~~~text
Configuration: playback.hw_accel=nvenc
Result: existing NVENC behavior; adaptive auto fallback is not activated
~~~

## Compatibility and non-goals

- No database migration is required.
- Existing setting keys are reused.
- Existing explicit administrator choices remain authoritative.
- No permanent transcoded-media cache is introduced.
- No per-media probe process is introduced.
- No background conversion of complete titles is introduced.
- Validated HDR/tone-map recipes continue through their existing specialized fallback logic.

## Performance impact

Steady-state overhead is negligible:

- candidate construction and cache lookup are in-memory operations;
- the cache is bounded to 256 small entries;
- no extra FFmpeg process is launched when the preferred path succeeds;
- no media is copied or pre-generated for detection;
- diagnostic parsing scans the already-built FFmpeg argument list only on lifecycle events.

Additional CPU work occurs only when compatibility requires CPU decoding or full software transcoding.

The first unsupported signature can incur one or two failed startup attempts. Later matching sessions can start from the last manifest-producing path during the 15-minute cache lifetime.

Throttling reduces total unnecessary resource use for viewers who stop early by preventing FFmpeg from racing to encode the entire title.

### Go tests

Passed locally:

~~~text
go test ./internal/config -count=1
go test ./internal/api/handlers -count=1
go test ./internal/jellycompat -count=1
go test ./internal/transcodenode -count=1
~~~

Focused playback coverage passed for:

- automatic stage ordering;
- eligibility boundaries;
- cache success/expiry/eviction/concurrency behavior;
- failed-device avoidance;
- manifest-ready success;
- startup exit fallback;
- no fallback for a still-running process;
- diagnostic field extraction;
- event-only diagnostic logging;
- recipe-card and stream-token propagation;
- throttle policy defaults;
- pause/resume hysteresis;
- segment progress and retention behavior.

The changed playback tests and <code>go vet</code> checks pass. The complete playback package still has pre-existing macOS GPU-probe timing failures caused by hard 200 ms process timeouts ending in <code>signal: killed</code>; these failures are unrelated to the changed paths.

### Static checks

Passed locally:

~~~text
go vet ./internal/config ./internal/playback ./internal/transcodenode ./internal/jellycompat ./internal/api/handlers ./cmd/silo
tsc -b --pretty false
prettier --check <changed web files>
eslint src/pages/AdminActivity.tsx src/pages/adminActivityTranscodeDebug.ts src/pages/adminActivityTranscodeDebug.test.ts
vite build
git diff --check
~~~

### Frontend tests

The three affected frontend test files pass: 49 tests total.

Coverage includes:

- diagnostic event merging;
- session fallbacks;
- FFmpeg command reconstruction and quoting;
- field formatting;
- readable pipeline labels;
- Playback settings warnings;
- throttle recommendation state;
- disabled segment-cleanup warning.

### Hardware validation

A real Tesla P4 using Jellyfin FFmpeg 7.1.4 successfully ran the intended mixed path:

~~~text
H.264 High 10 1080p
→ CPU decode
→ NV12 CUDA upload
→ scale_cuda to 720p
→ NVENC encode
~~~

This demonstrates the central use case: preserve the GPU encoder even when the hardware decode path cannot accept the source.

## Risks

| Risk | Mitigation |
| --- | --- |
| Longer first startup for an unsupported signature | At most two safer retries; successful fallback cached for 15 minutes |
| CPU increase on the mixed path | CPU performs decode only; GPU still filters/encodes where supported |
| High CPU on final software path | Used only after hardware paths fail |
| A transient GPU failure downgrades future sessions | Cache is media-specific and expires after 15 minutes |
| Cache growth | Hard limit of 256 entries with expiry and oldest-entry eviction |
| Duplicate FFmpeg work | Fallback only after the previous process exited before manifest |
| Temporary disk growth | 120-second forward throttle and 120-second back retention by default |
| Pause/resume oscillation | Resume threshold is half the pause threshold |
| Excess operational-log volume | Full diagnostics only on lifecycle events; stderr remains capped |
| Behavior drift after restart | Resource policy and source facts persist through requests/tokens/recipe cards |
| Impact to explicit recipes | Auto-only eligibility excludes explicit acceleration, copy, and frozen tone-map paths |

## Reviewer guide

The most important properties to verify are:

- automatic fallback activates only for <code>auto</code>;
- a running process is never duplicated because it is slow;
- each failed FFmpeg process is closed before the next attempt;
- a cached fallback is written only after manifest readiness;
- the cache key cannot downgrade unrelated media;
- explicit acceleration and validated tone-map recipes are unchanged;
- local, remote, Jellyfin, and reconstruction paths carry equivalent options;
- throttle settings and source resolution survive node/reconstruction boundaries;
- detailed snapshots are event-only, not attached to every stderr line;
- the Admin panel renders partial and legacy data safely.

## Checklist

- [x] Automatic hardware fallback is generic across supported backends.
- [x] GPU encode remains available when only hardware decode fails.
- [x] Full software remains the final compatibility path.
- [x] Cache lifetime and size are bounded.
- [x] Local, remote, Jellyfin, and reconstructed playback are covered.
- [x] Throttling and segment retention bound forward/backward work.
- [x] Explicit administrator opt-outs are preserved.
- [x] FFmpeg diagnostics are correlated with Activity sessions.
- [x] Admin settings explain and strongly recommend resource protection.
- [x] Backend, frontend, static, formatting, and hardware checks were performed.

## AI disclosure

- Harness: Codex desktop app
- Tooling: Codex <code>exec_command</code> and <code>apply_patch</code>
- Model family: GPT-5 Codex
- Involvement: AI-assisted implementation, tests, review, and PR documentation
- Adversarial review: examined fallback boundaries, startup process lifecycle, cache isolation/concurrency, device selection, tone-map exclusions, reconstruction parity, throttle defaults/hysteresis, segment retention, logging volume, session correlation, and Admin rendering fallbacks.

## Summary by CodeRabbit

* **New Features**
  * Automatic transcoding now retries safer hardware and software paths when playback startup fails, including improved GPU fallback handling.
  * Added detailed FFmpeg and transcoding diagnostics to Admin Activity, including the exact command, video/audio settings, HDR processing, caching, and throttling status.
  * Playback sessions now preserve transcoding resolution and resource-management settings across stream tokens and remote nodes.

* **Improvements**
  * Added clearer Playback settings guidance for throttling, buffer limits, hardware acceleration fallback, and segment cleanup.
  * Improved transcoding throttling stability to reduce unnecessary pause/resume cycling.
  * Updated default playback segment retention to 120 seconds.

### DIFF
--- a/internal/config/admin_settings.go
+++ b/internal/config/admin_settings.go
@@ -15,7 +15,6 @@ import (
 
 const (
 	cloudflareURLMode                  = "cloudflare_token"
-	playbackSegmentRetentionSettingKey = "playback.segment_retention_seconds"
 	chapterThumbnailSoftwareToneMapKey = "playback.chapter_thumbnail_software_tone_map_enabled"
 )
 
@@ -87,7 +86,7 @@ var adminSettingDefaults = map[string]string{
 
 	"playback.ffmpeg_path":                           "",
 	playbackTranscodeDirSettingKey:                   DefaultTranscodeDir,
-	playbackSegmentRetentionSettingKey:               "600",
+	PlaybackSegmentRetentionSettingKey:               strconv.Itoa(DefaultPlaybackSegmentRetentionSeconds),
 	"playback.hw_accel":                              "auto",
 	"playback.transcode_enabled":                     "true",
 	PlaybackRoutingDirectPlayEgressSettingKey:        string(PlaybackEgressPreferProxy),
@@ -105,8 +104,8 @@ var adminSettingDefaults = map[string]string{
 	"playback.watched_threshold":                     "90",
 	"playback.min_resume_threshold":                  "5",
 	Allow4KTranscodeSettingKey:                       "false",
-	"enable_transcode_throttle":                      "false",
-	"transcode_throttle_seconds":                     "300",
+	TranscodeThrottleEnabledSettingKey:               strconv.FormatBool(DefaultTranscodeThrottleEnabled),
+	TranscodeThrottleSecondsSettingKey:               strconv.Itoa(DefaultTranscodeThrottleSeconds),
 
 	"audiobookshelf_compat.enabled":           "true",
 	"jellyfin_compat.enabled":                 "true",
@@ -324,7 +323,7 @@ func NormalizeAdminSetting(key, raw string) (string, error) {
 	case "metadata.cache_images", "playback.transcode_enabled",
 		chapterThumbnailSoftwareToneMapKey, PlaybackTranscodeHardwareToneMapSettingKey,
 		PlaybackTranscodeSoftwareToneMapSettingKey,
-		Allow4KTranscodeSettingKey, "enable_transcode_throttle", "audiobookshelf_compat.enabled",
+		Allow4KTranscodeSettingKey, TranscodeThrottleEnabledSettingKey, "audiobookshelf_compat.enabled",
 		"jellyfin_compat.enabled", "jellyfin_compat.web_enabled", "recommendations.enabled",
 		"subtitle_ai.enabled", "subtitle_ai.transcribe_enabled", "metadata_ai.enabled",
 		"download.enabled", "download.transcode_enabled", DownloadLocalTranscodeFallbackSettingKey,
@@ -357,9 +356,9 @@ func NormalizeAdminSetting(key, raw string) (string, error) {
 		return normalizeAdminInt(key, value, 1, 100)
 	case "playback.min_resume_threshold":
 		return normalizeAdminInt(key, value, 1, 99)
-	case "transcode_throttle_seconds":
+	case TranscodeThrottleSecondsSettingKey:
 		return normalizeAdminInt(key, value, 60, 86400)
-	case playbackSegmentRetentionSettingKey:
+	case PlaybackSegmentRetentionSettingKey:
 		normalized, err := normalizeAdminInt(key, value, 0, 86400)
 		if err != nil {
 			return "", err
--- a/internal/playback/transcode.go
+++ b/internal/playback/transcode.go
@@ -40,14 +40,16 @@ type TranscodeOpts struct {
 	subtitleFilterInputPath string
 	// OutputSubdir is the signed, root-relative reconstruction directory. Empty
 	// retains the legacy flat {session_id} layout.
-	OutputSubdir         string
-	TranscodeTransportID string
-	SessionID            string
-	SourceVideoCodec     string
-	SourceVideoProfile   string
-	SourceVideoBitDepth  int
-	VideoBitstreamFilter string // validated copy-mode BSF, e.g. dovi_rpu=strip=1
-	VideoSampleEntry     string // allowlisted copy-HLS sample entry: dvh1 or hvc1
+	OutputSubdir          string
+	TranscodeTransportID  string
+	SessionID             string
+	PlaybackSessionID     string
+	SourceVideoCodec      string
+	SourceVideoProfile    string
+	SourceVideoBitDepth   int
+	SourceVideoResolution string
+	VideoBitstreamFilter  string // validated copy-mode BSF, e.g. dovi_rpu=strip=1
+	VideoSampleEntry      string // allowlisted copy-HLS sample entry: dvh1 or hvc1
 	// CopyVideoMPEGTS packages copied video in MPEG-TS instead of fMP4. It is
 	// durable because the segment extension and bytes must survive restarts.
 	CopyVideoMPEGTS bool
@@ -59,26 +61,28 @@ type TranscodeOpts struct {
 	StreamOriginSeconds float64
 	// CopySeekAnchorResolved distinguishes a valid zero-second origin from
 	// older/shared recipes that never resolved a copy seek anchor.
-	CopySeekAnchorResolved  bool
-	TargetResolution        string // e.g., 1080p, 720p
-	TargetCodecVideo        string // e.g., h264 (or hevc if allowed)
-	TargetCodecAudio        string // e.g., aac
-	SegmentDuration         int    // seconds, default 6
-	SegmentRetentionSeconds int    // downloaded media retained behind the client; 0 disables pruning
-	StartSegmentNumber      int    // -hls_segment_start_number, default 0
-	FFmpegPath              string // optional explicit ffmpeg binary path
-	HWAccel                 string // auto, qsv, vaapi, nvenc, videotoolbox, none
-	HWDevice                string // e.g., /dev/dri/renderD128 (default if empty)
+	CopySeekAnchorResolved   bool
+	TargetResolution         string // e.g., 1080p, 720p
+	TargetCodecVideo         string // e.g., h264 (or hevc if allowed)
+	TargetCodecAudio         string // e.g., aac
+	SegmentDuration          int    // seconds, default 6
+	SegmentRetentionSeconds  int    // downloaded media retained behind the client; 0 disables pruning
+	ThrottlePolicyConfigured bool
+	ThrottleEnabled          bool
+	ThrottleThresholdSeconds int
+	StartSegmentNumber       int    // -hls_segment_start_number, default 0
+	FFmpegPath               string // optional explicit ffmpeg binary path
+	HWAccel                  string // auto, qsv, vaapi, nvenc, videotoolbox, none
+	HWDevice                 string // e.g., /dev/dri/renderD128 (default if empty)
 	// AvoidHWDevice asks the initial multi-device allocator to prefer any other
 	// present render device. It is a process-local startup hint used after an
 	// early GPU failure; the selected concrete device remains fully reserved and
 	// is the value frozen into the session recipe.
 	AvoidHWDevice string
 	// SoftwareVideoDecode keeps a hardware encoder while decoding the source
-	// on the CPU. Intel's VAAPI/QSV decoders cannot accept 10-bit AVC, but the
-	// decoded frames can still be converted to NV12, uploaded, and encoded by
-	// QSV/VAAPI. The flag is frozen into recipe cards so restarts do not put the
-	// unsupported hardware decoder back.
+	// on the CPU. Unsupported decoder inputs can still be converted to NV12,
+	// uploaded, and encoded by QSV, VAAPI, NVENC, or VideoToolbox. The flag is
+	// frozen into recipe cards so restarts do not restore the failed decoder.
 	SoftwareVideoDecode        bool
 	ToneMapPolicy              tonemap.Policy
 	ToneMapMode                tonemap.Mode
@@ -199,6 +203,11 @@ type TranscodeSession struct {
 	restartCount         int
 	stderrLineIndex      int
 	stderrWriter         *ffmpegStderrWriter
+	ffmpegArgs           []string
+	throttleConfigured   bool
+	throttleEnabled      bool
+	throttleThreshold    int
+	throttlePaused       bool
 	restartHook          func(context.Context)
 	// generationStartedAt is when the currently-owning ffmpeg process was
 	// spawned. Output in the shared directory older than this timestamp was
@@ -369,6 +378,7 @@ func StartTranscode(ctx context.Context, opts TranscodeOpts) (*TranscodeSession,
 	}
 
 	args := buildFFmpegArgs(opts)
+	s.ffmpegArgs = append([]string(nil), args...)
 	bin := opts.FFmpegPath
 	if bin == "" {
 		bin = ffmpegBinary()
@@ -853,11 +863,10 @@ func resolveEffectiveTranscodeHWAccelContext(ctx context.Context, opts Transcode
 			return transcodeHWNone
 		}
 	}
-	// The bundled CUDA software-decode upload path has not been validated.
-	// Prefer the established libx264 fallback over selecting a decoder known
-	// not to accept this source. Intel QSV/VAAPI have the explicit upload paths
-	// below and retain hardware encoding.
-	if opts.SoftwareVideoDecode && hwAccel == transcodeHWNVENC {
+	// Hardware tone-map recipes are capability-validated as a complete decode,
+	// filter, and encode graph. Do not splice the ordinary CUDA upload path into
+	// that frozen graph; its owner must replan a complete software recipe.
+	if opts.SoftwareVideoDecode && hwAccel == transcodeHWNVENC && opts.ToneMapMode != "" {
 		return HWAccelNone
 	}
 	return hwAccel
@@ -1009,6 +1018,12 @@ func appendHWAccelArgs(args []string, opts TranscodeOpts) []string {
 			args = append(args, "-hwaccel", "vaapi", "-hwaccel_output_format", "vaapi")
 		}
 	case transcodeHWNVENC:
+		if opts.SoftwareVideoDecode {
+			if hwDevice := strings.TrimSpace(opts.HWDevice); hwDevice != "" {
+				args = append(args, "-init_hw_device", "cuda=cu:"+hwDevice, "-filter_hw_device", "cu")
+			}
+			return append(args, "-noautorotate")
+		}
 		args = append(args,
 			"-hwaccel", "cuda",
 			"-hwaccel_output_format", "cuda",
@@ -1206,6 +1221,8 @@ func appendVideoFilterArgs(args []string, opts TranscodeOpts) []string {
 		return append(args, "-vf", qsvSoftwareDecodeFilter(opts.TargetResolution))
 	case opts.HWAccel == "vaapi" && opts.SoftwareVideoDecode:
 		return append(args, "-vf", vaapiSoftwareDecodeFilter(opts.TargetResolution))
+	case opts.HWAccel == transcodeHWNVENC && opts.SoftwareVideoDecode:
+		return append(args, "-vf", nvencSoftwareDecodeFilter(opts.TargetResolution))
 	case opts.HWAccel == "qsv":
 		return append(args, "-vf", qsvScaleFilter(opts.TargetResolution))
 	case opts.HWAccel == "vaapi":
@@ -1570,9 +1587,14 @@ func appendBitmapSubtitleBurnInArgs(args []string, opts TranscodeOpts) []string 
 			cpuFilters += "," + scale
 		}
 		if opts.HWAccel == transcodeHWNVENC {
-			// Download to CPU for the overlay, then re-upload to CUDA.
-			graph = "[0:v:0]hwdownload,format=yuv420p[vmain];[vmain]" + cpuFilters +
-				",format=nv12,hwupload_cuda[vout]"
+			if opts.SoftwareVideoDecode {
+				graph = "[0:v:0]format=yuv420p[vmain];[vmain]" + cpuFilters +
+					",format=nv12,hwupload_cuda[vout]"
+			} else {
+				// Download to CPU for the overlay, then re-upload to CUDA.
+				graph = "[0:v:0]hwdownload,format=yuv420p[vmain];[vmain]" + cpuFilters +
+					",format=nv12,hwupload_cuda[vout]"
+			}
 		} else {
 			// CPU encoding: overlay directly on decoded frames.
 			graph = "[0:v:0]" + cpuFilters + "[vout]"
@@ -1641,8 +1663,13 @@ func appendSubtitleBurnInArgs(args []string, opts TranscodeOpts) []string {
 		vf := "hwdownload,format=yuv420p," + cpuFilters + ",format=nv12,hwupload"
 		args = append(args, "-vf", vf)
 	case transcodeHWNVENC:
-		// NVENC/CUDA: download to CPU for subtitle rendering, then upload back.
-		vf := "hwdownload,format=yuv420p," + cpuFilters + ",format=nv12,hwupload_cuda"
+		// NVENC/CUDA renders subtitles on the CPU and uploads the result. A
+		// software-decoded source is already in CPU memory and needs no download.
+		prefix := "hwdownload,format=yuv420p,"
+		if opts.SoftwareVideoDecode {
+			prefix = "format=yuv420p,"
+		}
+		vf := prefix + cpuFilters + ",format=nv12,hwupload_cuda"
 		args = append(args, "-vf", vf)
 	default:
 		// CPU encoding: filters run directly on decoded frames.
@@ -1766,6 +1793,13 @@ func nvencScaleFilter(res string) string {
 	default:
 		return "scale_cuda=format=nv12"
 	}
+}
+
+// nvencSoftwareDecodeFilter uploads CPU-decoded NV12 frames before CUDA scale
+// and NVENC encode, avoiding hardware decoder limitations without losing GPU
+// acceleration for the expensive output stages.
+func nvencSoftwareDecodeFilter(res string) string {
+	return "format=nv12,hwupload_cuda," + nvencScaleFilter(res)
 }
 
 // filterPathReplacer escapes special characters in file paths for ffmpeg filter syntax.
@@ -2966,7 +3000,7 @@ func (s *TranscodeSession) restart(
 	}
 
 	log.Printf("playback: ffmpeg restart cmd: %s %s", bin, strings.Join(args, " "))
-	s.logFFmpegEvent(ctx, "ffmpeg process restart", "")
+	s.logFFmpegAttemptEvent(ctx, opts, args, "ffmpeg process restart", "")
 
 	// As on initial start, the caller bounds synchronous validation; the new
 	// FFmpeg process keeps running after the segment request completes.
@@ -3006,7 +3040,7 @@ func (s *TranscodeSession) restart(
 		s.waitErr = err
 		s.mu.Unlock()
 		close(flight.done)
-		s.logFFmpegEvent(ctx, "ffmpeg process exit error", err.Error())
+		s.logFFmpegAttemptEvent(ctx, opts, args, "ffmpeg process exit error", err.Error())
 		return fmt.Errorf("restart ffmpeg: %w", err)
 	}
 
@@ -3017,6 +3051,7 @@ func (s *TranscodeSession) restart(
 	s.cmd = cmd
 	s.cancel = cancel
 	s.opts = opts
+	s.ffmpegArgs = append(s.ffmpegArgs[:0], args...)
 	s.running = true
 	s.restarting = nil
 	s.stdinPipe = stdinPipe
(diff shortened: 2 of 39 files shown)


MrDoudou:
## UI screenshots

### Automatic hardware behavior

The Playback settings explain that Auto keeps GPU encoding with CPU decoding before using full software.

Automatic hardware fallback settings

### Resource protection

The advanced settings show the recommended throttle state, 120-second forward buffer, and 120-second rewind retention.

Transcode resource protection settings

### Activity diagnostics

The expanded Activity row exposes the effective FFmpeg recipe, execution path, and resource policy.

Transcode diagnostics panel

The detailed view separates execution, video, audio/subtitles, HDR/tone-map, HLS/cache, and path information.

Detailed FFmpeg diagnostics

web/src/pages/AdminActivity.tsx
```diff
@@ -578,15 +582,31 @@ function StreamRow({
   );
   const logsHref = `/admin/logs?playback_session_id=${encodeURIComponent(session.session_id)}&focus=playback`;
   const ffmpegLogsHref = `${logsHref}&component=ffmpeg`;
-  const ffmpegLogs = useOperationalLogs(
+  const ffmpegDiagnostics = useOperationalLogs(
```
Quick104, line 596: **Medium: the FFmpeg console no longer shows stderr.** Before this change, the console under an Activity row rendered the last 12 `component=ffmpeg` rows. Decoder and filter errors show up there. `ffmpegRows` now comes from two `limit: 1` queries filtered to `ffmpeg diagnostic snapshot` and `ffmpeg throttle`, and the same array feeds both the new panel and the console. The console therefore shows at most those two events. An operator looking at a failing transcode can no longer see the error lines without leaving the page.

Fix: keep the original 12-row query for the console and give the diagnostics panel its own rows from the two event queries. I'm pushing that change to this branch.

Quick104 (commented):
Review of `d66f32e`. The core fallback is a real improvement: the CPU-decode → GPU-encode path fixes #918, and the design holds up (fallback only after FFmpeg exits, a bounded per-file cache, no duplicate encoders for a slow start). Four things need attention before merge. Two of them I'm fixing on this branch.

**Blocking outside the inline comments**

- **Merge conflicts.** The branch conflicts with current `main` in 8 files (`playback.go`, `playback_v3.go`, `admin_settings.go`, `jellycompat/handlers_playback.go`, `jellycompat/streams.go`, `recipecard_test.go`, `transcodenode/server.go`, `PlaybackSettings.tsx`). `main` is 340 commits past this branch's base, including the API v2 migration (#1075) and #1076 (node spawn-failure handling). This needs a rebase. My fix commits sit on top of the current head so they rebase with the rest.
- **The description understates one behavior change.** It says explicit `hw_accel=nvenc` keeps its old behavior. But removing the `SoftwareVideoDecode && NVENC → none` rule in `resolveEffectiveTranscodeHWAccelContext` also affects explicit NVENC. H.264 High 10 sources that used libx264 before now take the new CUDA upload path, and explicit mode has no software fallback if that path fails on a given driver. That may be the right call, but the description should say so.

Inline findings, most severe first:
1. The Activity FFmpeg console lost its stderr tail. I'm fixing it.
2. Local and node reconstruct regressed for `hw_accel=auto`, including CPU-only hosts. I'm fixing it.
3. The CPU-decode NVENC path ignores the configured GPU. This needs NVIDIA hardware to validate, so I've left it for the author.
4. Throttle and retention default changes are outside #918's scope. This needs a maintainer decision.

This review was produced with Claude Code (Claude Opus 5.5), and I checked each finding against the source.

internal/playback/transcode_manager.go
```diff
@@ -889,7 +889,13 @@ func (m *TranscodeManager) doReconstructTranscode(ctx context.Context, sessionID
 	if startTranscode == nil {
 		startTranscode = StartTranscode
 	}
-	transcodeSession, err := startTranscode(ctx, opts)
+	var transcodeSession *TranscodeSession
+	var err error
+	if strings.EqualFold(strings.TrimSpace(opts.HWAccel), hwAccelAuto) {
```
Quick104: **Medium: local reconstruct changes behavior on every `hw_accel=auto` host.** `auto` is the default, and this check runs before the pipeline decides whether fallback applies. On a CPU-only host, or anywhere auto resolves to software, every reconstruct now waits up to `ManifestStartupTimeout` for the manifest while holding the session lifecycle lock. If FFmpeg is still running at the deadline, `startReadyTranscodePipeline` kills it and the rebuild fails. Before this PR, the reconstruct registered the process immediately and the segment handler waited for it. Reconstructs also use `FastStart=false`, which needs three segments before readiness. That makes them the starts most likely to be slow, especially after a restart when many sessions rebuild at once. A failed attempt also goes through `Close()`, which deletes the output directory that still holds segments the client may request.

Fix: run the fallback loop only when `AutoTranscodePipeline.Enabled()`. For a reconstruct, keep a process that is still running, and stop a failed attempt without deleting the directory the next attempt writes into. I'm pushing this with tests.

internal/transcodenode/server.go
```diff
@@ -1790,7 +1796,13 @@ func (s *Server) spawnReconstruct(r *http.Request, sessionID string, requestedSe
 		opts.SeekSeconds = float64(requestedSegment * card.SegmentDuration)
 	}
 
-	session, err := playback.StartTranscode(r.Context(), opts)
+	autoRequested := strings.EqualFold(strings.TrimSpace(opts.HWAccel), "auto")
```
Quick104: **Medium: node reconstruct now deletes its output directory and kills slow processes.** The block this replaces used `CloseProcess()` on purpose ("Keep the shared output directory: the retry writes into it") and kept a slow reconstruct running. With the node's default `auto`, the new path calls `StartReadyTranscode`, which uses `Close()`. That removes the session directory and kills any process that hasn't produced a manifest within the timeout. It also applies when auto resolves to software, where no fallback exists.

The fix described on `transcode_manager.go` covers this path too. The fallback runs only when the pipeline is enabled, and the existing VideoToolbox retry handles every other case as before.

internal/playback/transcode.go
```diff
@@ -1009,6 +1018,12 @@ func appendHWAccelArgs(args []string, opts TranscodeOpts) []string {
 			args = append(args, "-hwaccel", "vaapi", "-hwaccel_output_format", "vaapi")
 		}
 	case transcodeHWNVENC:
+		if opts.SoftwareVideoDecode {
+			if hwDevice := strings.TrimSpace(opts.HWDevice); hwDevice != "" {
+				args = append(args, "-init_hw_device", "cuda=cu:"+hwDevice, "-filter_hw_device", "cu")
```
Quick104, line 1023: **Low: the CPU-decode NVENC path ignores the configured GPU.** `-init_hw_device cuda=cu:<dev> -filter_hw_device cu` has no effect on `hwupload_cuda`. In FFmpeg's `vf_hwupload_cuda.c`, `cudaupload_init` always creates its own CUDA context from the filter's `device` option, which defaults to 0. With `playback.hw_device=1`, the mixed path uploads, scales, and encodes on GPU 0 while the allocator counts the work against device 1. The two subtitle burn-in variants for this path (`format=nv12,hwupload_cuda`) have the same problem.

Possible fixes: use the generic `hwupload` filter, which does honor `-filter_hw_device`, or pass `hwupload_cuda=device=<index>`. I've left this one for you because it changes the FFmpeg graph you validated on the P4, and I don't have an NVIDIA host to check it on.

internal/config/admin_settings.go
```diff
@@ -105,8 +104,8 @@ var adminSettingDefaults = map[string]string{
 	"playback.watched_threshold":                     "90",
 	"playback.min_resume_threshold":                  "5",
 	Allow4KTranscodeSettingKey:                       "false",
-	"enable_transcode_throttle":                      "false",
-	"transcode_throttle_seconds":                     "300",
+	TranscodeThrottleEnabledSettingKey:               strconv.FormatBool(DefaultTranscodeThrottleEnabled),
```
Quick104, line 107: **Needs a maintainer decision: default changes outside #918.** This turns transcode throttling on for every install that never saved `enable_transcode_throttle`, and it adds throttling to jellycompat for the first time. It also lowers the default `playback.segment_retention_seconds` from 600 to 120, so seeking back more than two minutes now restarts FFmpeg instead of serving retained segments.

One more effect: the throttler pauses FFmpeg by writing `p` and `u` to stdin. Only jellyfin-ffmpeg handles those keys. Stock FFmpeg's `check_keyboard_interaction` ignores them (for example Homebrew builds on macOS), so on those hosts the throttle does nothing, yet the new diagnostics would report "FFmpeg paused: Yes".

These may well be the right defaults, but #918 is about hardware fallback, and CLAUDE.md asks for one concern per PR. I suggest moving the resource defaults into their own PR with an issue that records the decision, and doing the same for the Activity diagnostics panel. I haven't changed anything here.

web/src/pages/AdminActivity.tsx
Quick104: Fixed in c3ed66ffe8cacc26ab164ce0db5b8c15d2d02cdb. The console goes back to the original 12-row `component=ffmpeg` query. The diagnostics panel now takes its own `debugRows` from the snapshot and throttle queries. Checked with `tsc -b`, ESLint, and Prettier on `AdminActivity.tsx`, plus the three affected Vitest files (56 tests). I have not checked this in a browser.

internal/playback/transcode_manager.go
Quick104: Fixed in 8f9bfa56eb. Both reconstruct paths now build the pipeline only for `auto` and use the fallback loop only when `pipeline.Enabled()`. Everything else starts exactly as it did before this PR. The new reconstruct mode (`StartReconstructTranscode` / `startReconstructTranscodePipeline`) keeps a process that is still running at the readiness deadline. It stops a failed attempt with `CloseProcess()` when another attempt follows, and uses `Close()` only on the final failure. Two new tests cover this: `TestStartReconstructTranscodePipelineKeepsSlowProcess` and `TestStartReconstructTranscodePipelineKeepsOutputForNextAttempt`. Both fail with reconstruct mode turned off. `go test ./internal/playback ./internal/transcodenode` and `go vet` pass.

internal/transcodenode/server.go
Quick104: Fixed in the same commit (8f9bfa56eb). `spawnReconstruct` now takes the adaptive path only when the auto pipeline is enabled. Otherwise it runs the original `StartTranscode` plus the VideoToolbox retry block, which is now gated on `!adaptive` instead of `!autoRequested`. `go test ./internal/transcodenode` passes.

Quick104:
**Review summary for `c3ed66f`** (reviewed `d66f32e`, then pushed two fix commits)

Fixed on this branch:
- `8f9bfa5`: local and node reconstruct use adaptive fallback only when the auto pipeline is enabled. A reconstruct keeps a slow running process and keeps its output directory between attempts. Two new tests cover this.
- `c3ed66f`: the Activity FFmpeg console shows the stderr tail again, and the diagnostics panel uses its own event rows.

Still open:
- **Merge conflicts** with `main` in 8 files. `main` is 340 commits ahead of this branch's base, and the API v2 migration (#1075) and #1076 both touch these paths. This needs a rebase by the author.
- **The CPU-decode NVENC path ignores `playback.hw_device`**, because `hwupload_cuda` always opens CUDA device 0. The fix needs validation on NVIDIA hardware.
- **Needs a maintainer decision:** do the throttle and retention default changes, and the Activity diagnostics panel, belong in this PR? I recommend splitting both out from the #918 fallback.
- **The description should mention that explicit `hw_accel=nvenc` also changes.** H.264 High 10 sources now use the CUDA upload path instead of libx264.

Validation:
- On the PR head, `go test` passes locally for `internal/{playback,transcodenode,jellycompat,api/handlers,config,streamtoken}`.
- After the fixes, `go test` passes again for `playback` and `transcodenode`, and `go vet` passes for the touched packages.
- Web: `tsc -b`, ESLint, Prettier, and the three affected Vitest files pass.
- golangci-lint was not run.
- CI has not run on this PR. The fork workflow run on `d66f32e` is waiting for maintainer approval, and GitHub won't build the merge ref while the branch conflicts.

**Mergeability: 4/10, not ready.** The core fallback design is sound and fixes #918. Merge is blocked by the rebase onto `main`, the scope decision, and CI that has not run.

_Written with Claude Code (Claude Opus 5.5)._

PR #921, opened 2026-09-03
fix(jellycompat): report the actual transcode target resolution/bitrate to admin sessions

### LINKED ISSUE #920: jellycompat sessions show source resolution/bitrate instead of actual transcode target in admin panel
## Summary

The admin sessions panel (`/api/v1/admin/sessions`, backed by `internal/api/handlers/playback_sessions.go`) shows the wrong "Delivered" video resolution and a `null` `target_bitrate_kbps` for Jellyfin-compatible (jellycompat) playback sessions that are actually being transcoded and downscaled.

## Root cause

```go
// internal/api/handlers/playback_sessions.go:402-403
if row.VideoDecision == "transcode" && strings.TrimSpace(row.TargetResolution) == "" {
    row.TargetResolution = row.SourceVideoResolution
}
```

This falls back to the *source* resolution whenever `target_resolution` wasn't recorded. The native app's playback path (`playback_v3.go`) explicitly threads `TargetResolution`/`TargetBitrateKbps` into the telemetry/recipe-card recording, but the jellycompat playback path (`internal/jellycompat/streams.go`, `ensureTranscodeSessionWithToneMapMode`) never writes those fields into that same telemetry table — so the admin panel always shows the source's own resolution/bitrate for jellycompat sessions currently being transcoded, never the actual encode target.

## Reproduction

1. Deploy a build with #912 (MaxStreamingBitrate cap fix).
2. Play a high-bitrate 1080p source through a Jellyfin-compatible client (e.g. Jellyfin Android app) and pick a low quality profile (e.g. "360p - 420 Kbps").
3. Confirm via server logs that the actual ffmpeg command is correctly capped and downscaled (e.g. `-b:v 420k -maxrate 420k ... scale_vaapi=h=480`).
4. Open the admin sessions panel for that session — it shows `Delivered: H.264 · 1080p` and no target bitrate, even though the real encode is 480p/420kbps.

## Impact

Cosmetic/observability only — the actual transcode behavior is correct. But it makes verifying playback quality decisions from the admin UI unreliable for any jellycompat-driven session (misleads support/debugging), and was likely already broken before #912 for any jellycompat resolution/bitrate-limiting scenario (tone-map bitrate selection, 4K-disallowed downscales, etc.) — #912 just made it newly visible since jellycompat downscaling wasn't previously exercised.

## Suggested fix

Have the jellycompat transcode-start path (`ensureTranscodeSessionWithToneMapMode` / wherever it records stream details, see `recordTranscodeStreamDetails`) also populate `target_resolution` and `target_bitrate_kbps` in the same telemetry table the native `playback_v3.go` path writes to, so the admin panel reflects the actual encode target for jellycompat sessions too.

Found while testing #912.

rangoDJ:
## Summary

- Fixes #920: the admin sessions panel showed the *source's* resolution/bitrate instead of the actual transcode target for any Jellyfin-compatible (jellycompat) playback session being transcoded — including a `MaxStreamingBitrate`-capped downscale from #912.
- `SetTranscodeStreamDetails` recorded video/audio codec, hw accel, and tone-map mode on the upstream session, but never `TargetResolution`/`TargetBitrateKbps` — only the native (non-compat) playback path threaded those through. The admin view falls back to displaying the source's own resolution/bitrate whenever `target_resolution` is unset for a transcoding session.
- `recordTranscodeStreamDetails` now forwards `opts.TargetResolution` and `opts.TargetBitrateKbps` through the extended setter, so session sync and the admin activity views reflect what ffmpeg is actually doing.
- Cosmetic/observability only — the real encode was never affected, only what the admin panel reported about it.

**AI disclosure:** This PR was developed with AI assistance (Claude).

## Test plan

- [x] `TestSetTranscodeStreamDetails` (extended) — verifies `TargetResolution`/`TargetBitrateKbps` are recorded.
- [x] `TestRecordTranscodeStreamDetailsIncludesTargetResolutionAndBitrate` (new) — verifies the jellycompat handler forwards both fields through to the upstream session.
- [x] `gofmt -l` clean on all touched files.
- [x] `go build`/`go test ./internal/playback/...` locally.
- [x] `go test ./internal/jellycompat/...` (CI).
- [x] Verified live: deployed a temporary branch combining this fix with #912's `MaxStreamingBitrate` cap (main alone has no code path that sets a non-empty target resolution/bitrate to forward, so #912 was needed to produce something real to report), confirmed the admin sessions API showed the correct capped `target_resolution`/`target_bitrate_kbps` instead of falling back to the source's.

### DIFF
--- a/internal/jellycompat/audio_selection_test.go
+++ b/internal/jellycompat/audio_selection_test.go
@@ -149,7 +149,7 @@ func (m *testCompatSessionManager) SetTranscodeNodeURL(sessionID, url string) er
 }
 
 // SetTranscodeStreamDetails records the execution facts supplied by the compatibility handler.
-func (m *testCompatSessionManager) SetTranscodeStreamDetails(sessionID, targetVideoCodec, targetAudioCodec string, transcodeAudio bool, hwAccel string, toneMapMode tonemap.Mode) error {
+func (m *testCompatSessionManager) SetTranscodeStreamDetails(sessionID, targetVideoCodec, targetAudioCodec string, transcodeAudio bool, hwAccel string, toneMapMode tonemap.Mode, targetResolution string, targetBitrateKbps int) error {
 	session, ok := m.sessions[sessionID]
 	if !ok {
 		return playback.ErrSessionNotFound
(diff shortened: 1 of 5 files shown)


rangoDJ:
All checks pass. Live verification: main alone has no code path that sets a non-empty `TargetResolution`/`TargetBitrateKbps` for jellycompat's local transcode start (only the VideoToolbox tone-map path does, which isn't reachable on this Linux/QSV test host), so there was nothing capped to report without #912. I deployed a temporary branch combining this fix with #912 to a real server + Jellyfin Android client, capped a stream to 420 kbps, and confirmed the admin sessions API now correctly reports `target_resolution: "480p"`, `target_bitrate_kbps: 420` — matching the real ffmpeg command — instead of falling back to the source's `1080p`/uncapped values.

Also worth noting from that same testing session: selecting different quality tiers in the Jellyfin Android app and comparing the resulting dashboard resolution against Jellyfin's own canonical bitrate→resolution table (`jellyfin-web`'s `qualityOptions.js`) showed the dashboard is reporting *correctly* per Jellyfin's own reference semantics in every case checked (e.g. 8 Mbps → 1080p, 3 Mbps → 720p, 720 kbps → 480p) — the Android app's on-screen quality labels just don't match Jellyfin's own internal `maxHeight` for those same bitrate values. That's a labeling quirk in the Jellyfin Android client, not something actionable here.

Issue #922, opened 2026-09-03
Playback activity reporting and compatible-session correctness


blurbery:
## Problem

The Activity page needs to distinguish a whole playback session from its container, video, and audio operations. A copied video track is not sufficient to call an audio-transcoded session Direct Play. Output details must name the actual delivered container and protocol, rather than use “Remux” as a container name.

The same investigation exposed a separate lifecycle defect: parallel Jellyfin-compatible `Static=true` requests can create more than one playback session for one client play. A repeated title alone is not evidence of a duplicate; device, playback identity, and the selected media version matter.

## Proposed scope

The two linked, independently reviewable contributions are now open: [activity reporting #924](https://github.com/Silo-Server/silo-server/pull/924) and [compatible-session recovery #923](https://github.com/Silo-Server/silo-server/pull/923).

1. **Activity reporting:** use Direct Play / Remux / Direct Stream / Transcode consistently across row badges, summary bars, filters, and related admin views; retain Copy as a per-stream operation; propagate actual output-container/protocol facts through live session state, synchronization, and the admin API.
2. **Jellyfin-compatible lifecycle:** reserve static sessions atomically, resolve current playback by explicit device identity, preserve the selected edition during recovery, and only reconcile legacy duplicates when their identity and selected source can be proven.

The reporting change adds two nullable output-format fields and a capability. Neither contribution changes Android decoder advertisement or removes codec/route validation. The Android Direct Play work is already covered by [silo-android #268](https://github.com/Silo-Server/silo-android/pull/268) and server #876.

## Evidence and limits

I reproduced the display problems and duplicate activity on my deployment and directed the fixes and live verification. Production-derived evidence will contain only sanitized technical facts and aggregate/timing samples, with no account names, addresses, media titles, device identifiers, session identifiers, paths, or credentials.

A viewer reported faster startup and no buffering after the Android update. Existing server-retained client telemetry confirms original-file delivery, but does **not** establish a 1–2 second startup benchmark: the relevant retained build 16 first-frame sample is 3,917 ms. Session/transport preparation is a different metric and will be reported separately. I will not attribute the Android improvement to dashboard labels or present subjective feedback as a controlled benchmark.

Both contributions include focused regressions, disposable-PostgreSQL concurrency/migration coverage, upstream-based CI, reproducible microbenchmarks with graphs, and explicit limitations. #924 contains the shared pipeline explanation, related merged Android work, synthetic before/after UI evidence and the sanitized live timing table.

[...]

PR #284, opened 2026-07-02
feat(notifications): Recently-Added notifications for audiobooks

fluxis:
## Problem

Availability detection only ran for TV/movie/mixed libraries: `AvailabilityKinds` had only Episodes and Movies, and server channels only exposed `notify_new_movies` / `notify_new_episodes`. Audiobook libraries never created release events, so Discord/webhook channels could not announce newly added audiobooks (#270).

## Approach

Audiobooks become a third availability kind, reusing the existing seeding semantics end to end:

- **New `audiobook_availability` table** mirroring `movie_availability`; the movie recorder is generalized into `recordItemAvailability` (kind + dedupe-key parameterized) rather than copy-pasted, and gains `RecordAudiobookAvailabilityForLibrary/ForPaths`. `release_events` admits the new kind — item-shaped like movies (both CHECK constraints extended).
- **Detector + ingest executor**: `AvailabilityKinds.Audiobooks` is set for audiobook libraries; the first full scan seeds silently, later scans announce — identical semantics to movies.
- **Channel toggle** `notify_new_audiobooks`, **opt-in (default false)** for both existing and new channels: most channels were configured before this kind existed, and video-only servers shouldn't grow a default-on toggle. Threaded through repo/service/REST/admin UI; `WantsContentKind` routes it (old nodes skip unknown kinds by design).
- **Digest rendering**: audiobooks group like movies — one group per item, deduped across libraries, title-with-year, a "New audiobook available on Silo" Discord author line; metadata comes from the existing `media_items` batch fetch.
- **Upgrade path**: the system backfill sweep gains an audiobook pass with its own seed marker, so existing audiobook libraries seed on upgrade without spamming their back catalog.

## Testing

- DB-backed test: silent unseeded pass, event emission with kind `audiobook` once seeded, idempotent re-run. (This test caught both `release_events` CHECK constraints — the kind allowlist and the kind-shape check — which the migration now extends.)
- Unit tests: `WantsContentKind`, audiobook grouping + cross-library dedupe + generic-title fallback, Discord rendering.
- **Live-validated**: on upgrade boot, the backfill pass seeded 44,557 audiobook availability rows silently across 6 libraries and wrote the seed markers — exactly the no-back-catalog-spam behavior intended.

Additive-only per v1 rules: new JSON fields, new event kind, no renames or repurposes.

Fixes #270

## AI-use disclosure

Implemented with Claude Code (test-first, live-validated); human-reviewed before submission.

[...]

PR #275, opened 2026-07-02
fix(libraries): exclude manga chapters from admin Unmatched queue

fluxis:
## Problem

With the Manga plugin enabled, successfully matched manga items appear stuck in the admin Unmatched Items queue, typed as 'ebook'. Manga chapter rows are internal sub-units resolved through their series: they intentionally keep `status='pending'` and `type='ebook'`, so `HandleListUnmatchedItems` — which filters only on status — floods the queue with them even after the series matched. (Scan-time retry was already fixed by #142; this is the remaining admin-UI symptom.)

## Approach

Apply the shared `catalog.MangaChapterExclusionWhere` guard to both the count and list queries in `HandleListUnmatchedItems`, exactly as every other catalog listing surface (browse, sections, discovery, search) already does. The predicate is an index-backed anti-join on `manga_chapters.chapter_content_id` (primary key), and is a no-op for regular ebooks and non-ebook types.

Added a DB-backed regression test seeding a matched manga series + linked pending chapter + genuinely pending plain ebook: before the fix the chapter appeared and total was 2; after, only the plain ebook remains and total is 1.

## Risks / follow-ups

Genuinely orphaned chapter rows (no series link) still surface, which is correct — they need operator attention.

Fixes #204
Fixes #187 (same defect, reported on GitHub directly; #204 is the Discord-ingested duplicate)

## AI-use disclosure

Implemented with Claude Code (test-first); human-reviewed before submission.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Unmatched items now exclude manga chapters from the list of actionable entries.
  * The total count for unmatched items now matches what users actually see in the list.
* **Tests**
  * Added coverage to verify manga chapters are omitted while regular unmatched items still appear.

PR #924, opened 2026-09-03
fix(activity): report playback methods and delivered formats

blurbery:
## Problem

Related issue: #922.

I want Activity to describe what the server is actually doing. A session can copy HEVC video while converting its audio to AAC; that is not complete-file Direct Play. Likewise, “Remux” describes an operation, not the container the viewer receives.

On the live deployment, the collapsed badge, summary bar, and expanded details did not consistently describe the same operation. Output details could fall back to the source format or show “MKV → Remux”, concealing the actual output container. I reproduced those display gaps while confirming the improved Android playback pipeline.

### Keep whole-session classification separate from stream operations

| Authoritative admin value | Display label | Server operation |
| --- | --- | --- |
| `direct` | Direct Play | Deliver the original file unchanged. |
| `remux` | Remux | Repackage the audio and video without re-encoding either. |
| `direct_stream` | Direct Stream | Copy video without re-encoding; transcode audio. |
| `transcode` | Transcode | Re-encode video; audio may be copied or transcoded. |

Unknown telemetry remains Unknown; it is not another successful playback scope. **Copy remains a per-stream badge**, not a whole-session category. Its explanation is “Copied without re-encoding.” This avoids promising byte-identical packets when muxing or bitstream-header conversion may change representation without a video re-encode.

The classification describes work on the server. A client still has to decode the received media, and may make supported output or Dolby Vision base-layer choices; that does not turn original-file server delivery into server-side transcoding.

- One overall method badge appears beside Details on desktop and mobile.
- The live bar, legend, filtering, row badge, Now Playing widget, and admin session view use the shared vocabulary. History/user-detail rows format their stored transport labels; they do not invent a whole-session classification from historical records that lack component decisions.
- Audio/video decisions remain visible separately.
- Audio-transcode, hardware/software execution, and tone-map detail remain in the expanded panel instead of competing with a second top-level method badge.
- The server remains authoritative through `effective_play_method`; the existing fallback for rows without that field uses the same decision reduction.

This follows [Jellyfin's documented four playback methods](https://jellyfin.org/docs/general/post-install/transcoding/#types-of-transcoding). [Plex](https://support.plex.tv/articles/200250387-streaming-media-direct-play-and-direct-stream/) and [Emby](https://support.emby.media/support/articles/DirectPlay-Stream-Transcoding.html) use broader Direct Stream descriptions that include packaging and possible audio conversion. Their terminology supports the distinction but is not an identical four-bucket API or dashboard layout. [FFmpeg's streamcopy documentation](https://ffmpeg.org/ffmpeg.html#Streamcopy) explains the distinction between copying encoded streams and decoding/re-encoding them.

### Report the selected output format end to end

```text
Serving transport / selected muxer
    → playback session output_container + output_protocol
    → route replacement, snapshot and restoration
    → node session synchronization
    → nullable playback_sessions_sync columns
    → admin API
    → delivered format in collapsed and expanded Activity details
```

The output is recorded independently of session scope:

- Original HTTP reports the original file container.
- Progressive remux reports fMP4 over HTTP. A newly recorded format triggers a bounded immediate admin-session sync after the fields are set; unchanged requests do not flush again. This closes the first-snapshot gap without adding reporting flushes to Direct Play.
- HLS reports its actual segment container: fMP4 or MPEG-TS, plus HLS as the delivery protocol.
- `HLSOutputContainer` is shared with FFmpeg argument construction. Copied MPEG-2 and forced-TS cases cannot be guessed as fMP4 from the word “copy”.
- The V3 remux state records the actual video-copy operation. Native, Jellyfin-compatible, local/remote transcode reporting, and recipe-card reconstruction carry the output facts.
- A complete route replacement can clear stale output facts. Snapshot comparison includes both fields, so a format-only change is synchronized.
- When a node has not supplied output facts, the UI explicitly says the output container is unknown. It does not substitute “Remux”, treat HLS as a container, or infer the delivered container from the source.

The migration adds only nullable `output_container` and `output_protocol` text columns. The admin capabilities response advertises `output_format`, and the API documentation describes the fields.

### API and rollout boundaries

The admin `effective_play_method` value `audio` becomes `direct_stream`. This is a pre-lock vocabulary change, documented in `docs/architecture/v1-scope.md`; it is **not** a rename of the native playback plan's `play_method` or its delivery values.

I checked the current Android/Apple source trees and their playback API definitions. No consumer of this admin field was found there; the affected first-party admin views live in this repository and are updated here. Other admin API consumers must use the advertised vocabulary.

Apply the additive migration before running the new readers/writers. Older nodes may leave the fields absent; that case is deliberately visible as unknown. Keep the nullable columns if rolling application code back.

This PR does not change codec eligibility, Android capability advertisement, selected-track policy, playback authorization, or the chosen FFmpeg packaging behavior. The shared muxer helper exposes the existing decision to reporting. It does not force Direct Play or disable a legitimate fallback.

## Relationship to session recovery

The duplicate-session/recovery contribution is [#923](https://github.com/Silo-Server/silo-server/pull/923). It is linked because the defects were found in the same investigation, but it changes lifecycle/identity handling rather than classification. Each PR is based directly on upstream and can be reviewed or reverted independently. No fork-only deployment configuration or unrelated fork history is included.

### How this fits the merged Android work

I researched and directed the playback fixes, then confirmed the behavior through live testing. This is a connected client/server effort, not an isolated badge change.

The published [Android build 16 release](https://github.com/Silo-Server/silo-android/releases/tag/v1.0.0%2B16) points to `beed70f9aa2b17ac56ac852cafa4d25e0e689e4c`. I verified that its history includes:

| Merged work | Role in the pipeline |
| --- | --- |
| [Android #250](https://github.com/Silo-Server/silo-android/pull/250) | Ignore non-material output-route callbacks and retain failed-plan history; real capability changes still replan. |
| [Android #262](https://github.com/Silo-Server/silo-android/pull/262) / [server #841](https://github.com/Silo-Server/silo-server/pull/841) | Identify native Media3 HLS so the server chooses compatible HEVC/Dolby Vision packaging. This is the packaged fallback, not Original HTTP. |
| [Android #268](https://github.com/Silo-Server/silo-android/pull/268) / [server #876](https://github.com/Silo-Server/silo-server/pull/876) | Advertise verified Media3/FFmpeg audio decoders and selected-source-track support, making eligible MKV files usable through original HTTP. Server #876 adds contract tests; it does not change runtime behavior. |
| [Android #275](https://github.com/Silo-Server/silo-android/pull/275) / [server #899](https://github.com/Silo-Server/silo-server/pull/899) | Plan and validate Dolby Vision Profile 8 base-layer playback against the actual decoder and active display. |
| [Android #282](https://github.com/Silo-Server/silo-android/pull/282), [#284](https://github.com/Silo-Server/silo-android/pull/284) | Own and release display bindings correctly, require a named base range, and prevent audio failures from quarantining a video transformation. |

[Android #259](https://github.com/Silo-Server/silo-android/pull/259) is related seek work: use the mounted transport when it can serve the target, with server re-anchoring retained outside that range. Seeking is not the startup benchmark below.

These references explain how the wider pipeline became reliable. The current PR does not claim to enable those already-merged Android decoders, force unsupported files into Direct Play, or demonstrate that every referenced change caused one measured result.

### What the successful Direct Play route means

For the live same-file check, the source was an MKV containing HEVC 2160p video and TrueHD audio. With the updated client, the server chose `original_http` / `direct`: original container, original encoded video, and original encoded audio. The earlier packaged fallback copied the video and converted audio to AAC 5.1, so its whole-session display is Direct Stream, with video Copy and audio Transcode.

```text
Eligible updated client
  original MKV → original HTTP → client decode/output
  server scope: Direct Play; container/video/audio: Direct

Audio conversion required
  original MKV → copied HEVC + encoded AAC → streaming package → client
  server scope: Direct Stream; video: Copy; audio: Transcode
```

Those labels describe server processing, not a claim that the client's HDMI/audio sink receives the compressed source unchanged. Server-side Direct Play alone does not prove a particular Dolby Vision display mode, lossless audio passthrough, or Atmos output.

## Validation and evidence

Latest follow-up: `c46706b3e438ed692bd88d5bda13d4097994d46a`. A new regression reproduced the progressive-remux output fields missing from the immediate admin sync; it now passes, including unchanged-request and Direct Play checks. [The full upstream CI rerun](https://github.com/Silo-Server/silo-server/actions/runs/33712931641) is pending. The completed gate, screenshot artifact and benchmark below retain their exact measured commit; they are not relabeled as measurements of this follow-up.

[Full repository CI passed](https://github.com/blurbery/silo-server/actions/runs/33711307460) on `22012d7de83d447b59301a0096a1c17c4995aa7e`:

- Go build, gofmt, vet, changed-line golangci-lint, and `make test-go`.
- Web lint, format check, production build, and `make test-web`: **359 test files, 3,077 tests passed; one screenshot-export test skipped in the normal suite and explicitly run successfully for each of the two screenshot fixtures**.
- Settings-binding generation, playback-fixture generation, and docs-hygiene gates.
- A real PostgreSQL 17 test applies the additive migration in an isolated schema and round-trips output facts through the actual reconciler's insert, replacement, snapshot load and clear paths. It runs with an explicit test DSN, not a skipped database test.
- Focused regressions cover the four server buckets, unknown input, copied-video HLS with/without encoded audio, actual FFmpeg segment-container agreement, V3 delivery facts, session replacement/rollback, format-only synchronization, and desktop/mobile badge/legend/filter consistency.

Actual migration regression output:

```text
--- PASS: TestReconcilerOutputFormatPostgres (0.02s)
```

Local targeted Go tests, focused web tests, ESLint, TypeScript checking, formatting and `git diff --check` passed. The full gate and browser evidence ran on GitHub-hosted runners; no heavy local build, live production benchmark, new playback session, or production mutation was used for this contribution.

The runtime reporting changes were exercised on the deployed fork during the live investigation. The upstream port also tightens unknown-output wording and includes dedicated CI evidence/tests; those final upstream-only refinements have not been deployed over the working server.

### Before/after UI evidence

[Download the before/after PNG evidence from GitHub](https://github.com/blurbery/silo-server/actions/runs/33711307460/artifacts/9877146255).

| State | Before | After |
| --- | --- | --- |
| Collapsed Activity | `before/collapsed.png` | `after/collapsed.png` |
| Expanded first session | `before/expanded.png` | `after/expanded.png` |

These are Chromium renders of the actual Activity component's DOM fixtures and built stylesheet, in the default Cinema Dark theme, at 1440 × 1000. The fixtures use only synthetic viewers/media. The baseline component is upstream `e7677886fe850f545bf19cc198cac267b87d0c90`; the after component is `22012d7de83d447b59301a0096a1c17c4995aa7e`. The artifact includes those commits and fixture HTML; the workflow reproduces both views. No production screenshots or raw production logs were uploaded, and no PR-only image assets were committed.

I visually checked all four images: the four-color bar agrees with the overall row badges, the Direct Stream example retains Copy/Transcode component badges, and the expanded container is fMP4 (HLS), not “Remux”. This is an isolated-component visual check, not a new authenticated production-browser test. The CI artifact is retained for 14 days and can be regenerated by rerunning the workflow.

### Reporting microbenchmark

Measured on `22012d7de83d447b59301a0096a1c17c4995aa7e` by [GitHub Actions / Go](https://github.com/blurbery/silo-server/actions/runs/33711307460/job/100511186000), using Go 1.26.4 on Linux/amd64. The runner reports an AMD EPYC 7763 host CPU and benchmark suffix `-4`. Each case has five 250 ms benchmark samples.

```sh
go test ./internal/api/handlers -run '^$' \
  -bench '^BenchmarkActivitySessionEnrichment$' \
  -benchmem -benchtime=250ms -count=5
```

| Activity enrichment case | Median ns/op | Range ns/op | B/op | Allocs/op |
| --- | ---: | ---: | ---: | ---: |
| Direct Play | 73.25 | 72.92–73.89 | 24 | 1 |
| Remux | 71.15 | 70.81–71.73 | 24 | 1 |
| Direct Stream | 74.62 | 74.34–81.77 | 24 | 1 |
| Transcode | 82.91 | 82.67–83.45 | 24 | 1 |

Median reporting cost, one block ≈ 5 ns/op; lower is shorter:

```text
Direct Play     ███████████████           73.25 ns/op
Remux           ██████████████            71.15 ns/op
Direct Stream   ███████████████           74.62 ns/op
Transcode       █████████████████         82.91 ns/op
```

This measures `enrichPlaybackSessionRow` for known source metadata with no audio-track JSON payload. It excludes database access, media I/O, synchronization, FFmpeg, browser rendering, and client startup. It is a reproducible cost measurement of the changed reporting function, **not a before/after startup or server-throughput speedup**.

<details>
<summary>Actual CI benchmark output (all samples)</summary>

```text
BenchmarkActivitySessionEnrichment/direct-4         	 3941522	        73.89 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct-4         	 4034472	        73.25 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct-4         	 4030903	        73.63 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct-4         	 4061958	        73.19 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct-4         	 4108917	        72.92 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/remux-4          	 4131652	        71.15 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/remux-4          	 4135554	        71.26 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/remux-4          	 4233698	        70.81 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/remux-4          	 4093534	        71.73 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/remux-4          	 4170685	        71.05 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct_stream-4  	 3918577	        81.77 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct_stream-4  	 4016407	        74.62 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct_stream-4  	 3956775	        74.48 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct_stream-4  	 3946285	        74.83 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/direct_stream-4  	 4005766	        74.34 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/transcode-4      	 3583460	        82.67 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/transcode-4      	 3571839	        83.03 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/transcode-4      	 3578325	        82.86 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/transcode-4      	 3618751	        82.91 ns/op	      24 B/op	       1 allocs/op
BenchmarkActivitySessionEnrichment/transcode-4      	 3535753	        83.45 ns/op	      24 B/op	       1 allocs/op
```

</details>

### Live timing evidence: what is measured and what is not

A viewer reported that startup changed from roughly five seconds followed by buffering to roughly one–two seconds without further buffering after updating the Android client. I confirmed successful Direct Play on the deployment. I kept that feedback separate from the retained timing measurements.

I queried a bounded 24-hour window read-only and correlated the same viewer, effective media file, and playback attempt on the server. Only the initial `first_frame` event for each distinct attempt is included; later seek/replan frames are excluded. All identity joins stayed private. The table below is a sanitized extraction of numeric fields, **not reconstructed raw log lines**.

| Same-file attempt | Client build | Delivery | Server start total (ms) | Session/transport commit (ms) | Client first frame (ms) |
| --- | --- | --- | ---: | ---: | ---: |
| Earlier A | 15 | `server_remux_hls` | 560 | 427 | 2736 |
| Earlier B | 15 | `server_remux_hls` | 1371 | 1358 | 2395 |
| Earlier C | 15 | `server_remux_hls` | 1434 | 1417 | 943 |
| Earlier D | 15 | `server_remux_hls` | 1936 | 1928 | 2216 |
| Earlier E | 15 | `server_remux_hls` | 1818 | 971 | 2286 |
| Earlier F | 15 | `server_remux_hls` | 2770 | 2705 | 1688 |
| Earlier G | 15 | `server_remux_hls` | 1570 | 1117 | 2236 |
| Later A | 16 | `original_http` | 1285 | 21 | 3917 |

Separate timer scopes; lower is shorter. Bars share a 0–4,000 ms scale, approximately 100 ms per block (`▏` marks a nonzero value below one block):

```text
Session/transport commit
  build 15 median, n=7  ██████████████                            1358 ms
  build 16 sample, n=1  ▏                                           21 ms

Server start total
  build 15 median, n=7  ████████████████                          1570 ms
  build 16 sample, n=1  █████████████                             1285 ms

Client plan-installed → first frame
  build 15 median, n=7  ██████████████████████                    2236 ms
  build 16 sample, n=1  ███████████████████████████████████████   3917 ms
```

The server's preparation stage is shorter on the retained Original HTTP attempt, but that attempt spent 1,253 ms in planning. The client first-frame measurement is **not faster** than the earlier median. There is only one later same-file sample, with uncontrolled cache, network, seek/resume position, and device conditions. This is observational evidence, not a controlled before/after performance test.

The Android [route-event timer](https://github.com/Silo-Server/silo-android/blob/beed70f9aa2b17ac56ac852cafa4d25e0e689e4c/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackSessionManager.kt#L2394-L2413) measures from installation of the active client plan to the first rendered frame, after the server's decision. It is not a click-to-picture stopwatch; I have not added the two timer scopes and presented the result as one.

A second build 16 file had different timings and is deliberately excluded from this same-file comparison. The server's retained route-event allowlist does not supply a rebuffer-count/duration series. Consequently, **the exact one–two second startup and zero-buffering claims remain attributed viewer feedback, not benchmark conclusions**. None of these media-start timings are credited to an Activity label change.

## Review, risks and limits

The adversarial review traced execution facts through session replacement/reconstruction, synchronization, API reduction, and every changed display consumer. I checked unknown and legacy values, copied MPEG-2/forced-TS packaging, copied video with encoded audio, format-only updates, and the separation of admin metadata from executable playback plans. That review tightened unknown-container behavior; CI caught a duplicated tone-map literal, which now uses the existing constants. The repository's automatic CodeRabbit review identified a progressive-remux synchronization gap. I verified it with a failing handler regression and fixed the ordering so changed format fields are flushed immediately; repeated unchanged requests and Direct Play add no such flush. This follow-up is verified in tests and is not claimed as a new production deployment.

Remaining limits are explicit: an old node cannot report facts it never collected; third-party admin clients need the new vocabulary; the nullable migration must precede application rollout; and the live samples do not establish a one–two second startup benchmark. No production logs, account names, network addresses, media titles, device/session identifiers, private paths, or credentials are included.

## AI Disclosure

- Harness: OpenAI Codex desktop.
- Tool(s): Codex local command execution (`exec_command`), `apply_patch`, GitHub CLI, and Codex web search.
- Model(s): `gpt-5.6-sol`, reasoning effort `ultra` (Sol Ultra).
- Involvement: AI-assisted. I researched the playback behavior, designed the intended behavior and scope, directed Codex through the investigation and implementation, and performed the live test confirmations. Codex assisted with code, regression tests, reference checks, CI analysis, and this write-up.
- Adversarial review: Same-agent source-path and diff review, supported by focused tests, real PostgreSQL CI coverage, generated synthetic UI comparisons, and the repository CI gate. I also used the repository's automatic CodeRabbit review after checking the finding against the code and reproducing it. Its underlying model was not exposed, so I do not claim a model identity for it. No separately delegated Codex agent was used. Findings and resolutions are summarized above.


internal/jellycompat/streams.go
```diff
@@ -463,6 +463,9 @@ func (h *PlaybackHandler) HandleVideoStream(w http.ResponseWriter, r *http.Reque
 	// The attach above is a no-op on the first request of a session, which has
 	// no upstream id yet. Now it does, and no byte has been written.
 	attachCompatStream(r.Context(), session, playSession, source.FileID)
+	if method == "remux" {
+		h.recordOutputFormat(playSession.UpstreamSessionID, playback.OutputContainerFMP4, playback.OutputProtocolHTTP)
```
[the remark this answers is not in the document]
blurbery: Fixed in c46706b3. I first reproduced the missing immediate flush with a handler regression. Recording a changed fMP4/HTTP format now triggers bounded sync after the fields are set. The test also repeats the request and covers Direct Play: neither adds an unnecessary reporting flush. The nearby stream/start/stop tests pass. Full CI is rerunning; production has not been changed.

internal/api/handlers/playback_sessions.go
```diff
@@ -59,6 +61,8 @@ type playbackSessionRow struct {
 	TranscodeNodeURL     string    `json:"-"`
 	TargetResolution     string    `json:"target_resolution,omitempty"`
 	TargetVideoCodec     string    `json:"target_video_codec,omitempty"`
+	OutputContainer      string    `json:"output_container,omitempty"`
```
Quick104: **Blocking: after merging with main, the output format never reaches the dashboard.** Since #1075 the web Activity views read `/api/v2/admin/sessions`, and `internal/apiv2/admin_sessions.go` copies each field of this row explicitly. These fields are missing from that projection, so every non-direct session would render "Unknown output container". The v2 household list would also keep sending `audio`, which the new web vocabulary renders as "Unknown".

This struct is also the frozen `/api/v1` payload for `/admin/sessions`, `/profiles/household/sessions`, and the `sessions` websocket snapshot, so these JSON tags would add fields to the frozen bridge.

I'm handling this in the follow-up push. The fields become `json:"-"` here, the same pattern `routing_network_provider` uses. They are added to the v2 row, and the v2 capability response advertises them with `output_format`.

internal/api/handlers/playback_sessions.go
```diff
@@ -471,13 +497,13 @@ func effectivePlayMethod(videoDecision, audioDecision string) string {
 	case videoDecision == "" && audioDecision == "":
 		return ""
 	case videoDecision == "transcode":
-		return "transcode"
+		return effectivePlayMethodTranscode
 	case audioDecision == "transcode":
-		return "audio"
+		return effectivePlayMethodDirectStream
```
Quick104: **Blocking under the current API policy: this changes the frozen `/api/v1` vocabulary.** When the PR was opened, v1 was still an alpha contract, and a recorded pre-lock change was allowed. #1075 has since frozen v1. Only critical bridge fixes land there, and new contract work targets v2 (`docs/architecture/api-contract.md`, "v1 lifecycle and release sequence"). Renaming `audio` here changes `effective_play_method` and the capability values on every v1 surface that serializes this row.

I'm handling this in the follow-up push. The shared read model and v1 keep `audio`. The v2 projection reports `direct_stream` on the admin and household lists and in `effective_play_method_values`. The `v1-scope.md` removal entry goes away because v1 no longer changes.

.github/workflows/ci.yml
```diff
@@ -172,6 +195,33 @@ jobs:
         working-directory: .
         run: make verify-settings-bindings-web
 
+      - name: Capture synthetic Activity UI evidence
```
Quick104: **Project fit: PR-evidence tooling in the shared CI gate.** This step runs on every PR and push. It fetches upstream main, rewrites two source files to their merge-base versions in the middle of the job, and installs Playwright and Chromium with system packages, all to produce screenshots for this PR. The repository guidance (CLAUDE.md, "Pull requests") asks contributors not to capture screenshots just to prepare a PR and not to commit PR-only assets. The benchmark step at line 136 has no threshold, so it adds time to every Go run without gating anything. This workflow also conflicts with main, which has since split the Go job.

The Postgres test at `internal/worker/reconciler_postgres_test.go:44` hand-builds `playback_sessions_sync`. After the merge the reconciler also writes `routing_network_provider`, which that table lacks, so the test fails whenever `SILO_TEST_DATABASE_URL` is set.

I'm handling this in the follow-up push. It keeps main's workflow, drops `scripts/capture-activity-evidence.mjs` and the fixture exporter, and moves the output-format round trip into main's DB-gated `TestReconcileNodeSessionsAfterAccountDeletion`, which builds on the migrated schema.

Quick104 (commented):
Review of `c46706b3e`. The reporting design holds up. Output facts come from the transport's own muxer decision (`HLSOutputContainer` is shared with FFmpeg argument construction), full route replacements clear stale values, snapshot comparison includes both fields, and unknown stays unknown. The branch no longer merges with main, though, and main's API v2 migration (#1075) changes where this work has to land. The three findings are inline. I'm pushing a merge from main plus a v2 wiring commit to this branch and will reply in each thread.

Non-blocking design note, left unchanged: the collapsed row no longer shows the HW/SW encoder or tone-map badges. Software transcodes used to show in red there; now they appear only in the expanded details. Maintainers may want to keep an at-a-glance signal for software encoding.

Review and fixes by Claude Opus 5.5 (Claude Code), posted on behalf of @Quick104.

internal/api/handlers/playback_sessions.go
Quick104: Fixed in [`8dc2a53a9`](https://github.com/Silo-Server/silo-server/pull/924/commits/8dc2a53a94130707703e04263bb1b90363a3a167), on top of the merge in [`d9d15bd4f`](https://github.com/Silo-Server/silo-server/pull/924/commits/d9d15bd4feb6bb9417f25f3ce527569e3494fa9a). The shared row keeps `output_container` and `output_protocol` as `json:"-"`, so the v1 admin list, v1 household list, and websocket snapshot keep their shape. `AdminPlaybackSession` in `internal/apiv2/admin_sessions.go` now carries both fields, and the capability response advertises `output_format`. The OpenAPI artifact and web schema types are regenerated.

Verified by `TestAdminSessionOutputFormatProjection` (the v2 row carries the fields, the v1 JSON omits them, and the capability advertises them), `TestActivityOutputFormatIsIndependentOfSessionScope`, a loader read against a freshly migrated database, the v2 contract gates, and green CI on `64ed6a83a` ([run](https://github.com/Silo-Server/silo-server/actions/runs/35796314796)).

internal/api/handlers/playback_sessions.go
Quick104: Fixed in [`8dc2a53a9`](https://github.com/Silo-Server/silo-server/pull/924/commits/8dc2a53a94130707703e04263bb1b90363a3a167), with the bridge side in the merge [`d9d15bd4f`](https://github.com/Silo-Server/silo-server/pull/924/commits/d9d15bd4feb6bb9417f25f3ce527569e3494fa9a). `effectivePlayMethod` and `AdminPlaybackSessionFeatures` keep the frozen `audio` value. The v2 projection maps it to `direct_stream` on the admin and household lists and in `effective_play_method_values`. The `v1-scope.md` entry is removed because v1 no longer changes.

Verified by `TestEffectivePlayMethodBuckets` and `TestSessionsCapabilitiesAdvertisesActivityFields` (bridge vocabulary unchanged), `TestAdminSessionOutputFormatProjection` (v2 vocabulary), the v2 breaking-change diff (additive only), and green CI on `64ed6a83a` ([run](https://github.com/Silo-Server/silo-server/actions/runs/35796314796)).

.github/workflows/ci.yml
Quick104: Fixed in the merge [`d9d15bd4f`](https://github.com/Silo-Server/silo-server/pull/924/commits/d9d15bd4feb6bb9417f25f3ce527569e3494fa9a). The branch uses main's workflow. `scripts/capture-activity-evidence.mjs`, the screenshot fixture exporter, and the per-run benchmark step are removed; the benchmark function stays in the test file. The output-format round trip now runs inside `TestReconcileNodeSessionsAfterAccountDeletion`, which copies the migrated `playback_sessions_sync` shape, so later columns can't drift from it.

Verified by running that test against a freshly migrated PostgreSQL 18 database (pass), and by green CI on `64ed6a83a` ([run](https://github.com/Silo-Server/silo-server/actions/runs/35796314796)).

Quick104:
## Review summary

Reviewed head: `c46706b3e`. Current head: `64ed6a83a`. CI passed on the current head.

**Addressed** (3 findings, all fixed on this branch and resolved):

1. The branch conflicted with main. After the API v2 migration, the web Activity views read `/api/v2`, so the output fields never reached the dashboard. Fixed with merges from main and a v2 projection of `output_container`, `output_protocol`, and `output_format`.
2. The `audio` → `direct_stream` rename and the new fields changed the frozen `/api/v1` bridge. v1 now keeps its alpha shape, v2 reports `direct_stream`, and the `v1-scope.md` entry is removed.
3. CI screenshot-evidence tooling, a per-run benchmark, and a hand-built DB schema test didn't fit the shared gate. The branch now uses main's workflow, and the Postgres round trip moves into main's DB-gated reconciler test.

**Still open (non-blocking; needs a maintainer call):** the collapsed Activity row no longer shows the HW/SW encoder and tone-map badges. Software transcodes used to show in red there; now they appear only in the expanded details. I left this as the author designed it.

**Validation:** focused Go tests for the six touched packages. The DB-gated reconciler test and a loader read ran against a freshly migrated PostgreSQL 18 database. Also run: `make test-web` (608 files, 4,727 tests), web lint, format check, typecheck, `make lint-changed`, and the v2 OpenAPI, web-type, fixture, and breaking-change gates. No live playback session was run against a deployed server.

**Mergeability: 8/10 — conditional.** The design is sound: output facts come from the transport's own muxer decision and stay unknown when unreported, and the change now respects the v1 freeze. Required checks pass. Before merge, a maintainer should accept or revert the collapsed-row badge change.

Review and fixes by Claude Opus 5.5 (Claude Code), posted on behalf of @Quick104.

PR #1004, opened 2026-09-08
fix(web): render neutral empty state and loading placeholder for ambiguous roots

### LINKED ISSUE #1000: [bug] Ambiguous Roots shows a warning with 0 items and treats unloaded results as zero
## Problem

Admin → Libraries displays an amber “Ambiguous Roots” section with a count of `0`, even when there is nothing to resolve. This makes a healthy library look like it needs attention.

## Steps to reproduce

1. Open Admin → Libraries with at least one library configured and no ambiguous roots.
2. Observe the collapsed “Ambiguous Roots” section showing `0`.
3. Expand it. After loading, it shows “No ambiguous roots for this library.”

## Expected

Confirmed empty results should not appear as an actionable warning. Before results load, the UI should distinguish an unknown count from a confirmed zero.

## Actual

The section remains visible with zero results. On a fresh load without cached results, it also displays `0` before checking whether any ambiguous roots exist.

## Technical findings

- [`AmbiguousRootsSection`](https://github.com/Silo-Server/silo-server/blob/8771e3a02e71575e562c1eb10e9662e60f63c18e/web/src/pages/AdminLibraries.tsx#L1411) only hides when there are no libraries. It does not hide when there are no ambiguous roots.
- The query uses `enabled: open`, so it waits until expansion. Meanwhile, `rootPages?.pages[0]?.total ?? 0` turns missing data into a displayed zero.
- The count belongs to the selected library and current search, not all libraries.
- Simply hiding the section when the count is zero would prevent opening it and triggering the initial query. It could also make library selection or search controls inaccessible.
- An [existing test](https://github.com/Silo-Server/silo-server/blob/8771e3a02e71575e562c1eb10e9662e60f63c18e/web/src/pages/AdminLibraries.test.tsx#L358) explicitly preserves the empty section.

## Acceptance criteria

- Distinguish unloaded, loading, failed, and confirmed empty results.
- Use a neutral empty state or hide the warning when emptiness is confirmed.
- Keep other libraries and search controls reachable.
- Preserve on-demand pagination.
- Cover initial loading, confirmed zero, populated results, and switching libraries in tests.

## Validation

Confirmed through source inspection at `8771e3a02`; all 15 Admin Libraries tests passed. The reported deployment was not inspected directly.

rangoDJ:
## Summary

Fixes #1000.

Resolves an issue where `AmbiguousRootsSection` in `AdminLibraries.tsx` displayed an amber warning badge with count `0` on healthy libraries and treated in-flight or unloaded query data as a premature zero.

---

### Background & Problem

1. **False-Positive Warning Styling:**
   - In `AdminLibraries.tsx`, `AmbiguousRootsSection` wrapped its contents in `<CollapsibleDiagnosticsSection>` with hardcoded warning styling (`<FolderOpen className="h-4 w-4 text-amber-500" />` and default `bg-amber-500/10`).
   - On healthy servers where no ambiguous roots exist (count is `0`), the section still presented as an active, amber diagnostic warning. This caused administrators to believe there was an unresolved scanner problem when the library was in fact completely healthy.

2. **Premature Zero During Loading:**
   - `const { data: roots = [] } = useLibraryRoots(...)` defaulted missing or in-flight data to `[]`.
   - On initial mount or library switch, the header immediately flashed an amber `0` before the network request resolved, conflating an unloaded query with a confirmed zero.

3. **Why the Section Cannot Be Conditionally Hidden on Zero:**
   - The library selector dropdown (`<Select>`) lives inside `AmbiguousRootsSection`. If library A has 0 ambiguous roots but library B has 3, hiding the section on count 0 would permanently lock out the administrator from switching to library B.
   - The section must remain accessible while reflecting the correct severity state.

---

#### 1. `CollapsibleDiagnosticsSection.tsx`
- **Added `isLoading?: boolean` Support:**
  - When `isLoading` is true, renders a subtle placeholder dash (`—` with `aria-label="Loading count"`) in both open and collapsed badge/counter positions instead of a bold `0`.
- **Neutral Styling for Zero Counts:**
  - Applied `text-muted-foreground` to `count === 0` in the collapsed count view, visually distinguishing confirmed zero/healthy states from actionable non-zero counts.

#### 2. `AdminLibraries.tsx` (`AmbiguousRootsSection`)
- **Destructured Query State:**
  - Extracted `{ data: roots = [], isLoading }` from `useLibraryRoots(effectiveSelectedLibraryId, "ambiguous")`.
  - Derived `isWarning = !isLoading && roots.length > 0`.
- **Dynamic Icon & Container Styling:**
  - When confirmed zero (`!isLoading && roots.length === 0`): Renders neutral icon styling (`<FolderOpen className="h-4 w-4 text-muted-foreground" />` with `iconClassName="bg-muted/50"`).
  - When populated (`roots.length > 0`): Renders the amber warning styling (`<FolderOpen className="h-4 w-4 text-amber-500" />` with `iconClassName="bg-amber-500/10"`).
  - While loading: Passes `isLoading={isLoading}` to display the placeholder dash.
- **Contextual Table Empty Message:**
  - When `filteredRoots.length === 0`:
    - With active search: *"No ambiguous roots match your filter."*
    - Without active search: *"No ambiguous roots for this library."*

#### 3. Tests
- **`CollapsibleDiagnosticsSection.test.tsx` (New):**
  - Verified rendering collapsed title, description, and count.
  - Verified loading indicator (`aria-label="Loading count"`) when `isLoading` is true.
  - Verified neutral text styling on zero count.
  - Verified toggle interaction and child content visibility.
- **`AdminLibraries.test.tsx`:**
  - Updated empty-state test to verify neutral styling (`text-muted-foreground`, `bg-muted/50`) when count is 0.
  - Added test verifying loading placeholder state during in-flight queries.
  - Updated populated test to verify amber warning classes (`text-amber-500`, `bg-amber-500/10`) when roots exist.

---

### Test Plan

- [x] Ran unit tests for `CollapsibleDiagnosticsSection.test.tsx` and `AdminLibraries.test.tsx` via Vitest (`15 passed`).
- [x] Verified full TypeScript compilation with `pnpm tsc -b` (`0 errors`).
- [x] Verified ESLint checks with `pnpm lint` (`0 errors`).
- [x] Verified code formatting with Prettier.

## Summary by CodeRabbit

- **Improvements**
  - Added loading and error indicators for diagnostic counts and ambiguous library roots.
  - Error states now include clear messaging and destructive styling when results cannot be loaded.
  - Updated empty states to distinguish between no ambiguous roots and no matches for an active search filter.
  - Warning colors appear only when ambiguous roots are successfully found; empty states use neutral styling.
  - Diagnostic counts display an em dash while loading and use neutral styling when the count is zero.
  - Collapsible diagnostic sections now support clearer loading, error, and accessible status messaging.

### DIFF
--- a/web/src/pages/AdminLibraries.tsx
+++ b/web/src/pages/AdminLibraries.tsx
@@ -1412,7 +1412,7 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
   const [search, setSearch] = useState("");
   const [editingRoot, setEditingRoot] = useState<LibraryRoot | null>(null);
   const effectiveSelectedLibraryId = selectedLibraryId ?? libraries[0]?.id;
-  const { data: roots = [] } = useLibraryRoots(effectiveSelectedLibraryId, "ambiguous");
+  const { data: roots = [], isLoading } = useLibraryRoots(effectiveSelectedLibraryId, "ambiguous");
 
   const filteredRoots = useMemo(() => {
     if (!search) return roots;
@@ -1431,12 +1431,20 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
     return null;
   }
 
+  const isWarning = !isLoading && roots.length > 0;
+
   return (
     <CollapsibleDiagnosticsSection
       title="Ambiguous Roots"
       description="Scanner roots that stay visible but do not enter unattended metadata matching."
       count={roots.length}
-      icon={<FolderOpen className="h-4 w-4 text-amber-500" />}
+      isLoading={isLoading}
+      icon={
+        <FolderOpen
+          className={cn("h-4 w-4", isWarning ? "text-amber-500" : "text-muted-foreground")}
+        />
+      }
+      iconClassName={isWarning ? "bg-amber-500/10" : "bg-muted/50"}
       open={open}
       onOpenChange={setOpen}
     >
@@ -1490,7 +1498,9 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
             {filteredRoots.length === 0 ? (
               <TableRow>
                 <TableCell colSpan={5} className="text-muted-foreground text-center text-sm">
-                  No ambiguous roots for this library.
+                  {search
+                    ? "No ambiguous roots match your filter."
+                    : "No ambiguous roots for this library."}
                 </TableCell>
               </TableRow>
             ) : (
--- a/web/src/pages/AdminLibraries.tsx
+++ b/web/src/pages/AdminLibraries.tsx
@@ -1406,13 +1406,21 @@ function useSort<K extends string>(defaultField: K, defaultDir: SortDir = "desc"
 
 /* ─── Skipped Roots (Troubleshooting) ───────────────────────────── */
 
+/**
+ * AmbiguousRootsSection displays scanner roots that require manual resolution,
+ * handling loading, confirmed empty, populated warning, and error states.
+ */
 function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
   const [open, setOpen] = useState(false);
   const [selectedLibraryId, setSelectedLibraryId] = useState<number | undefined>(libraries[0]?.id);
   const [search, setSearch] = useState("");
   const [editingRoot, setEditingRoot] = useState<LibraryRoot | null>(null);
   const effectiveSelectedLibraryId = selectedLibraryId ?? libraries[0]?.id;
-  const { data: roots = [], isLoading } = useLibraryRoots(effectiveSelectedLibraryId, "ambiguous");
+  const {
+    data: roots = [],
+    isLoading,
+    isError,
+  } = useLibraryRoots(effectiveSelectedLibraryId, "ambiguous");
 
   const filteredRoots = useMemo(() => {
     if (!search) return roots;
@@ -1431,7 +1439,7 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
     return null;
   }
 
-  const isWarning = !isLoading && roots.length > 0;
+  const isWarning = !isLoading && !isError && roots.length > 0;
 
   return (
     <CollapsibleDiagnosticsSection
@@ -1439,12 +1447,17 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
       description="Scanner roots that stay visible but do not enter unattended metadata matching."
       count={roots.length}
       isLoading={isLoading}
+      isError={isError}
       icon={
-        <FolderOpen
-          className={cn("h-4 w-4", isWarning ? "text-amber-500" : "text-muted-foreground")}
-        />
+        isError ? (
+          <AlertTriangle className="text-destructive h-4 w-4" />
+        ) : (
+          <FolderOpen
+            className={cn("h-4 w-4", isWarning ? "text-amber-500" : "text-muted-foreground")}
+          />
+        )
       }
-      iconClassName={isWarning ? "bg-amber-500/10" : "bg-muted/50"}
+      iconClassName={isError ? "bg-destructive/10" : isWarning ? "bg-amber-500/10" : "bg-muted/50"}
       open={open}
       onOpenChange={setOpen}
     >
@@ -1495,7 +1508,13 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
             </TableRow>
           </TableHeader>
           <TableBody>
-            {filteredRoots.length === 0 ? (
+            {isError ? (
+              <TableRow>
+                <TableCell colSpan={5} className="text-destructive text-center text-sm">
+                  Failed to load ambiguous roots for this library.
+                </TableCell>
+              </TableRow>
+            ) : filteredRoots.length === 0 ? (
               <TableRow>
                 <TableCell colSpan={5} className="text-muted-foreground text-center text-sm">
                   {search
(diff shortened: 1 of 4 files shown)


web/src/pages/AdminLibraries.tsx
```diff
@@ -1406,13 +1406,21 @@ function useSort<K extends string>(defaultField: K, defaultDir: SortDir = "desc"
...
   const [selectedLibraryId, setSelectedLibraryId] = useState<number | undefined>(libraries[0]?.id);
   const [search, setSearch] = useState("");
   const [editingRoot, setEditingRoot] = useState<LibraryRoot | null>(null);
   const effectiveSelectedLibraryId = selectedLibraryId ?? libraries[0]?.id;
-  const { data: roots = [] } = useLibraryRoots(effectiveSelectedLibraryId, "ambiguous");
+  const {
+    data: roots = [],
+    isLoading,
+    isError,
+  } = useLibraryRoots(effectiveSelectedLibraryId, "ambiguous");
```
Quick104: **Blocking: after merging with `main`, a section that was never opened still shows a confirmed `0`.**

`main` now pages this listing on demand: `useLibraryRoots(..., { enabled: open, search })` returns infinite-query pages, and the header count is `pages[0].total`. Until the section opens, the query stays disabled. In TanStack Query v5, a disabled query with no data reports `isLoading === false` (`isLoading` is `isPending && isFetching`). If the branch merges and keeps `isLoading` as the "unknown" signal, the collapsed header on every page load shows a neutral `0` it never fetched. That's the "treats unloaded results as zero" half of #1000.

Fix I'm pushing: resolve the conflict onto main's paged query and treat "no page loaded yet for this library and search" (`rootPages === undefined`) as the unknown state. The header then shows the `—` placeholder until the first page arrives, and the table shows the loading row while it's open. I also renamed the component prop from `isLoading` to `countPending` because it now covers both "never fetched" and "fetching", and changed the screen-reader text to "Count not loaded".

web/src/pages/AdminLibraries.tsx
```diff
@@ -1487,10 +1508,24 @@ function AmbiguousRootsSection({ libraries }: { libraries: Library[] }) {
             </TableRow>
           </TableHeader>
           <TableBody>
-            {filteredRoots.length === 0 ? (
+            {isError ? (
```
Quick104: **Minor: a failure after rows have loaded replaces them with the error row.**

`isError` is also true when a query that already has data fails a later fetch, such as a background refetch here or, after the merge, a failed "Load more". This branch runs first, so the roots the admin was already viewing disappear behind "Failed to load…" and the header count turns into `!`.

Fix I'm pushing: show the error row and error badge only when no page has loaded (`rootPages === undefined && isError`). Rows that already loaded stay visible. A regression test covers it.

Quick104 (commented):
Review at a5e511f. The loading, error, and neutral-empty states are the right fix for #1000, and the three earlier CodeRabbit findings are addressed at this head. The branch conflicts with `main`, though, and the state checks here stop working once it's merged (see the first inline comment). I'm pushing a merge commit that resolves the conflict against main's paged roots query and fixes both inline findings, so you don't need to change anything.

_Reviewed with Claude Opus 5.5 (`claude-opus-5-5`) in Claude Code._

web/src/pages/AdminLibraries.tsx
Quick104: Fixed in [5ab3116](https://github.com/Silo-Server/silo-server/pull/1004/commits/5ab31162ae71595d78cae9d7d37a5fbff018426c). The merge resolves the conflict onto main's paged query. `AmbiguousRootsSection` treats `rootPages === undefined` as an unknown count and passes it as `countPending`, so the collapsed header shows `—` ("Count not loaded" for screen readers) until the first page for the selected library and search arrives. The open table shows the loading row in that state. New tests cover the never-opened header, the open loading row, confirmed zero, and switching libraries through the selector. The first two fail if the unknown check is removed. Verified with `vitest` on both suites (26 passed) and `tsc -b`.

web/src/pages/AdminLibraries.tsx
Quick104: Fixed in [5ab3116](https://github.com/Silo-Server/silo-server/pull/1004/commits/5ab31162ae71595d78cae9d7d37a5fbff018426c). The error badge, icon, and row now require `rootPages === undefined && isError`, so rows that already loaded stay visible when a later page or a refetch fails. "keeps loaded Ambiguous Roots visible when a later page fails" covers this and fails without the guard.

Quick104:
**Review summary at 5ab3116.** Both findings from the review are fixed in the merge commit, and their threads are resolved:

- **Blocking:** after the merge with `main`, the header showed a confirmed `0` for a query that never ran. The unknown state is now keyed on a missing first page.
- **Minor:** a failure after rows had loaded hid them behind the error row. Loaded rows now stay visible.

The three earlier CodeRabbit findings are also handled at this head.

Validation: `vitest` on both suites (26 passed), including library switching, which #1000 asks for. `tsc -b` is clean, and all CI checks pass. The first Go test run failed in `TestControlSocketReconnectResumesOnlySameOwnerAndInstallation` with a socket read timeout. That package isn't touched by this PR, and the test passed on rerun.

Remaining: nothing blocking. Changing the component prop from `isLoading` to `countPending` only affects this call site.

**Mergeability: 9/10, ready.** It fixes both halves of #1000, stays in the owning component, and keeps main's on-demand paging and the library and search controls.

_Reviewed with Claude Opus 5.5 (`claude-opus-5-5`) in Claude Code._

PR #1002, opened 2026-09-08
fix(playback): reconcile failed starts and retry retained requests

Quick104:
Playback starts could remain unconfirmed after worker startup failure or a lost response, and onboarding progress became unavailable behind the notification store wrapper. This draft adds global retained-attempt reconciliation, aligns worker manifest startup timing, releases owners when idle sessions expire, preserves optional store capabilities, and automatically retries the browser's exact saved START and STOP requests.

Related issue: #135

### Known blockers and handoff

- **Do not merge yet:** a confirmed missing-file START rejection remains saved in the browser. Later Play actions replay that old request before sending the newly selected file, so unrelated items all appear unavailable. Reproduce by starting a missing catalog file, then selecting a different item: the new page refers to the new file while POST START still contains the previous file and attempt IDs. Distinguish a confirmed pre-start rejection from an uncertain response without weakening the existing identity and ownership fences.
- `verify-route-inventory` rejects the generic type assertion in `internal/userstore/capability.go`. Resolve this interaction with the route verifier.
- Full web tests expose an audiobook retained-start assertion failure and a subtitle-font import failure; focused tests alone missed them.
- Deployment follow-up remains outside this patch: the audiobook media root is absent from the running development container's mounts. The video media root is mounted. Verify and restore the expected deployment override mounts separately.
- Apple and Android wire behavior remains unchanged; their existing terminal handling was inspected. Jellyfin compatibility keeps its legacy finalization path.

### Validation

Earlier focused verification passed 90 web tests, TypeScript checking, database-backed reconciliation/onboarding tests, and focused race tests for idle owner release and recovery. Live validation exercised a lost successful START response, automatic exact-request replay, rendered playback, and sequenced STOP completion. It did not cover the confirmed missing-file rejection described above.

Pre-PR checks passed: Go build and vet, formatting of Go files under cmd/internal, settings bindings, playback fixtures, offline routes, v2 OpenAPI, v2 contract compatibility against origin/apiv2, v2 fixtures, committed router artifact test, local-path checks, and production web build.

Required checks that failed or could not complete:
- Frozen dependency install refused noninteractive replacement of the existing modules directory.
- `golangci-lint` is not installed.
- Full web lint: constant-condition errors in adminImages.ts and adminSplit.ts, plus warnings.
- Web format check: 34 files reported.
- Full web tests: 4,365 tests passed, one audiobook recovery assertion failed, and the ASS subtitle test suite could not import a font asset.
- Route inventory: generic capability assertion rejected.
[...]

PR #1008, opened 2026-09-09
fix(jellycompat): redact credentials from request and debug logs

blurbery:
## Problem

Related issue: N/A, narrow security fix.

Jellyfin-compatible requests can put a session token or API key in `ApiKey` or `api_key`. The request logger writes `r.URL.RawQuery` under the ordinary `query` attribute, so the existing key-based log redactor does not mask it. The `HandleSeasons` panic logger has the same path. Rejected requests are logged too.

Optional `JELLYCOMPAT_DEBUG_LOG` capture also writes complete URLs and request/response bodies directly to its file. That can retain login `Pw`/`Password` fields, returned `AccessToken` values and tokens inside playback URLs. This file writer bypasses the slog redactor.

This is a CWE-532 credential-exposure fix. Someone who can read retained or exported logs could reuse a still-valid credential with its existing permissions. It does not establish unauthenticated access to logs, an administrator-level credential, or an actual account compromise. The normal request-log exposure was reproduced on a running deployment; debug-body exposure was reproduced in tests, not by enabling debug capture on production.

## Approach

I kept the fix at the logging boundary so clients can keep using the compatibility protocol without changes:

- Add shared query, request-URL and JSON helpers in `internal/logredact`.
- Use query redaction in both normal compatibility request logs and the seasons panic path. Match credential aliases case-insensitively after decoding, handle duplicate keys, and omit malformed queries instead of falling back to the original input.
- Remove URL userinfo and fragments, redact credential query parameters, and preserve ordinary parameters. Apply this to debug request URLs and URL strings in JSON, including relative stream URLs.
- Recursively mask credential-bearing JSON fields, including Jellyfin's `Pw` alias. Preserve exact JSON numbers. Omit incomplete, scalar and non-JSON bodies rather than writing an unsafe raw-text fallback.
- Capture request bytes as the handler reads them, up to the existing 256 KiB limit. The old debug middleware replaced the request body with that limited prefix, so a larger request was truncated before reaching its handler. The new wrapper forwards the full stream and its read/close results.

This changes diagnostic output, not live authentication, request parameters, response bodies, route registration or API schemas. The contribution is based on upstream `main` at `dc887b3c567303ca5554e3a1101d2dd6324345a4`; it contains no deployment settings or unrelated fork changes.

### Maintainer review follow-up

Revision `868d856d` keeps on-disk file paths verbatim in debug JSON. Before this change, any string starting with `/` was re-rendered as a URL, so `Path` values gained percent-encoding or became `[invalid URL]`. Strings are now rewritten only when they contain `?`, `#` or `@`, the only places a credential can appear. Redaction coverage is unchanged. `TestSanitizeJSONPreservesFilePaths` covers the case, and upstream CI passed on this revision.

AI disclosure for this follow-up: written by Claude Opus 5.5 (`claude-opus-5-5`) in Claude Code during maintainer review.

### CodeRabbit follow-up

Current contribution revision: `ad4eb4df`.

I addressed the two diagnostic-quality findings from CodeRabbit:

- Ordinary metadata containing question marks or embedded links stays unchanged. URL detection requires a rooted path or a valid scheme prefix. I did not use the proposed whitespace exclusion, because a URL with spaces can still carry a credential. Regression cases cover plain text, embedded links, root-relative and absolute URLs with spaces, mixed-case schemes and malformed URLs.
- Request capture records whether reads actually exceeded the 256 KiB capture budget. Only then does it report truncation. Tests cover one byte below, exactly at and one byte above the limit with an unknown content length, and confirm the handler receives the complete request.

`TestSanitizeJSONPreservesOrdinaryText` and the oversized case of `TestDebugLoggerReportsActualRequestTruncation` failed on the previous PR revision and pass with this follow-up. The full focused redaction/debug test selection and targeted vet passed. Upstream CI for this new revision is pending.

The original deployment evidence below applies to the initial fix, not this follow-up. This follow-up has not been deployed or merged into the fork's production `main`. The docstring coverage warning is not being addressed with boilerplate comments on test functions; the new production helper has a contract comment.

### Before/after evidence

The initial four middleware regressions were run before implementation on `08691635c6c315f091e224603271036071ee97db`. They reproduced query leaks, panic-path leaks, debug URL/body leaks, unsafe raw-body fallback and request truncation. The affected logging implementation at that baseline matches the upstream base. These tests pass with the fix.

The same five changed files were deployed in fork revision `8d5b5c34ff050c1039ccead5e68c890e553b3dc1`. This is separate from the upstream contribution revision `88ca57e7afb6df86bd5a07af53adecbf9c670f79`; file comparisons confirmed they contained the same initial fix before the CodeRabbit follow-up.

Live verification used deliberately invalid synthetic tokens, never a real credential. A GET to `/Users/Me` returned 401 before and after deployment. The following are verifier counters, not reconstructed raw log lines. Private request logs are deliberately excluded.

To repeat the query check, replace the reserved example host with a test instance, send the request below, and inspect only the request-log entry carrying `codex_log_probe=redaction-check`. The expected response is 401 in both versions. The old logger retains the synthetic token; the patched logger replaces its value with `[REDACTED]` while keeping `Limit=20`. Repeat with the aliases listed below. The example token is deliberately non-functional.

```sh
curl -s -o /dev/null -w '%{http_code}\n' \
  'https://media.example.invalid/Users/Me?ApiKey=synthetic-only-not-a-real-token&Limit=20&codex_log_probe=redaction-check'
```

Before:

```json
{"phase":"before","http_status":401,"matching_log_lines":1,"synthetic_secret_logged":true}
```

After, using `ApiKey`, `API_KEY`, `%41piKey` and `Api%5fKey` (all four requests returned 401):

```json
{"phase":"after","matching_log_lines":4,"synthetic_secret_logged":false,"redaction_present":true,"safe_parameter_retained":true}
```

After deployment, both health checks returned 200 and the container was healthy with no restarts. Subsequent read-only checks found an active, unpaused playback session updated within four seconds and 30 successful `POST /Sessions/Playing/Progress` responses with status 204 in a five-minute sample. That demonstrates continuing server-side playback reporting, not a fresh-start, seek or client-rendering test.

### Regression coverage

- `TestRequestLoggerRedactsQueryCredentials` and `TestSeasonsPanicLoggerRedactsQueryCredentials`: secret removal at the actual sinks, retained safe parameters, and unchanged live requests/statuses.
- `TestDebugLoggerRedactsCredentialsAndPreservesTraffic`: passwords, nested credential fields, access tokens and stream URLs are masked while the handler and client receive unchanged bodies.
- `TestDebugLoggerOmitsUnsafeBodiesWithoutTruncatingRequests`: malformed/form bodies and oversized requests do not leak through the raw fallback or truncate the handler's input.
- `TestDebugRequestBodyPreservesReadAndCloseErrors` and `TestDebugLoggerFilterLeavesRequestUnread`: original I/O results and the user-agent filter remain intact.
- The `TestSanitize*` cases cover aliases, percent-encoded and duplicate query keys, invalid encodings, URL userinfo/fragments, escaped/duplicate JSON keys, arrays, exact large numbers and unsafe body representations.

### Commands and CI

Initial-fix evidence, before the CodeRabbit follow-up:

On contribution revision `88ca57e7afb6df86bd5a07af53adecbf9c670f79`:

```sh
git diff upstream/main HEAD --check
GOMAXPROCS=2 go test -p 1 ./internal/logredact ./internal/jellycompat -run 'Test(Sanitize|RequestLoggerRedacts|SeasonsPanicLoggerRedacts|Debug)' -count=1 -timeout=2m
GOMAXPROCS=2 go vet -p 1 ./internal/logredact ./internal/jellycompat
```

All three passed. [Full CI for this exact contribution revision](https://github.com/blurbery/silo-server/actions/runs/34295443197) also passed: Go, Web and Docs hygiene are green. Upstream-owned PR checks will run separately after publication; those are not being reported as passed in advance.

The full gate runs `make embed-stub`, `go build ./...`, `gofmt -l .`, `go vet ./...`, changed-line golangci-lint, `make test-go`, settings-binding and playback-fixture checks, plus web lint/format/build/tests and docs-path hygiene. The fork-hosted dispatch compares lint against its `origin/main`; the upstream PR check will compare against the upstream target. Neither is described as a full-tree lint cleanup.

Additional evidence for the deployed fork revision, not a substitute for this PR's revision checks: [Go/Web/docs and PostgreSQL integration passed](https://github.com/blurbery/silo-server/actions/runs/34292481849), and [both container architectures built successfully](https://github.com/blurbery/silo-server/actions/runs/34293561415). The earlier development revision failed four lint checks; those corrections are included in this PR. Debug logging stayed disabled on production, so its enabled-mode evidence is from regression tests.

### API v2 / Huma compatibility

I checked the Huma foundation and pilot work (#936 and #941), plus the recent playback work in #1002 and #1005. The [API v2 contract](https://github.com/Silo-Server/silo-server/blob/26661d76f451ed790cf74221538b1048a8d193d6/docs/architecture/api-contract.md#L1060) explicitly keeps Jellyfin compatibility outside the native v2 contract. Huma is mounted through `humachi` on the native API listener; this fix stays in compatibility logging and standard-library redaction helpers.

At `apiv2` revision `26661d76f451ed790cf74221538b1048a8d193d6`, the existing affected logging files and `internal/logredact` are identical to the upstream base. Applying the complete contribution diff to a temporary index populated from that revision passed `git apply --cached --check` without conflicts. The reviewed playback PRs did not change this patch's logging files.

The compatibility claim is therefore source and patch compatibility with the inspected v2 branch, not a claim that the full Huma runtime or migration suite was executed. No v2 operations, OpenAPI artifacts, client bindings or playback contracts change. Apple and Android changes are not required for this logging-only patch. Recheck integration when the v2 branch moves.

## Benchmarks

Not run. Query parsing and JSON redaction add work to the logging path; I am not claiming zero overhead or a performance improvement. The debug capture remains bounded at 256 KiB. Latency, throughput and allocation effects have not been measured.

## Risks

- Diagnostic query ordering/encoding and JSON field ordering can change. Non-JSON, malformed and incomplete body captures are intentionally replaced by omission markers; raw XML, forms and text will no longer be dumped verbatim.
- Redaction follows recognised credential fields and URL representations. This is not a general guarantee that arbitrary free-text messages or every other logger in the repository contain no sensitive information.
- Existing logs are not rewritten, and exposed credentials are not automatically revoked. Operators should identify and revoke affected sessions/keys separately and review log retention and access.
- No database migration, settings change, authentication change or production deployment is part of this upstream PR.

## Checklist

- [x] I read and can explain the complete diff. Agent review is complete; human maintainer review is requested.
- [x] This pull request addresses one concern: credential-safe compatibility diagnostics, including preserving requests while capturing them.

## AI Disclosure

AI-assisted with Astra. I directed the task and designed the work.

- Harness: OpenAI Codex desktop.
- Tool(s): Codex Security plugin, CodeRabbit (external review suggestions), Git, GitHub CLI, Go tooling and SSH for authorised deployment verification.
- Model(s): Astra (`gpt-6-astra`), confirmed for this task. CodeRabbit did not report its underlying model identifier.
- Involvement: AI-assisted implementation, tests, review, verification and PR drafting under my direction. No claim of independent human code review.
- Adversarial review: a same-agent review traced the query, panic, debug URL and body sinks; checked encoded/duplicate aliases, malformed input and nested JSON; and checked ordinary request/response bytes, request-length limits, I/O errors and filtering. The debug request-truncation defect was addressed and covered by regression tests. Separate workers were not used, so this is not an independent multi-agent review. All reported patch lint failures were corrected and revalidated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive credentials are now redacted from diagnostic logs, including request URLs, query strings, request bodies, response bodies, and panic-recovery messages.
  * Malformed, oversized, or unsafe diagnostic content is omitted rather than logged.
  * Truncated request-body captures are clearly marked.
  * Traffic remains unaffected while debug logging captures sanitized information.

* **Tests**
  * Added coverage for credential redaction, safe handling of malformed content, truncation notices, body-read errors, and filtered requests.

### DIFF
--- a/internal/logredact/diagnostic.go
+++ b/internal/logredact/diagnostic.go
@@ -89,9 +89,33 @@ func sanitizeDiagnosticValue(value any) any {
 	case string:
 		// Playback responses contain token-bearing relative and absolute URLs
 		// under ordinary keys such as DirectStreamUrl and MediaSources[].Path.
-		if strings.Contains(v, "?") || strings.Contains(v, "://") {
+		if looksLikeDiagnosticURL(v) {
 			return SanitizeRequestURL(v)
 		}
 	}
 	return value
+}
+
+// looksLikeDiagnosticURL matches rooted paths and scheme-prefixed URLs,
+// not a question mark or an embedded link in ordinary metadata. Spaces do not
+// disqualify a URL: its query can still contain a credential needing redaction.
+func looksLikeDiagnosticURL(value string) bool {
+	value = strings.TrimSpace(value)
+	if strings.HasPrefix(value, "/") {
+		return true
+	}
+	scheme, _, found := strings.Cut(value, "://")
+	if !found || scheme == "" {
+		return false
+	}
+	for i, c := range scheme {
+		if c >= 'a' && c <= 'z' || c >= 'A' && c <= 'Z' {
+			continue
+		}
+		if i > 0 && (c >= '0' && c <= '9' || c == '+' || c == '-' || c == '.') {
+			continue
+		}
+		return false
+	}
+	return true
 }
--- a/internal/logredact/diagnostic.go
+++ b/internal/logredact/diagnostic.go
@@ -1,0 +1,97 @@
+package logredact
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/url"
+	"strings"
+)
+
+// SanitizeQuery preserves diagnostic parameters while masking credentials.
+// Parse errors omit the entire query: ParseQuery can otherwise return a partial
+// result and tempt callers to fall back to the unsafe original input.
+func SanitizeQuery(raw string) string {
+	query, err := url.ParseQuery(raw)
+	if err != nil {
+		return "[query omitted: invalid encoding]"
+	}
+	for key := range query {
+		if diagnosticSecretKey(key) {
+			query[key] = []string{Placeholder}
+		}
+	}
+	return query.Encode()
+}
+
+// SanitizeRequestURL accepts relative request targets as well as absolute URLs.
+// Unlike SanitizeURL (transport error diagnostics), it retains safe query fields.
+// The caller's URL and live request are never modified.
+func SanitizeRequestURL(raw string) string {
+	u, err := url.Parse(raw)
+	if err != nil || u.Opaque != "" {
+		return invalidURLPlaceholder
+	}
+	u.User = nil
+	u.RawQuery = SanitizeQuery(u.RawQuery)
+	u.Fragment = ""
+	u.RawFragment = ""
+	return u.String()
+}
+
+// SanitizeJSON keeps structured diagnostics, including exact JSON numbers, but
+// masks credential fields and credentials in URL strings. Invalid, truncated,
+// scalar and non-JSON bodies are omitted rather than logged as raw text.
+func SanitizeJSON(body []byte) []byte {
+	const omitted = "[body omitted: not a complete JSON object or array]"
+	if !json.Valid(body) {
+		return []byte(omitted)
+	}
+	decoder := json.NewDecoder(bytes.NewReader(body))
+	decoder.UseNumber()
+	var value any
+	if err := decoder.Decode(&value); err != nil {
+		return []byte(omitted)
+	}
+	switch value.(type) {
+	case map[string]any, []any:
+	default:
+		return []byte(omitted)
+	}
+	result, err := json.MarshalIndent(sanitizeDiagnosticValue(value), "", "  ")
+	if err != nil {
+		return []byte(omitted)
+	}
+	return result
+}
+
+func diagnosticSecretKey(key string) bool {
+	// Pw is the Jellyfin login alias. Normalisation covers the casing and
+	// separators used by compatibility clients without changing auth parsing.
+	key = strings.ToLower(strings.TrimSpace(key))
+	key = strings.ReplaceAll(strings.ReplaceAll(key, "_", ""), "-", "")
+	return key == "pw" || SecretKey(key)
+}
+
+func sanitizeDiagnosticValue(value any) any {
+	switch v := value.(type) {
+	case map[string]any:
+		for key, child := range v {
+			if diagnosticSecretKey(key) {
+				v[key] = Placeholder
+			} else {
+				v[key] = sanitizeDiagnosticValue(child)
+			}
+		}
+	case []any:
+		for i, child := range v {
+			v[i] = sanitizeDiagnosticValue(child)
+		}
+	case string:
+		// Playback responses contain token-bearing relative and absolute URLs
+		// under ordinary keys such as DirectStreamUrl and MediaSources[].Path.
+		if strings.Contains(v, "?") || strings.Contains(v, "://") {
+			return SanitizeRequestURL(v)
+		}
+	}
+	return value
+}
--- a/internal/jellycompat/handlers_items.go
+++ b/internal/jellycompat/handlers_items.go
@@ -21,6 +21,7 @@ import (
 	"github.com/Silo-Server/silo-server/internal/access"
 	"github.com/Silo-Server/silo-server/internal/catalog"
 	"github.com/Silo-Server/silo-server/internal/config"
+	"github.com/Silo-Server/silo-server/internal/logredact"
 	"github.com/Silo-Server/silo-server/internal/models"
 	"github.com/Silo-Server/silo-server/internal/playback"
 	"github.com/Silo-Server/silo-server/internal/recommendations"
@@ -1457,7 +1458,7 @@ func (h *ItemsHandler) HandleSeasons(w http.ResponseWriter, r *http.Request) {
 			slog.ErrorContext(r.Context(), "jellycompat HandleSeasons panic", "component", "jellycompat",
 				"error", fmt.Sprint(rv),
 				"path", r.URL.Path,
-				"query", r.URL.RawQuery,
+				"query", logredact.SanitizeQuery(r.URL.RawQuery),
 				"stack", string(debug.Stack()),
 			)
 			writeError(w, http.StatusInternalServerError, "ServerError", "Internal error")
--- a/internal/jellycompat/logging.go
+++ b/internal/jellycompat/logging.go
@@ -233,6 +233,9 @@ func newDebugLogMiddleware(logFile io.Writer, userAgentFilter string) func(http.
 
 			if requestCapture != nil && requestCapture.body.Len() > 0 {
 				_, _ = fmt.Fprintf(logFile, "Request Body (%d bytes captured):\n", requestCapture.body.Len())
+				if requestCapture.truncated {
+					_, _ = fmt.Fprintf(logFile, "[truncated at %d bytes]\n", debugMaxBodyCapture)
+				}
 				writeIndentedJSON(logFile, requestCapture.body.Bytes())
 			}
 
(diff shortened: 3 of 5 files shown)


internal/logredact/diagnostic.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+	case []any:
+		for i, child := range v {
+			v[i] = sanitizeDiagnosticValue(child)
+		}
+	case string:
+		// Playback responses contain token-bearing relative and absolute URLs
+		// under ordinary keys such as DirectStreamUrl and MediaSources[].Path.
+		if looksLikeDiagnosticURL(v) {
```
Quick104: **Minor, debug-log only: on-disk file paths are rewritten as URLs.**

The rooted-path rule sends every string that starts with `/` through `url.Parse`/`String()`. jellycompat `Path` fields hold real file paths (`compatMediaPath` returns `FilePath`), so a debug capture of `/Items` or `PlaybackInfo` now shows:

| Value | Logged as |
|---|---|
| `/media/Movies/Foo Bar (2020)/Foo Bar (2020).mkv` | `/media/Movies/Foo%20Bar%20%282020%29/Foo%20Bar%20%282020%29.mkv` |
| `/media/Movies/100% Wolf (2020)/100% Wolf.mkv` | `[invalid URL]` |
| `/media/TV/Show #1/S01E01.mkv` | `/media/TV/Show%20` |

Non-ASCII names get percent-encoded too. File paths are usually what you open this log to check, so they shouldn't be distorted.

Fix I'm pushing: a string can only carry a credential through a query (`?`), fragment (`#`), or userinfo (`@`), and `SanitizeRequestURL` only touches those parts. So a value with none of those characters is left exactly as it was. Redaction coverage doesn't change. A regression test covers the three paths above plus a non-ASCII one.

Quick104 (commented):
Review of `ad4eb4df`. One finding, inline. I'm pushing the fix to this branch, so there's nothing for you to change.

_Posted by Claude Code (`claude-opus-5-5`) on behalf of a maintainer._

internal/logredact/diagnostic.go
Quick104: Fixed in https://github.com/Silo-Server/silo-server/commit/868d856dbfc7968a35b0ff2a0b350dcf6ddac709. `sanitizeDiagnosticValue` now rewrites a URL-looking string only if it contains `?`, `#` or `@`. Anything else is kept verbatim, and redaction coverage stays the same.

A correction to my comment above: a file path that contains `#` or `?` still goes through `SanitizeRequestURL`, which drops the fragment or query to stay on the safe side. That trade-off is intentional. The new `TestSanitizeJSONPreservesFilePaths` covers paths with spaces, a literal `%`, and non-ASCII names. Verified with `go test ./internal/logredact ./internal/jellycompat -count=1`, `go vet`, and `gofmt`.

Quick104:
## Review summary

Reviewed head: `868d856d` (the fix pushed on top of `ad4eb4df`).

**Fixed:** debug JSON rewrote on-disk file paths as URLs, so `Path` values gained percent-encoding, lost everything after `#`, or became `[invalid URL]`. Fixed in 868d856dbfc7968a35b0ff2a0b350dcf6ddac709 with a regression test. The thread is resolved.

**Checked, no issue found:**
- The request logger and the `HandleSeasons` panic logger were the only jellycompat sinks writing the raw query.
- Stream URLs carry credentials only as `api_key`, `ApiKey` or `token`, and all of them are redacted.
- The native API request logger does not log the query.
- `debugRequestBody` passes the full body through. `net/http` drains its own reference to the body after the handler returns, not the wrapper, so capture has no race.

**Remaining, not blocking:** jellycompat request logs still write the raw `path`. No jellycompat route carries a token in the path today, so nothing leaks. The native logger uses `activitylog.RedactSecretPathParams` for this, and it would be worth reusing if a jellycompat token path is ever added.

**Validation:** `go test ./internal/logredact ./internal/jellycompat`, `go vet` and `gofmt` passed locally. Upstream Go test, Go lint, Go contract checks and Web passed on `868d856d`.

**Mergeability: 9/10, ready.** The fix sits at the logging boundary and leaves auth and the API contract untouched. It fails closed on malformed input, and the tests exercise the real sinks.

_Review by Claude Code (`claude-opus-5-5`) on behalf of a maintainer._

PR #1076, opened 2026-09-13
fix(transcode): preserve live session on spawn failure

Quick104:
## Problem

A duplicate `/transcode/start` could tear down a healthy session before replacement startup finished, turning a transient spawn failure into an outage.

## Approach

Start and validate process creation first; only then retire the existing session and its segment directory.

## Validation

## Risks

None identified.

## Checklist

## AI Disclosure

- Tool(s): functions.exec, gh
- Model(s): GPT-6
- Adversarial review: Muse finding independently verified against source; fix scoped to spawn failure preservation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Existing transcoding sessions now remain available when starting a replacement session fails.
  - Previous sessions are retired only after the replacement starts successfully, preventing unnecessary disruption during transient errors.


Quick104:
Addressed the review feedback: replacements use an isolated generation directory, old-session retirement occurs after readiness, and staged directories are removed on all pre-registration failures.

PR #1077, opened 2026-09-13
fix(playback): replay finalized stop receipts

Quick104:
## Problem

A concurrent stop caller that lost the finalization claim could return before the winner stored the history ID, producing an incomplete receipt.

## Approach

When claim ownership is lost, re-read the durable stop row and return the stored receipt.

## Validation

## Risks

None identified.

## Checklist

## AI Disclosure

- Tool(s): functions.exec, gh
- Model(s): GPT-6
- Adversarial review: Muse finding independently verified against source; fix adds a durable receipt re-read on claim loss.


Quick104:
The loser now polls the durable receipt until finalization is recorded, with a bounded retry window.

Quick104:
Closing as superseded by current playback stop finalization, which rereads durable receipts and can take over after lease expiry. This PR's fixed three-second wait is weaker than the recovery path now in main.

PR #1078, opened 2026-09-13
fix(playback): retain progress side effect lock identity

Quick104:
## Problem

Deleting a per-session progress mutex while a writer still held it allowed a later writer to create a second mutex and run side effects concurrently. Reclaiming entries without coordinating acquisition could recreate the same race.

## Approach

Retain stable lock identity while writers are active, coordinate map acquisition and reclamation, and remove the entry after the final writer exits. Same-session transcode replacements publish into the canonical output directory so orphan cleanup protects active sessions.

## Validation

## Risks

The lock map is reclaimed after the last writer; replacement publication now renames the temporary directory after readiness and before registration.

## Checklist

## AI Disclosure

- Tool(s): functions.exec, gh
- Model(s): GPT-6
- Adversarial review: Lock reclamation and replacement-directory lifecycle findings were independently verified and fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback progress handling to prevent concurrent side effects for the same session.
  * Preserved session-level synchronization when late progress updates arrive.


Quick104:
The lock entry must remain stable while late writers can still hold it; deleting it immediately recreates the reported race. I am keeping this PR open while a lifecycle-safe refcounted entry is prepared. The missing regression test will be added with that change.

PR #1087, opened 2026-09-14
perf(metadata): index the columns the artwork GC reference check filters on

RXWatcher:
Artwork revision GC checks catalog artwork paths before deleting cached objects. On the author's large catalog, the unindexed batched reference check exceeded ten minutes. This migration adds six partial B-tree indexes on artwork paths in `media_items`, `episodes`, `people`, and `seasons`, excluding NULL values.

The indexes are created concurrently to allow catalog writes during the build. Each build is preceded by a concurrent drop of its named index. This lets Goose retry after an interrupted run leaves completed or invalid indexes behind. Without cleanup, retrying the previous revision failed with `SQLSTATE 42P07` and prevented server startup. Rollback drops all six indexes concurrently.

Localization and collection tables remain outside this optimization. Their small row counts in the author's deployment do not establish their size in other deployments. This changes database access costs without changing API contracts, Apple or Android behavior, or Jellyfin compatibility.

Validation:

- The author reported a prepared `text[]` reference check over 10,000 candidates taking 0.475–2.093 seconds across seven executions after indexing, versus cancellation at 10m09s before indexing. Reported cleanup passes held around 18.5 seconds, deleting about 38,000 objects per pass and continuing past pass 13. These deployment measurements were not repeated during this review.
- On isolated PostgreSQL 18, `TestArtworkReferenceIndexesRetryPostgres` failed on `1c870dfe8` with an already-existing index, then passed with the repair. It seeds a completed build and a failed concurrent build, verifies six usable indexes, and exercises rollback and fresh application while preserving catalog rows.
- The full application migration chain, rollback to the preceding version, and reapplication passed on an empty PostgreSQL 18 database.
- `go build ./...`, `go vet ./...`, `gofmt -l .`, and `golangci-lint run --new-from-merge-base=69389ca3d00a1fe6ff0ad44ab7cf3f9959080daa ./...` passed (lint: zero issues). Focused metadata and migration tests passed, as did the settings, playback, route, migration-ledger, scenario, offline, OpenAPI, contract, fixture, router-artifact, and local-path gates.
- Full Go tests failed in `TestAdminResourceCapabilitiesAndScope` (Linux-only capability expectation on macOS) and `TestResolveViewerScopeParity` (25 ms evaluation deadline). Both failures reproduced on base `69389ca3d`.
- Full web tests: 4,293 passed and six failed. The four `AdminUserDetail` failures also reproduced on the base. The other two failures were timeouts; both affected files passed on the base and on this revision in isolation (13 tests). The web tree is unchanged from the base.
- Frontend lint, formatting, typecheck, and production build passed.
- Review revision: `d7dd60995a3d2b1e50126e8c41bb81e1333cdecf`. Required GitHub Go and Web CI is still running; this is not a claim that the full gate is green.

Index storage and artwork-update overhead increase. Concurrent builds and drops can wait for existing transactions. Retrying rebuilds previously completed indexes too; a large deployment may need a longer `SILO_MIGRATE_TIMEOUT`.

Related issue: N/A — narrow fix. The separate object-deletion batching change is #1086.

AI-use disclosure: Original implementation and deployment measurements used Claude Code with `claude-opus-5[1m]`, as reported by the author. Adversarial review, migration retry repair, regression coverage, and this description update used Codex with `gpt-6-astra`. The review traced the GC query and startup migration path, checked prior bot findings against the current revision, and reproduced the remaining retry failure on PostgreSQL before fixing it. CodeRabbit, Greptile, and Macroscope also supplied automated PR analysis.


RXWatcher:
Correcting this PR: the first version indexed `media_items` only, and that was wrong.

I scoped it that way from `pg_stat_user_tables.n_live_tup`, which reported `episodes` at 451 rows and `people` at 264. Those numbers were stale. The real counts are `episodes` 2,206,674 and `people` 1,219,902, with `seasons` at 138,952 — so the reference check still sequentially scanned three large tables and GC passes still stalled after the first version was applied.

`4815c232b` adds the missing three. The description has been rewritten with the real row counts and with end-to-end numbers rather than a component measurement.

Two things worth flagging for anyone reading the history here:

- My earlier "10 minutes to 313ms" figure was measured with the paths supplied as a subquery, which the planner handles differently from the `text[]` parameter the GC actually passes. The numbers now in the description use the real parameterised shape and all twelve surfaces.
- The stall reproducing on the sixth pass twice in a row looked like the custom-to-generic plan switch. It is not — execution 6 is faster, not slower. It was the unindexed scans.

Re-verified on the updated branch:

```
$ go run ./cmd/silo/ --env ... --migrate-only
2026/09/14 15:03:44 INFO database migrations applied

$ psql -d gcidx_verify -tAc "select indexname from pg_indexes where indexname like 'idx_%_gc' order by 1"
idx_episodes_still_path_gc
idx_media_items_backdrop_path_gc
idx_media_items_logo_path_gc
idx_media_items_poster_path_gc
idx_people_photo_path_gc
idx_seasons_poster_path_gc

$ go run ./cmd/silo/ --env ... --migrate-down-to 20260913174919
2026/09/14 15:03:50 INFO database migrations rolled back kept_through_version=20260913174919

$ psql -d gcidx_verify -tAc "select count(*) from pg_indexes where indexname like 'idx_%_gc'"
0
```

RXWatcher:
Rebased onto `69389ca3d` to pick up the `internal/api/handlers` test fix — the Go check on the earlier push was failing on two tests unrelated to this change, both pre-existing on the base commit and failing identically on `main` itself.

No change to the migration in the rebase; the diff is still the single file.

RXWatcher:
Thanks for `1c870dfe8` and `d7dd60995` — you got there before I did, and your version is better than the one I had written.

I had a `DO` block that dropped only *invalid* leftovers and kept `IF NOT EXISTS`. That preserves already-built indexes on a retry, but the drop has to be an ordinary `DROP INDEX` to run inside the block, which takes a brief exclusive lock. Dropping each index CONCURRENTLY and removing `IF NOT EXISTS` avoids the lock entirely and makes a failed build surface as an error instead of being silently skipped. Rebuilding valid indexes on a retry is the right price for never blocking writes.

I have discarded mine and adopted yours.

Verified on a scratch database, for the record: the full chain applies with 6 valid indexes, and replaying the Up body with one index marked `indisvalid = false` in `pg_index` returns 6 valid, 0 invalid — so the recovery path does what it says.

PR #1088, opened 2026-09-14
perf(metadata): batch artwork revision cleanup at 10000 candidates

RXWatcher:
With #1086 and #1087 merged, this PR raises the artwork revision cleanup limit from 100 to 10,000 candidates per scheduled pass and preserves durable recovery for incomplete or canceled bulk deletions.

Deletion now locks the claimed registry rows and rechecks ownership, current manifests, and catalog references before sending keys to storage. Upload registration waits for those locks. Successful deletions retain the existing pending-heal path; errors and short deletion counts persist tombstones and schedule retries so partially deleted artwork cannot be parked as a live reference. A bounded commit context preserves those tombstones when the task is canceled during storage deletion.

The review reproduced two defects in the earlier bulk-delete implementation: deletion ran without the lock used by upload registration, and partial failures could park broken references. Both are fixed, with PostgreSQL regression coverage for upload locking, batch contents, concurrent retracking and references, partial errors, short counts, cancellation, and subsequent healing.

The reference-path indexes from #1087 and the shared batch processor from #1086 are now included through the merge of main. The earlier reported 17.5-second measurement for 10,000 candidates predates the locking correction; throughput and upload contention with the final implementation still need validation against a representative backlog. Registry locks are held through each bulk storage call, and database failure or process termination during the storage/commit window remains a recovery limitation.

No native API contract changes are involved. Apple, Android, web, and Jellyfin clients share the corrected artwork storage and healing behavior; no client changes are required.

Merge resolution at `5363c7d7fa751c320d0319588bc1ba57ac1ef44f` retains the shared batch processor and the durable retry/cancellation behavior. PostgreSQL GC integration tests passed with the race detector; focused vet, lint, formatting, and diff checks passed. CI for the merge commit is pending.

Earlier review validation at `93934a91bcf89114bb77daf7e84778548919d38a`:

- PostgreSQL GC integration tests passed with `-race`, including the new regressions. The upload-lock and partial-failure tests failed on the previous PR head.
- Formatting and diff checks, settings bindings, playback fixtures, route inventory, migration ledger, scenario catalogs, offline routes, OpenAPI, API contract compatibility, API fixtures, and the committed-router artifact test passed. Full Go build and vet passed before the final bounded-error follow-up; metadata build, vet, and GC race tests also passed on the final revision. Web lint passed with existing warnings, and web formatting passed.
- The broader PostgreSQL metadata/catalog run hit `TestSyncMergeStepsAfterPlexTableDropPostgres` and `TestOrphanCleanupAfterPlexTableDropPostgres` failures caused by missing `users.role` values in their fixtures. Both failures were reproduced on unchanged base `69389ca3d00a1fe6ff0ad44ab7cf3f9959080daa`.
- The full Go run reported a timeout in `TestHandleStartPlaybackV3DoesNotWaitForSessionSync`; the web run reported timeouts in `AdminSections.concurrency.test.tsx` and `AdminUserDetail.test.tsx`. These suites were stopped after those failures in unchanged code; they are not passing full-suite results. Full Go lint, web build, and CI remained pending at review handoff. No production throughput claim is made for the revised implementation.

AI-use disclosure: The original implementation and reported measurements were produced with Claude Code using `claude-opus-5[1m]`. Adversarial review, fixes, regression tests, and this description used OpenAI Codex desktop with `gpt-6-astra` and the Modern Go Guidelines CLI. The review traced registry ownership, upload registration, reference checks, storage errors, cancellation, and healing through the source and PostgreSQL tests; it also inspected prior bot findings.

> [!NOTE]
> ### Batch artwork revision cleanup at 10000 candidates in GC runner
> - Increases `metadata.artworkRevisionGCBatchSize` from 100 to 10,000 in [artwork_revision_gc.go](https://github.com/Silo-Server/silo-server/pull/1088/files#diff-b86d13c74fd4e1bc1f2feb83b453fc922ea45ff398afae9ba0788ee0ed0e9168) to allow larger GC batches.
> - Changes `metadata.ArtworkRevisionGarbageCollector.processCandidatesToHeal` to a single batch deletion call and removes the per-candidate fallback.
> - On batch failure or short deletion count, the runner persists tombstones for all attempted IDs using a 5-second context, schedules retries, and returns a joined error.
> - Adds tests for partial deletion retention, row locking during batch delete, and ownership rechecking.
> - Risk: Removing the per-manifest fallback in `processCandidatesToHeal` means a single storage error now aborts the remaining deletions in the batch and relies on retry scheduling.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5363c7d.</sup>
> 
>


RXWatcher:
Flagging a problem with this PR that only became visible after @Quick104's rework landed on #1086, and it argues against merging this as written.

That rework holds one transaction open across `DeleteObjects`, including a per-manifest retry loop that issues one storage call per candidate when a batch comes back short. At the current batch of 100 that bounds the worst case at roughly 100 sequential calls. **This PR raises the batch to 10000**, which would make that worst case up to 10,000 sequential storage calls inside a single open transaction — row locks held and a snapshot pinned throughout.

On the deployment I measured, B2 bills ~2.9s per `DeleteObjects` call regardless of how many keys it carries. 10,000 × 2.9s is not a rounding error; it is a multi-hour transaction during a storage incident, which on Postgres also blocks vacuum and bloats the table it is holding.

So the measurement in this PR's description is still accurate — per-candidate database work really is 0.17ms, and the batch really can be far larger than 100 on the happy path — but the safe ceiling is not set by throughput. It is set by the degraded path, and I had not accounted for that.

Three ways forward, and I have no attachment to this PR:

1. **Close it.** The two changes that actually made the GC drain are #1087 (indexes) and the batched delete itself. The batch size is a much smaller win than either and is not worth a correctness risk.
2. **A smaller increase** — 500 or 1000 — which keeps the degraded path bounded at minutes rather than hours while still giving 5–10x the throughput.
3. **Move the retry loop outside the transaction**: commit the successful deletions, leave the failures for the next run. That keeps the transaction short and makes a large batch safe, but it is a change to @Quick104's design and theirs is the one I would defer to.

My preference is (1) or (2). Happy to rework or close on your call.

RXWatcher:
Withdrawing my previous comment — `a41e4b54b` already fixes the problem I raised, and I should have checked the branch before writing it.

My concern was that the per-manifest retry loop ran inside the open transaction, so at a 10000 batch a storage incident could hold one transaction open for hours with row locks held and a snapshot pinned. That loop is gone: the batch path now makes a single `DeleteObjects` call, records the error, writes tombstones for every attempted revision, and commits under its own bounded context.

What remains inside the transaction is the internal 1000-key chunking, so ~40 sequential requests for a 10000-candidate batch. Measured on the deployment these numbers come from, 40,000 keys delete in ~8.5s — seconds, not hours, and comfortably inside the 15-minute lease. So the batch increase is safe as it stands and I no longer think this should be closed or reduced.

Two things I got wrong that are worth recording:

- I sized the batch from the happy path only. The measurements in the description are accurate, but the ceiling is set by the degraded path, and I had not looked at it.
- I wrote that comment against the branch as I had last pushed it rather than as it actually stood. You had already fixed it.

Giving the tombstone commit `context.WithoutCancel` plus a 5s timeout is a nice detail — a cancelled run still records what storage already removed, which is exactly the case where losing the tombstone would strand a broken reference.

I have taken all three commits into my downstream fork.

PR #1092, opened 2026-09-14
feat(metadata): sweep artwork objects the catalog no longer references

RXWatcher:
## Problem

The artwork revision GC collects revisions it was told about — rows in `artwork_revision_gc_candidates`. A superseded revision whose candidate row was never enqueued is invisible to it forever, and nothing else in the server ever reads the bucket back to notice. There is no path by which those objects are reclaimed; they stay until someone deletes them by hand.

`s3client.ListObjects` has exactly one caller (`DeleteByPrefix`). `reconcile_artwork_cache` runs DB→storage, repairing records whose objects went missing. Nothing runs storage→DB.

I found this measuring why one deployment's artwork bucket was several times the size of another holding a near-identical catalog:

| | objects | size | orphaned | |
|---|---|---|---|---|
| deployment A | 10,388,529 | 652.8 GB | 2,244,601 | **178 GB (27%)** |
| deployment B | 2,493,834 | 204.0 GB | 514,821 | **34 GB (17%)** |

Per category on A: backdrops 47.4% orphaned, logos 48.3%, posters 31.6%, stills 8.7%, profiles 0.1%. Sampled directories held 2–3 distinct content hashes where only one was referenced.

## Solution

A scheduled storage-side sweep, as a backstop to the GC rather than a replacement. The GC still collects promptly on displacement; this catches what it never learned about.

Per page of 1,000 objects:

1. `ListObjectInfosPage` from the saved continuation token
2. Rebuild the path each object would be referenced by, and check it against the catalog
3. One batched `DeleteObjects` for the remainder

Step 2 relies on one property worth stating explicitly: **every cached artwork column stores the `original` variant of its ladder.** Verified across 2,139,166 cached rows on deployment A, where `original` was the only variant present. So an object at `<dir>/<variant>.<hash>.<ext>` is referenced exactly when the catalog holds `<dir>/original.<hash>.<ext>`. Reconstructing that turns the check into an indexed equality lookup rather than a scan, and it reuses `artworkReferenceUnionSQL` so the sweep can never take a narrower view of "referenced" than the GC does.

Cost is roughly three API calls per 1,000 objects, against storage that charges per call.

## Safety

Each of these exists because of a specific way this could go wrong:

- **Age floor (24h).** Artwork cached while the sweep is running has not necessarily been committed to a catalog row yet. Without this, the sweep deletes objects moments after they are written. Objects under the floor never reach the reference check at all.
- **Anomaly guard (>80% of a full page unreferenced → stop, don't delete).** Every way the reference check can break — a swallowed query error, a schema change that empties a surface, a key-prefix mismatch that makes every reconstructed path miss — presents identically: a page where almost nothing looks referenced. Deleting on that signal would erase the artwork library. The threshold sits well above the 52% worst-case category rate measured in practice. Short trailing pages are exempt, since a final page may legitimately be entirely unreferenced.
- **Unrecognized keys are counted and skipped, never deleted.** An unmodelled key shape is the one case where deleting could destroy something this code does not understand. (0 unparsable in 12.9M objects observed across both deployments.)
- **Checkpoint scoped to storage identity**, reusing the existing `ArtworkStorageIdentity` fingerprint — a continuation token from one bucket is meaningless in another, so a bucket move restarts rather than resuming into nowhere.
- **Only `local/`, `tmdb/`, `tvdb/` are swept.** The bucket also holds namespaces this task does not model (book-provider caches, diagnostics, collection images). Sweeping those would be deleting on an assumption.

A partial delete logs and moves on: the objects stay unreferenced, so the next pass retries them. Nothing is lost and there is nothing to repair.

## Validation

```
$ gofmt -l internal/metadata/artwork_storage_sweep.go internal/taskmanager/tasks/sweep_artwork_storage.go cmd/silo/main.go
$ go vet ./internal/metadata/ ./internal/taskmanager/tasks/
$ go build ./internal/... ./cmd/...
$ go test -count=1 ./internal/metadata/ ./internal/taskmanager/tasks/
ok  	github.com/Silo-Server/silo-server/internal/metadata	3.442s
ok  	github.com/Silo-Server/silo-server/internal/taskmanager/tasks	0.973s
```

Six unit tests cover the key-parsing contract (including that `original` maps to itself — the case where getting it wrong would make every referenced object look orphaned) and assert the anomaly-guard thresholds directly against the real measured rates.

**Operationally validated ahead of the code**: the same algorithm, run as a one-off script against both deployments, deleted 2,244,601 and 514,821 objects with 0 errors. Re-enumerating and re-classifying deployment A afterwards returned **0 orphans across all five categories**, 652.8 GB → 474.4 GB. Before deleting anything I verified 99.43% of referenced paths had their objects present, so the DB↔storage mapping was coherent and the classification trustworthy.

One caveat from that exercise, since it argues for the age floor: identical image content is sometimes stored under multiple paths (a `local/…` copy left behind by a path migration). Verifying an object by hash alone gives false positives; the check must be directory **and** hash, which is what this does.

## Risks

Deletes objects, so the guards above are the substance of the change. It is additive — one new task plus its registration — and takes no path that existing code takes today. Default schedule is daily; the sweep only reclaims space, so running it more often buys nothing.

## AI-use disclosure

Written with Claude Code (agent harness) using model `claude-opus-5[1m]`. No other AI tooling. All figures above are measured, not estimated.

https://claude.ai/code/session_013LPxFvVsa6TUFv1dPvduoK

> [!NOTE]
> ### Add scheduled `ArtworkStorageSweeper` to delete unreferenced artwork objects
> - Adds `ArtworkStorageSweeper` which walks storage prefixes page-by-page, parses artwork ladder keys to reconstruct original-variant paths, and deletes objects whose originals are no longer catalog-referenced
> - Adds `SweepArtworkStorageTask` with a daily schedule, persisted checkpoint across three owned prefixes, and per-run page limit. Checkpoint advances after normal progress and stays put on lock-skipped or anomaly-stopped runs
> - Server bootstrap in [main.go](https://github.com/Silo-Server/silo-server/pull/1092/files#diff-c8a7988838b46a61cad8d450a07cf1f77898988683073e4e5636c1f2452b5c4c) registers the task when both database pool and public storage lister are available
> - Objects younger than 24 hours or missing a modification timestamp are skipped; an anomaly guard stops a run when a full page of ≥50 objects is >80% unreferenced
> - Risk: `DeleteObjects` is called on storage keys whose reconstructed original paths are absent from the catalog reference union; incorrect `parseArtworkObjectKey` reconstruction or stale catalog state could delete live artwork. The anomaly guard and age floor mitigate but do not eliminate this. A shared advisory lock coordinates cross-node runs; no-pool configurations skip locking and run unlocked
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57ecb90.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added an automated background task that reclaims unreferenced artwork storage.
  * Sweeps storage incrementally, resumes from saved progress, and retries incomplete deletions.
  * Protects against anomalous cleanup results and reports detailed progress and outcomes.
  * Runs on a daily schedule when storage cleanup is configured.


RXWatcher:
Pushed `45a837e6e` addressing all three findings.

**1. Missing timestamps bypassed the age guard — real bug, fixed.** `modified != nil && modified.After(cutoff)` skipped the guard entirely for objects with no reported modification time, so exactly the objects whose age could not be established were the ones the age floor did not protect. It now fails closed: unknown age is treated as too new. Skipping costs a little unreclaimed space; guessing costs freshly cached artwork.

**2. Nodes raced on the checkpoint — fixed.** You are right that this matters here: Silo deploys as a cluster, every node runs its own triggers, and the checkpoint is a single shared settings row. The sweep now takes the same advisory lock `PgExecutionRepository.Prune` uses for history cleanup. A node that cannot acquire it reports the run skipped and — the part that actually matters — leaves the checkpoint untouched, since writing a stale cursor back would drag the sweeping node's progress backwards.

Worth being precise about the severity: deletions are idempotent and each node computes its own reference set, so concurrent sweeps would not have deleted anything wrongly. What they would have done is fail to make progress, with two nodes overwriting each other's cursor and a prefix potentially staying unswept indefinitely. Progress, not correctness.

**3. Safeguards were not exercised — fixed.** Fair criticism; the thresholds were asserted arithmetically, which tests the constants rather than the code. There is now a fake storage backend and the destructive path is tested directly:

- only unreferenced objects are deleted, referenced siblings survive
- objects under the age floor are skipped and never reach the reference check
- an object with no timestamp fails closed
- an anomalous page stops the sweep with zero deletions
- pagination resumes from the right token and honours maxPages
- a sweeper without a pool still runs rather than tripping over `pglock`'s nil-pool behaviour

```
--- PASS: TestSweepDeletesOnlyUnreferencedObjects
--- PASS: TestSweepSkipsObjectsUnderTheAgeFloor
--- PASS: TestSweepFailsClosedOnMissingTimestamp
--- PASS: TestSweepStopsInsteadOfDeletingAnAnomalousPage
--- PASS: TestSweepResumesAcrossPagesAndStopsAtMaxPages
--- PASS: TestSweepWithoutAPoolSkipsClusterLocking
ok  	github.com/Silo-Server/silo-server/internal/metadata	3.418s
ok  	github.com/Silo-Server/silo-server/internal/taskmanager/tasks	0.943s
```

Additional evidence since opening this: the task has been running on a live deployment. Its first scheduled run scanned 200,000 objects in 45.6s and reported 200,000 referenced, 0 deleted, 0 unparsable, no anomaly — the correct result against a bucket independently verified to have no orphans, and a useful signal precisely because a wrong implementation would have shown up as deletions there.

Issue #1032, opened 2026-09-11
Catalog item API ignores library_id when returning media versions


Quick104:
**Discord:** https://discord.com/channels/1487838737248292864/1545800869205770310
**Author:** srbgod

## Goal
Keep item details and media versions scoped to the library being browsed.

## Steps
1. Create two movie libraries containing different files for the same movie (same canonical content ID).
2. Use a profile that can access both libraries.
3. Browse the second library and open the shared movie.
4. Inspect GET /api/v1/catalog/items/{id}?library_id={second_library_id}
5. Check returned versions/media files and Media locations in the web UI.

## Expected
When library_id is supplied, the API response should contain only media files belonging to that library.

## Actual
The API returns media files from both accessible libraries. The web UI displays versions/locations from the other library.

## What is slow/broken
Library scoping in the catalog item API and item details in the web UI.

## Scope
Consistent when the same canonical movie exists in multiple libraries accessible to the profile. Catalog membership per library is correct; problem is files returned for an individual item.

## Version/branch
Official upstream main, commit 3131494e52e069fc1ac10e1a31503c797d09f26e, image build-535.

## Deployment
ghcr.io/silo-server/silo-server:latest

## Technical notes
Web client preserves selected library and sends library_id correctly. Backend parses into AccessFilter.PresentationLibraryID and validates item membership, but FilterMediaFilesByAccess does not apply PresentationLibraryID when filtering the item's files.

PR #1109, opened 2026-09-16
fix(catalog): scope media versions to the requested library

raiden202:
## Problem

Related issue: #1032

`GET /api/v2/catalog/items/{id}?library_id=...` (and the v1 route the web client uses) returned media files from every library the profile could access, not just the one being browsed. When the same movie exists in two movie libraries, the detail page showed both versions and both media locations regardless of which library the viewer opened it from. The handler parsed `library_id` into `AccessFilter.PresentationLibraryID` and validated membership, but `FilterMediaFilesByAccess` never applied it.

## Approach

Apply the presentation library scope inside `FilterMediaFilesByAccess`, including its unrestricted fast path, so every caller that supplies a presentation library gets narrowed results. Every current caller that sets it (catalog detail, jellycompat direct content, watch-together selection, playback) wants that narrowing; audiobooks never sets it. `FileAllowedByAccess` is unchanged, so direct file authorization still follows access policy and cross-library playback of an explicitly selected file keeps working. Requests without `library_id` behave exactly as before.

Fixing the one filter every path routes through was smaller and safer than patching the catalog handler alone, which would have left the other scoped callers inconsistent. No web or client contract changes are needed; the web client already sends `library_id`.

## Validation

- New `TestFilterMediaFilesByAccessPresentationLibrary` (seven cases: scoped and unscoped reads; the scope cannot bypass the allowlist, disabled libraries, quality ceiling, or a selected file outside the scope) fails on `main` and passes here.
- `go test` passes for `internal/catalog`, `internal/jellycompat`, `internal/watchtogether`, `internal/audiobooks`. `internal/apiv2` has two failures (`TestAdminResourceCapabilitiesAndScope`, `TestContractFixtures`) that fail identically on untouched `main`.
- `gofmt` clean. Local `golangci-lint` could not run against the Go 1.27 toolchain (typecheck errors in the standard library), so changed-line lint is left to CI.
- Manual check on a real deployment with one movie present in two movie libraries: before the fix, the item endpoint returned both files for every `library_id`; with this build it returns only the requested library's file for each `library_id` and both files when `library_id` is omitted.

## Risks

Callers that set `PresentationLibraryID` now receive fewer files when an item spans libraries. All current callers were read and want this. If an item is a member of the library but has no files there, the response has an empty version list, matching existing detail behavior. No migration or contract change.

## Checklist

## AI Disclosure

- Harness: Claude Code (orchestration) and Codex CLI (implementation and review)
- Tool(s): Claude Code, Codex CLI (`codex exec`)
- Model(s): claude-fable-5-1 (orchestration, verification, PR text); gpt-6-astra, reasoning effort xhigh (implementation); gpt-5.6-terra (independent review)
- Adversarial review: independent read-only Codex review of the full diff and every `FilterMediaFilesByAccess` caller, checking for callers that must not be narrowed, zero-file behavior after membership validation, the fast-path change, and test gaps. No findings.

> [!NOTE]
> ### Fix `catalog.FilterMediaFilesByAccess` to scope media versions to selected `PresentationLibraryID`
> - When a `PresentationLibraryID` is set, `FilterMediaFilesByAccess` now requires each file's `MediaFolderID` to match that library, preventing media versions from other libraries from appearing in results.
> - Existing access predicates (`FileAllowedByAccess`, allowlists, disabled-library checks, quality ceilings, and `SelectedFileID` handling) still apply on top of the library scope.
> - Behavior is unchanged when no presentation library is selected; the unrestricted fast path remains only in that case.
> - Adds table-driven tests in [access_filter_test.go](https://github.com/Silo-Server/silo-server/pull/1109/files#diff-44e47dde15465733391e3160cf419b2d57a89728d020941915afcafe90b07acc) covering scoping, allowlist interaction, disabled libraries, quality ceilings, and `SelectedFileID`.
> - Behavioral Change: callers that previously received media files across all libraries when a `PresentationLibraryID` was set will now only see files from that library.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 684174b.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * File access filtering now limits results to the selected presentation library when one is specified.
  * Existing access restrictions, disabled libraries, quality limits, and file selections continue to be enforced without bypass.


internal/catalog/access_filter.go
```diff
@@ -222,7 +221,8 @@ func FilterMediaFilesByAccess(files []*models.MediaFile, filter AccessFilter) []
 
 	filtered := make([]*models.MediaFile, 0, len(files))
 	for _, file := range files {
-		if FileAllowedByAccess(file, filter) {
+		if FileAllowedByAccess(file, filter) &&
+			(filter.PresentationLibraryID == nil || file.MediaFolderID == *filter.PresentationLibraryID) {
```
Quick104, line 225: Putting the scope here applies it to every caller, including the two playback paths. `GetWatchDetail` (`internal/catalog/detail.go:2890`, `:2947`) and `internal/watchtogether/selection_resolver.go` both filter through this function, and the resolver takes `detail.Versions[0].FileID` at `:73` and rejects an empty list at `:55`. The result is that which file actually plays depends on whether the client sent `library_id`, not on quality or user choice.

If the intent is presentation scoping, the catalog detail seam is the narrower place for it. See the review body for the broader question.

Quick104 (commented):
The code does what the description says and the new test is honest. My concern is the premise. This turns `library_id` into a file-visibility filter for every caller of `FilterMediaFilesByAccess`, unconditionally, and that changes behavior well outside the detail page #1032 is about. I'd like a decision on the product question before this merges.

## The premise

Silo merges the same movie across libraries by content ID on purpose: one item, many versions. `library_id` has meant "which library am I browsing" — it picks the metadata-language fallback (`internal/catalog/detail.go:815`) and validates item membership (`internal/catalog/detail.go:1003`). This makes it also decide which files exist. Four consequences worth deciding on explicitly:

- **Results depend on how you got there.** `web/src/pages/ItemDetail/index.tsx:167` reads `libraryId` from the URL, so it is present when you arrive from a library page or a library-scoped section and absent from Home, search, and direct links. After this change the same movie shows one version from the library grid and two from Home, with nothing in the UI explaining the difference.
- **The "Movies" + "Movies 4K" layout loses its point.** Opening the shared title from "Movies" hides the 4K file entirely. That split-library setup is the case version merging was built for.
- **The admin curation view narrows too.** "Media locations" (`web/src/pages/ItemDetail/MovieContent.tsx:322`, `EpisodeContent.tsx:403`, gated on `canCurateMetadata`) exists to show every file backing an item. Scoping it by browse path removes the signal an admin needs to spot a cross-library duplicate.
- **It reaches playback, not just presentation.** `GetWatchDetail` (`internal/catalog/detail.go:2890`, `:2947`) and watch-together selection run through the same filter. `internal/watchtogether/selection_resolver.go:73` takes `detail.Versions[0].FileID` and `:55` returns `ErrInvalidSelection` on an empty list, so the file a room plays now depends on whether the host's client sent `library_id`.

The reporter isn't wrong to want per-library scoping. Both behaviors are defensible, which is the argument for making it the operator's call rather than a silent server-wide change. Options:

1. A server setting under admin → Libraries, default off. Server-wide rather than per-library: an item that lives in two libraries would otherwise have two conflicting answers.
2. Apply the scope at the catalog detail seam only, leaving `GetWatchDetail` and watch-together on the full set, so presentation narrows but playback never loses a version.
3. Close #1032 as working as intended and document that versions are unified by content ID.

I'd take 1 with the default off. The current shape is the one option with no way back.

## Multi-repo checklist

The body says "No web or client contract changes are needed." The wire format doesn't change, but the meaning of `library_id` does — on `/api/v1`, `/api/v2`, and the Jellyfin surface — so CLAUDE.md's client-visible-change checklist applies, and none of its three items is addressed:

- **silo-apple / silo-android**: any client sending `library_id` silently loses versions. No follow-up filed or ruled out.
- **jellycompat parity**: `internal/jellycompat/content_direct.go:267` feeds six read paths, and `internal/jellycompat/handlers_items.go:2436` derives the library from the Jellyfin `ParentId` query parameter. In Jellyfin `ParentId` means "children of this folder," not "filter this item's MediaSources," so third-party clients get behavior their target server doesn't have. `internal/jellycompat/handlers_playback.go:1954` passes `nil`, so one session sees narrowed sources in the item list and the full set at PlaybackInfo.
- **docs**: no `docs/*-api.md` update describing what `library_id` now does.

Separately, `/api/v1` is frozen to critical fixes that keep the bridge usable. This lands there through the shared service; worth a sentence on why it qualifies, or scoping it to v2.

## Optional

`MediaFileAccessSQL` (`internal/catalog/access_filter.go:261`) carries a comment that it must stay in step with its Go counterpart. That's still literally true — `FileAllowedByAccess` is unchanged — but there are now two file-filtering paths with different scoping rules, and `internal/sections/overlay_summaries.go:55` uses the SQL one. It agrees today only because no sections caller sets `PresentationLibraryID`. Either mirror the scope in SQL or note in the Go comment that the scope is deliberately Go-only.

## Verified

`go test ./internal/catalog/... ./internal/watchtogether/...` passes at 684174b. CI is green. I found no correctness defect in the diff itself; the blocker is the product decision above.

---

AI disclosure: review written with Claude Code, model `claude-opus-5[1m]`.

Quick104:
Closing this in favor of a maintainer PR that makes the scoping opt-in.

The diff itself is correct and the test is good — we're carrying the test forward. The change we can't take as-is is the premise: Silo merges the same title across libraries by content ID on purpose, and an unconditional `library_id` scope inside `FilterMediaFilesByAccess` also narrows playback and watch-together selection, hides versions from the admin "Media locations" view, and changes Jellyfin `ParentId` semantics. Details in the review above.

The replacement adds a server setting (`catalog.scope_versions_to_library`, default off), applies the scope only to `/api/v2` catalog reads when it's on, and reports it through a capability endpoint. It will reference #1032 and link back here.

Thanks for the clear diagnosis in #1032 and the fix; it saved real time.

AI disclosure: comment written with Claude Code, model `claude-fable-5-1[1m]`.

Quick104:
Replacement is up: #1111.

PR #1112, opened 2026-09-16
refactor(storage): generalize the artwork store into a blob store

Quick104:
## Problem

Related issue: #774 — extends the S3-optional storage work past artwork.

`internal/artworkstore` is already a complete backend-neutral blob store: filesystem and S3
implementations, an identity lock, atomic publish-by-rename, prefix listing and deletion. Only
artwork uses it. Subtitles, diagnostic bundles, and catalog-seed artifacts each kept their own
S3-shaped storage and were gated on a configured bucket, so a single-node install without object
storage lost those features entirely.

Nothing can move onto the store while it is named for one caller and returns one store, so this
change prepares the ground. It carries no behavior change of its own.

## Approach

Rename the package to `internal/blobstore` and have `Open` return a `Stores` pair:

- **Assets** — artwork, branding, markers, chapter thumbnails, and (next PR) subtitles.
- **Operational** — diagnostic bundles, job artifacts, avatars.

On an S3 backend those are the existing public and private buckets, unchanged. The public/private
split exists because the public bucket can serve browsers directly under token auth; a filesystem
has no such distinction, so a local backend puts both in one root and the key prefixes each caller
already uses keep the namespaces apart. Those prefixes are documented in
`docs/architecture/blob-storage.md`. Artwork keys start with a metadata provider segment, so they
do not collide with them.

Only Assets is wrapped to record the storage identity. A local backend shares that wrapper with
Operational, so a first write through any caller records it. The private S3 bucket is deliberately
left unwrapped: recording its identity would name it as the catalog's assets location and refuse
the real assets store on the next start.

**The settings keys keep their artwork-era names.** `artwork.storage_backend`,
`artwork.local_path`, and `artwork.storage_identity` now govern more than artwork, which is not
ideal, but renaming them costs a migration, a YAML alias, and a rollback hazard for deployments
that have already stored artwork locally. The admin UI labels change instead. This is reversible
later if the naming proves confusing.

A configured private bucket owns the operational store on either backend, including a local one.
Avatars have always lived there whatever backed artwork, so an install that configured a private
bucket and then moved artwork to disk must keep reading its existing `profile-avatars/` keys. Only a
local backend with no private bucket shares one root. Review caught this; the first revision let a
local backend override the private bucket.

Two things came along because the shared root makes them matter:

- The artwork sweep now refuses an empty prefix. Nothing passes one today — `artworkSweepPrefixes`
  names `local/`, `tmdb/`, and `tvdb/` explicitly — but the root now also holds bundles and
  artifacts, and `parseArtworkObjectKey` accepts any `a.b.c` filename, so it would read
  `report.tar.gz` as a revisioned variant and delete it.
- The `S3UserDB` client was constructed in `cmd/silo/main.go` and read nowhere. The SQLite user
  store takes its configuration directly through Litestream. Removed.

## Validation

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean.
- Full `go test ./...`. One failure, `TestAdminResourceCapabilitiesAndScope` in `internal/apiv2`,
  is pre-existing and environment-specific: it reports `"state":"unsupported"` because the
  development machine is macOS with no cgroups. Confirmed identical at the merge base in a scratch
  worktree, so it is not caused by this change.
- `make verify-route-inventory`, `make verify-local-paths`, `make verify-migration-ledger`.
- Affected web tests and the setup-wizard/infrastructure settings suites.
- New tests: local mode with no private bucket returns the same wrapped instance for both stores, a
  configured private bucket still owns the operational store on a local backend, and an S3 backend
  keeps them separate; a write through Operational on a private S3 bucket does not record an identity; a
  missing private bucket leaves Operational nil; the sweep refuses an empty prefix and deletes
  nothing when it does.

Note that `make verify-settings-bindings-all` does not cover these settings keys — they are absent
from `contracts/settings/v1/manifest.json` and `internal/settingskeys/keys.go`. Nothing is renamed
here so it does not matter for this PR, but it is not the safety net it looks like.

## Risks

No migration, no settings changes, no default path changes. An existing local deployment keeps its
recorded identity and its volume mount; an S3 deployment keeps both buckets and every key.

The naming debt is the known tradeoff: `artwork.local_path` now names a root that will also hold
subtitles and diagnostic bundles.

Behavior is otherwise unchanged, so this is a mechanical rename plus a restructured constructor.
It is large by line count and almost entirely `artworkstore` → `blobstore`.

## Checklist

## AI Disclosure

- Harness: Claude Code (T3 Code)
- Tool(s): Claude Code; Claude Code subagent used for adversarial plan review
- Model(s): `claude-opus-5[1m]` (Opus 5, 1M context) for implementation; Fable for the review pass
- Adversarial review: A separate model reviewed the implementation plan against the source before
  any code was written, checking the plan's file and line claims, the encrypted-settings set, the
  prefix-collision assumption, and whether anything sweeps a whole store root. It confirmed the
  sweep is prefix-bounded and that none of the three settings keys is encrypted. It raised four
  blocking findings, all verified against the source and all resolved. Two affect this PR: the
  identity lock would have been bypassed by any write method not forwarded by `recordingStore`
  (relevant once `PutStream` lands in the third PR, and called out in the `Store` interface here),
  and a proposed default-root change would have silently relocated storage for local deployments
  that had not yet recorded an identity — that change was dropped, along with the settings-key
  rename that motivated it. The other two findings land in the third PR.

> [!NOTE]
> ### Generalize `artworkstore` into `blobstore` with separate `Assets` and `Operational` stores
> - Renames the `artworkstore` package to `blobstore` and updates all cross-package references to `blobstore.Store`, `blobstore.ObjectInfo`, `blobstore.ErrNotFound`, etc.
> - `blobstore.Open` now returns a `Stores` struct with `Assets` and `Operational` fields instead of a single `Store`. Operational storage defaults to the same local store as assets, uses a configured private S3 bucket when present, or is nil for S3 without a private bucket. Added `Options.S3Private` for the private client.
> - `api.Dependencies` replaces the `Artwork` and `S3UserDB` fields with a single `Blobs` field (`blobstore.Stores`); router wiring reads `deps.Blobs.Assets` throughout.
> - `configureS3Clients` in [main.go](https://github.com/Silo-Server/silo-server/pull/1112/files#diff-c8a7988838b46a61cad8d450a07cf1f77898988683073e4e5636c1f2452b5c4c) no longer creates a user-database `s3client.Client` from `cfg.S3.UserDB`, leaving user-db replication to the Litestream path.
> - `ArtworkStorageSweeper.SweepPrefix` in [artwork_storage_sweep.go](https://github.com/Silo-Server/silo-server/pull/1112/files#diff-7eba01af6ddbf55c729eb2866415763ecbc2d10b1fdc497ec5594e00b537181a) now rejects empty or whitespace-only prefixes before any storage operations.
> - Admin settings UI and setup wizard labels change from "Artwork storage"/"Local artwork path" to generic "Storage"/"Local storage path".
> - Risk: `blobstore.Open`'s return type change and the removal of `deps.Artwork`/`deps.S3UserDB` are breaking for any out-of-tree consumers of these APIs. Review [open.go](https://github.com/Silo-Server/silo-server/pull/1112/files#diff-5a1b3216f1ec2a201778255cef1643f9db369650f30ecc3cf16a81595c4dfcad) and [router.go](https://github.com/Silo-Server/silo-server/pull/1112/files#diff-d6605b19340f42c0e80ddbc341238970e4996bf5ad100cba281f231f63276e9e) for the new contract.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dd5db06.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Storage now supports separate asset and operational storage, including optional private S3 storage.
  * Local storage can share asset and operational data while retaining readiness checks.
  * Storage settings and setup flows now use broader terminology for artwork and other library assets.

* **Bug Fixes**
  * Storage cleanup now rejects blank paths, preventing unintended processing of unrelated files.

* **Documentation**
  * Updated storage architecture documentation to explain storage models and configuration behavior.


docs/architecture/blob-storage.md
```diff
@@ -1,7 +1,48 @@
-# Artwork storage
-
-Artwork writes and cleanup use `internal/artworkstore.Store`. The store owns
-its filesystem root or S3 bucket; callers use the existing logical artwork keys.
+# Blob storage
+
+Every blob Silo owns is written and cleaned up through `internal/blobstore.Store`.
```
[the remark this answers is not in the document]
Quick104: Fixed in dd5db06bb87e, and the same applies to the Codex comment below.

You are right that the claim was false at this commit. The doc now says artwork, branding, markers, and chapter thumbnails go through the store today, and names `internal/diagnostics` and `internal/adminjob` as still bucket-oriented and still requiring S3, with the reserved prefixes listed so the namespaces do not have to be renegotiated when they migrate. #1113 adds subtitles to that sentence and #1114 completes it.

The storage setting's description in the admin UI and setup wizard was premature for the same reason and is now scoped to library assets.

web/src/pages/setup-wizard/steps/StorageStep.tsx
```diff
@@ -305,16 +305,16 @@ export function StorageStep() {
           disabled={artworkLocked}
           description={
             artworkLocked
-              ? "Locked: artwork has already been stored on this backend and cannot be moved."
-              : undefined
+              ? "Locked: files have already been stored on this backend and cannot be moved."
+              : "Where Silo keeps artwork, subtitles, diagnostic bundles, and exports."
```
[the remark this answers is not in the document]
Quick104: Fixed in dd5db06bb87e — good catch, and it is right beyond this commit too.

`artwork.storage_backend` selects the Assets backend. When a private bucket is configured it owns diagnostics, job artifacts, and avatars whatever that setting says, so describing operational data here was wrong in S3 mode regardless of migration state. The description now reads "Where Silo keeps artwork and other library assets" in both the wizard and Admin › Infrastructure, and the private bucket control keeps describing its own scope.

Reviewing your point also surfaced a real bug one layer down: `blobstore.Open` was overriding a configured private bucket with the local store on a local backend, which would have stranded existing `profile-avatars/` keys. That is fixed in the same commit with a regression test.

docs/architecture/blob-storage.md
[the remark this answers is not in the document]
Quick104: Fixed in dd5db06bb87e.

You were right on both halves. The doc and the storage setting's description are now scoped to what this commit actually migrates — artwork, branding, markers, and chapter thumbnails — and name `internal/diagnostics` and `internal/adminjob` as still requiring S3.

On the wiring: subtitles move onto the store in #1113 and diagnostics, job artifacts, and avatars in #1114, which is where `deps.Blobs.Operational` gains its consumers. This PR is deliberately the no-behavior-change base of that stack.

docs/architecture/blob-storage.md
```diff
@@ -1,7 +1,57 @@
...
+token auth and the private one never does. A configured private bucket owns the
+operational store whatever the backend is: avatars have always lived there, and
+moving artwork to disk must not strand the `profile-avatars/` keys an install
+already uploaded. Only a local backend with no private bucket puts both in one
+root, where the key prefixes each caller already uses keep the namespaces apart:
+
+| Prefix | Owner |
+|---|---|
+| `<provider>/<kind>/<id>/<imageType>/…` | artwork (`internal/artworkkey`) |
+| `branding/…` | branding assets |
```
Quick104, line 33: **Low — reserved-prefix table is incomplete.** The table is meant to fix the namespaces in the shared local root so later callers don't have to renegotiate them. It leaves out three prefixes the Assets store already writes: `collection-images/` (`internal/api/handlers/collection_artwork.go`), `library-posters/` (`internal/api/handlers/libraries_service.go`), and `chapter-images/` (`internal/chapterthumbs/service.go`). A later caller that reads this table as the complete list could choose one of those names. I'm adding the three rows.

internal/blobstore/store.go
```diff
@@ -1,5 +1,9 @@
-// Package artworkstore stores logical artwork keys in local or S3 storage.
-package artworkstore
+// Package blobstore stores logical object keys in local or S3 storage. It backs
```
Quick104, line 1: **Low — package doc overstates current scope.** The comment says the store backs downloaded subtitles, diagnostic bundles, and job artifacts. At this commit those still use their bucket-oriented code, the same over-claim already fixed in `blob-storage.md`. The `Stores.Assets` and `Stores.Operational` field comments in `open.go` say the same thing, and nothing reads `Operational` yet. I'm narrowing all three comments to match the doc.

Quick104 (commented):
Review of `dd5db06`. Two low-severity documentation findings are inline; I'm fixing both in the push that also merges `main`. The branch conflicts with `main` on one line: the Audiobookshelf listener in `cmd/silo/main.go` now takes `networkAccess.Registry`. I'm resolving that in the same push.

docs/architecture/blob-storage.md
Quick104: Fixed in [7fa727f](https://github.com/Silo-Server/silo-server/commit/7fa727f7f1bfdf62aaa96b13eb61aa2618cc33f5). Added `collection-images/`, `library-posters/`, and `chapter-images/` to the table. I checked each against the key it writes: `collection_artwork.go` (`adminCollectionImagePrefix`), `libraries_service.go`, and `chapterthumbs/service.go`. `make verify-local-paths` passes.

internal/blobstore/store.go
Quick104: Fixed in [7fa727f](https://github.com/Silo-Server/silo-server/commit/7fa727f7f1bfdf62aaa96b13eb61aa2618cc33f5). The package doc now says subtitles, diagnostic bundles, and job artifacts still use bucket-oriented code. `Stores.Assets` says subtitles are "(once migrated)", and `Stores.Operational` notes that its callers still take their storage directly. This is a comment-only change; `go build` and `go test ./internal/blobstore/...` pass.

Quick104:
## Review summary: head `2a5bf31`

**Mergeability: 9/10, ready.**

**Addressed**
- Merged `main`, which fixes the conflict. The one conflicting line was the Audiobookshelf listener in `cmd/silo/main.go`: `main` added `networkAccess.Registry`, and this branch renamed `deps.Artwork` to `deps.Blobs.Assets`. After the merge, nothing in the tree references `artworkstore`, `deps.Artwork`, or `S3UserDB`.
- Low: the reserved-prefix table in `blob-storage.md` now lists `collection-images/`, `library-posters/`, and `chapter-images/` (7fa727f).
- Low: the `blobstore` package and `Stores` comments no longer describe subtitles, diagnostics, and job artifacts as migrated (7fa727f).

**Checked, no finding**
- Every production consumer of the renamed store still gets the same instance it got before. Avatars still go through `NewProfileAvatarStore` with the same inputs, and nothing reads `Operational` in this PR. #1114 is the first consumer.
- A private bucket stays unwrapped, so it never records the identity. A local backend with no private bucket shares the recording wrapper.
- The sweep's empty-prefix guard can't block the scheduled task: `readCheckpoint` only resumes from a prefix listed in `artworkSweepPrefixes`.
- Removing `S3UserDB` is safe. Nothing read it, and Litestream reads `cfg.S3.UserDB` directly.

**Validation**
- Local: `go build ./...`, `go vet ./...`, `make verify-local-paths`, focused tests for `blobstore`, `cmd/silo`, `metadata`, `api/handlers`, `apiv2`, `taskmanager`, `branding`, `imagecache`, and `adminjob`, plus the `InfrastructureSettings` and `StorageStep` web tests.
- CI is green at `2a5bf31`. The first Go test run failed in `TestControlSocketReconnectResumesOnlySameOwnerAndInstallation` with a playback control-socket read timeout. That code isn't touched here, the test passed 5/5 locally, and the rerun passed.

**Remaining**
- Optional: `Stores.Local()` has no production caller, here or in #1114.
- The stacked branches (`feat/blobstore-subtitles`, #1114) will need the same `main` merge.

Reviewed with Claude Code (`claude-opus-5-5`).

Issue #1052, opened 2026-09-12
[bug] Library diagnostic counts show zero before loading and partial totals afterward


zZebrahz:
## Problem

Admin → Libraries shows `0` for the collapsed **Troubleshooting** and **Stale External IDs** sections even when they contain items. Expanding a section loads its first page and changes the header count, so the collapsed number does not reflect the current diagnostic state.

## Steps to reproduce

1. Open Admin → Libraries on a server with skipped roots or stale external IDs.
2. Observe that the collapsed diagnostic section shows `0`.
3. Expand the section and wait for its rows to load.
4. Collapse it again and observe that the number has changed to the number of rows loaded.

## Expected

The collapsed headers should distinguish an unloaded count from a confirmed zero. Once loaded, the number should represent the full result count rather than only the pages currently held by the browser.

## Actual

Both sections show `0` before their first query runs. After expansion, they show the number of loaded rows. Because each request is limited to 50 items, a displayed `50` may mean that the first page loaded while additional results still exist.

## Scope

This affects both **Troubleshooting** and **Stale External IDs**. It is the same unloaded-count pattern reported for Ambiguous Roots in #1000, but these sections also use partial loaded-row counts as their header totals. PR #1004 addresses the Ambiguous Roots presentation only.

## Technical notes

- [`SkippedRootsSection`](https://github.com/Silo-Server/silo-server/blob/537abe41566630474122b4a80d08795a4ffd44c3/web/src/pages/AdminLibraries.tsx#L1763) calls `useSkippedLibraryRoots` with `enabled: open`, defaults missing pages to an empty array, and passes `skippedRoots.length` to the header.
- [`StaleIDsSection`](https://github.com/Silo-Server/silo-server/blob/537abe41566630474122b4a80d08795a4ffd44c3/web/src/pages/AdminLibraries.tsx#L2120) follows the same pattern and passes `staleIDs.length` to the header.
- Both query adapters retain a pagination cursor but do not expose a full result total to the component.

## Acceptance criteria

- Show an unknown or loading state before the count has been fetched instead of `0`.
- Show `0` only after the server confirms that the section has no results.
- Display the full result count, or label a loaded-row count so it cannot be mistaken for the total.
- Preserve on-demand loading and pagination for the detailed tables.
- Cover collapsed, loading, empty, populated, multi-page, and error states in tests.

PR #1116, opened 2026-09-17
fix(libraries): show full diagnostic totals instead of loaded-row counts

rangoDJ:
## Problem

Related issue: #1052

Admin → Libraries shows `0` in the collapsed **Troubleshooting** and **Stale External IDs** headers before either section has loaded. After a section opens, the header shows the number of rows the browser holds, so a `50` can mean "first page loaded" rather than "50 results".

## Approach

The web client could not show a real total because neither listing returned one. `GET /api/v2/libraries/skipped-roots` and `GET /api/v2/libraries/stale-ids` now return `total`, the same field `GET /api/v2/libraries/roots` already carries.

- Each repository gets a count query that reuses the page query's search predicate (`skippedRootSearchWhere`, `actionableStaleIDSearchFrom`), so the page and the total cannot disagree about which rows match. `ListActionable`'s parameters are renumbered so the shared fragment can use `$1` for the search.
- `ListSkippedRoots` and `ListStaleIDs` return the total alongside the rows, matching `ListLibraryRoots`. The v1 handlers ignore it, so v1 responses are unchanged.
- `CollapsibleDiagnosticsSection` accepts `count: number | undefined` and renders an em dash with the screen-reader text "Count not loaded" when the count is unknown. Both sections pass `pages[0].total`, so the header stays unknown until the server answers, then shows the full count even with more pages to load.

On-demand loading is unchanged: the first page still loads only when a section opens. The header therefore shows the unknown state until the admin opens it once. Loading a count on mount would add a query to every visit to the page, which the current design avoids.

Ambiguous Roots has the same pre-load `0` (#1000). I left it out because #1004 is open against it and also edits `CollapsibleDiagnosticsSection`, so one of the two will need a small rebase.

## Validation

- CI (run 35168553861) passes both jobs. Go covers build, `gofmt`, `go vet`, `golangci-lint` on changed lines, the settings, route-inventory, migration-ledger and scenario-catalog checks, the OpenAPI artifact byte comparison, the v2 contract diff, the v2 fixtures, and the Go test suite. Web covers lint, format, build and the web test suite.
- The first CI run caught a stale `contract_digest` in `get_system_info_ok.json`; it hashes `openapi.json`, which this change edits, and is updated in the second commit.
- The second run failed only in `TestControlSocketReconnectResumesOnlySameOwnerAndInstallation` (`internal/api/handlers/playback_control_socket_v2_test.go:277`, a 2-second read timeout). This change does not touch playback, and the test passed on the retry; the third commit is empty and only restarts CI.
- Locally, `AdminLibraries.test.tsx` and `libraries.v2.test.ts` pass (40 tests); with the web source changes reverted, 6 of the new assertions fail. Go could not build locally on Windows (`bimg` needs libvips through cgo), so the Go side was verified only in CI.
- The count assertions added to `skipped_root_repo_db_test.go` and `stale_media_id_repo_db_test.go` need `SILO_TEST_DATABASE_URL` and were not run.
- Not exercised against a running server.

## Risks

- Each page request for these two listings now runs a second query. The stale-ID count repeats the lateral library join the page uses so a library-name search counts correctly. These are admin-only diagnostic listings, and roots already pays the same cost.
- `total` is a new required response field on two v2 operations. This is additive for clients; the v2 fixtures vendored into `silo-apple` and `silo-android` will pick it up on their next sync.

## Checklist

## AI Disclosure

- Harness: Claude Code (Claude desktop app, Code tab)
- Tool(s): Claude Code shell and file-editing tools, GitHub CLI (`gh`)
- Model(s): claude-opus-5
- Adversarial review: No independent reviewer. The authoring session reviewed the complete diff for v1 response changes, page/total predicate drift, query parameter renumbering, the unknown-versus-zero header states, and interaction with #1004; findings are reflected in the Approach and Risks sections.

> [!NOTE]
> ### Add `total` to v2 skipped-root and stale-ID collections and show it in admin diagnostics
> - Adds `Count` and `CountActionable` repository methods so skipped-root and stale-ID listings can report the full number of matches independent of page size.
> - Updates v2 response models `apiv2.SkippedRootCollection` and `apiv2.StaleMediaIDCollection` to serialize a `total` field; v1 handlers receive and discard the total to preserve existing responses.
> - Admin diagnostic headers in `admin.SkippedRootsSection` and `admin.StaleIDsSection` now show the server-reported total instead of the number of locally loaded rows, rendering an em dash until a page is available.
> - Risk: `handlers.LibraryHandler.ListSkippedRoots` and `ListStaleIDs` now return an internal API error when the underlying count query fails, whereas previously a count failure was not possible.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c30c05c.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added total matching counts to skipped library roots and stale external ID results, covering all pages rather than only currently loaded items.
  * Admin library section headers now display server-reported totals and accurately indicate when counts are still loading or unavailable.
* **Documentation**
  * Updated API documentation and response schemas to describe the new `total` field for skipped roots and stale IDs.
* **Bug Fixes**
  * Prevented loading and error states from appearing as misleading zero counts.


Quick104 (commented):
Review of c30c05c.

No defects found. Both repositories build the page query and the count query from one shared search predicate. The handlers don't drop rows after the SQL, so `total` counts exactly the rows that paging returns. v1 responses are unchanged. The rematch mutation invalidates the stale-ID query, so the header total refreshes after a rematch. The change covers each acceptance criterion in #1052.

Validation, run on a local merge of this branch with current `main`:

- The DB-backed repository tests (`TestSkippedRootPageSearchBeforePagination`, `TestStaleMediaIDRepository_ListActionable_MatchesPredicate`) pass against a migrated Postgres, including the new `Count` and `CountActionable` assertions that hadn't been run yet.
- `internal/api/handlers` and `internal/contractspec` pass, and so does `internal/apiv2`, apart from `TestAdminResourceCapabilitiesAndScope`. That test also fails on `main` on macOS, so this change didn't cause it.
- `make verify-apiv2-openapi verify-apiv2-web-types verify-apiv2-fixtures verify-apiv2-contract` passes after regeneration.
- `AdminLibraries.test.tsx` and `libraries.v2.test.ts` pass (41 tests). eslint, prettier, and `tsc` are clean on the web files this PR touches.

Remaining: the branch conflicts with `main` in `contracts/api/v2/fixtures/get_system_info_ok.json` (the contract digest). To fix it, merge `main`, then run `make apiv2-openapi apiv2-web-types apiv2-fixtures`. After that, the diff against `main` is exactly this PR's 20 files.

**Mergeability: 7/10 (conditional).** The code is ready to merge once the conflict is resolved and CI passes on the merged head.

Reviewed with Claude Code (claude-opus-5-5).

